PR #31, opened 2024-09-25
Edit first and last name labels

edbourque0:
Added spaces between first/last and name in the application configuration


stonith404:
Thanks :)

PR #71, opened 2024-11-26
fix: if already not root, don't bother

cdanis:
This is useful when running in environments like Kubernetes, where the orchestration system can take care of both setting your effective uid/gid, as well as filesystem permissions on the volume.


stonith404:
Thanks, I somehow missed that there are container runtimes that run the container as a non root user.

stonith404 (approved):

PR #72, opened 2024-11-26
chore: simplify container setup, drop down to node user 

kashalls:
This PR uses the pre-existing [node user](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user). I tried to make the container smaller, but due to the size of node dependencies you can only really shave off 30MiB. 

To further shave off and reduce the complexity of the docker container, you would need to split each application into their each container instead of offering a "all-in-one" container.


stonith404 (changes requested):
This needs some changes before we can merge it. We've discussed in Discord.

PR #77, opened 2024-11-28
feat(geolite): add Tailscale IP detection with CGNAT range check 

s0up4200:
- Added proper detection of Tailscale IPs using CGNAT range (100.64.0.0/10)

Source: https://tailscale.com/kb/1015/100.x-addresses

image


stonith404:
That's a nice addition. Thanks :)

PR #85, opened 2024-12-30
docs: Add "groups" scope to the oauth2-proxy sample configuration

janpfischer:
Adding the "groups" scope which is needed with the introduction of groups in pocket-id. Since this can be easily overseen it should be added to the example.


stonith404:
Makes sense, thanks :)

PR #100, opened 2025-01-06
feat: allow sign in with email

### LINKED ISSUE #18: 🚀 Feature: Fallback auth?
### Feature description

This project is really interesting - what's the fallback scenario for lost keys, though?

### Pitch

Maybe an email fallback, just until a new key is setup?

stonith404:
This feature resolves #18 which suggest that users should be able to sign in with an link sent to their email. This function is only recommended for environment where security isn't a big concern like hosting it in the local network.

The feature has to be enabled in the settings:
image


samiralam:
Seems to work as advertised!

Some feedback:

- Suggest changing the link text from "You don't have access to your passkey?" to "Don't have access to your passkey?" as a more natural English phrase
- Recommend adding the ability to enable email logins separately from email notifications; currently there's no way to enable the first while disabling the second. While "new device" email notifications are a good idea, currently it sends an email every time I login, even from the same device, which gets a bit spammy.

stonith404:
@samiralam Thanks for your feedback. These two changes are now implemented in the latest `stonith404/pocket-id:dev-email-login` image.

samiralam:
Pulled the new image and retested:
- [x] Updated link text
- [x] Email login link redirects to client after logging in
- [x] Email notifications can be enabled separately from email logins

Looks great!

PR #106, opened 2025-01-12
feat: add LDAP sync

kmendell:
PoC of users from ldap so far, Drafting this PR for code review first, and we can continue on from there. 

## TODO
- [x] If a user is removed from LDAP the user does not get deleted in pocket-id https://github.com/stonith404/pocket-id/pull/106/commits/dbc8e3822935da149570c7fbab9cf2553fcc1056
- [x] Add a LDAP Claim for Admin Rights in Pocket ID https://github.com/stonith404/pocket-id/pull/106/commits/3443d4a7112079c9b3d7a06da9a77acd65fba02e
- [x] Admins Shouldn't be able to Delete users synced from LDAP in the UI? Maybe they should? 
- [x] We need to create the Postgres migration for the LDAP ID  and GROUP ID value in the DB. https://github.com/stonith404/pocket-id/pull/106/commits/f3cc45d918517448c580a03c540c9118ccd51ad0
- [x] Fix the initial setup because it doesn't work anymore.
  - This should be fixed now that ldap is not enabled by ENV Variables, so no users will get synced on launch.
- [x] Add and edit e2e tests
- [ ] Merge and Bump version and Release


kmendell:
Current State of this PR as of commit: [7025e91](https://github.com/stonith404/pocket-id/pull/106/commits/7025e91f087be453281f581a720b58b817d815e0) is users only in the ldap service, waiting for testing to confirm functionality, and then we can continue on with groups.

kmendell:
Functionality for Sync is all working. 
Few Issues/Notes:
- If a user is removed from LDAP the user does not get deleted in pocket-id 
   - I compared that to other services and it seems to be the same the users don't get deleted, that is more of SCIM territory. 
- We need to create the Postgres migration for the LDAP ID value in the DB.
- Complete the Frontend UI Tweaks to not allow editing of LDAP Users or LDAP Groups

PR #118, opened 2025-01-16
docs: add docusaurus docs

kmendell:
This should be a full layout with all current docs built in docusaurus. This does include the updated LDAP Sync Environment Variables as well.


.gitignore
```diff
@@ -36,4 +36,22 @@ data
 /frontend/tests/.auth
 /frontend/tests/.report
 pocket-id-backend
-/backend/GeoLite2-City.mmdb
\ No newline at end of file
+/backend/GeoLite2-City.mmdb
+
+# Production
+/build
```
stonith404, line 42:
```suggestion
build
```

PR #130, opened 2025-01-18
fix: always set secure on cookie

cdanis:
HTTPS is required for WebAuthn, so there's no downside to enabling this. And it prevents accidentally leaking an access token to the network in cleartext on an initial HTTP connection.


stonith404:
Makes sense, thanks!

PR #138, opened 2025-01-20
fix: ensure the downloaded GeoLite2 DB is not corrupted & prevent RW race condition

wargio:
This fixes a race condition while updating the mmdb, ensures the downloaded file is not corrupted and avoids pocket-id not being able to generate audit logs with locations due a bad database.

In github you cannot see the char replaced but this is the one.
image


stonith404 (changes requested):
Looks good to me, thank you!

stonith404 (approved):

PR #139, opened 2025-01-20
feat: private IP ranges

cdanis:


stonith404 (approved):
Great, thank you!

eboman:
Hi , I have noticed since this PR I am no longer seeing my public ip addresses in my audit logs. I am behind an Cloudflare tunnel. Until the 20th of Jan I was able to see public ip's . But now I am seeing my tunnel endpoint ip. Not sure if this is an Cloudflare issue only ? But could you check the x-forwarded-for headers?

kmendell:
@eboman  Can you open up a issue for this please? and we can investigate further.

cdanis:
Do you have TRUST_PROXY=true in your env?

PR #142, opened 2025-01-20
Create sample-configurations.md

kamilkosek:
Add a file containing instructions on how to configure well-known self hosted services which support OIDC.


stonith404:
Thanks :)

PR #152, opened 2025-01-20
fix: make CONTRIBUTING instructions work & fix example envs

cdanis:


stonith404:
Thanks. `PUBLIC_APP_URL` doesn't need SSL if it's running on localhost though. And `MAXMIND_LICENSE_KEY` is not required, if not provided it just won't lookup the IPs

cdanis:
> Thanks. `PUBLIC_APP_URL` doesn't need SSL if it's running on localhost though. 

Does Pocket ID meaningfully operate without SSL? I thought you couldn't add a passkey at all

> And `MAXMIND_LICENSE_KEY` is not required, if not provided it just won't lookup the IPs
Good point, and that might change soon anyway with the custom download url. Will adjust soon

stonith404:
Yeah `localhost` is the only exception where adding passkeys work without SSL.

cdanis:
Oh thanks, very good to know. Is it specifically localhost, or localhost as the final domain component, or does it have to do with the resolved IP address?

stonith404:
I'm not sure to be honest but probably the resolved IP.

> Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely.

https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

CONTRIBUTING.md
```diff
@@ -41,8 +41,9 @@ The backend is built with [Gin](https://gin-gonic.com) and written in Go.
 #### Setup
 
 1. Open the `backend` folder
-2. Copy the `.env.example` file to `.env` and change the `APP_ENV` to `development`
-3. Start the backend with `go run cmd/main.go`
+2. Copy the `.env.example` file to `.env` and change the `APP_ENV` to `development`.
+3. `mkdir data` in the backend folder
```
stonith404, line 45:
```suggestion
3. `cd data` in the backend folder
```

Probably `cd`, right?

cdanis: No, I meant `mkdir` -- the instructions don't work with a fresh checkout:

```
💔cdanis@wmftop.nucleosynth.space /tmp/pocket-id/backend 🕟🍵 go run cmd/main.go 

2025/01/21 16:31:02 /tmp/pocket-id/backend/internal/bootstrap/db_bootstrap.go:88
[error] failed to initialize database, got error unable to open database file: no such file or directory
2025/01/21 16:31:02 Attempt 1: Failed to initialize database. Retrying...

2025/01/21 16:31:05 /tmp/pocket-id/backend/internal/bootstrap/db_bootstrap.go:88
[error] failed to initialize database, got error unable to open database file: no such file or directory
2025/01/21 16:31:05 Attempt 2: Failed to initialize database. Retrying...

2025/01/21 16:31:08 /tmp/pocket-id/backend/internal/bootstrap/db_bootstrap.go:88
[error] failed to initialize database, got error unable to open database file: no such file or directory
2025/01/21 16:31:08 Attempt 3: Failed to initialize database. Retrying...```

Maybe we should just fix it in the repo though, and add a data directory?

Comment (kmendell) [reply CONTRIBUTING.md:45]: I just tested, and if you compile the backend and have a valid .env file it should work. As long as the user has permissions it should be able to create the data/pocket-id.db file. At least in my testing

Comment (cdanis) [reply CONTRIBUTING.md:45]: Was it a fresh clone?

The data directory does not exist in a fresh checkout. This will indeed lead to the error I showed unless you mkdir.

Comment (kmendell) [reply CONTRIBUTING.md:45]: @cdanis  When i ran it it created the directory for me , as my user had permissions to create it , ill test again later tonight on a fresh folder , but i did reclone everything before i tried it.

Comment (cdanis) [reply CONTRIBUTING.md:45]: Okay, I tried it again and definitely needed the `mkdir`.

Comment (kmendell) [reply CONTRIBUTING.md:45]: @cdanis I was able to replicate it as well, so i see your point but it also works sometimes i do it, not sure why... even on a fresh clone. Weird...

Comment (cdanis) [reply CONTRIBUTING.md:45]: Another option aside from editing the instructions:

- Remove data from the repo root .gitignore
- Add a .gitignore inside data which ignores everything except itself

Comment (kmendell) [reply CONTRIBUTING.md:45]: I think since this is for contributing, editing the instructions is fine? since most people that would be doing this will be 'advanced' users. Up to you guys, but good catch on this one, not sure why the behavior was inconsistent on my end.

Comment (stonith404) [reply CONTRIBUTING.md:45]: @cdanis Sorry I missed that the `data` folder doesn't get created automatically.

If we would reorder the `internal/bootstrap/bootstrap.go` to

```
func Bootstrap() {
	initApplicationImages()

	db := newDatabase()
	appConfigService := service.NewAppConfigService(db)
	
	initRouter(db, appConfigService)
}
```

the data folder would be created before the database gets initialized. I think this would be the best solution, rigght?

kmendell: I was thining that same thing last night, better to have the code handle it than commit a empty directory to the git repo.

PR #155, opened 2025-01-21
draft: feat: ISP (AS) names and numbers

cdanis:
I don't really know what I'm doing in Svelte, and this needs some doc work and cleanups I'm sure, but this seems to work:

Screenshot_20250119_200714

Tested thanks to `tailscale funnel` plus a VPN.


stonith404:
Thanks for your contribution. I'm not really sure if this addition is useful though. Is it really important for the user to see  the ISP and AS Number when he wants to check if there are unrecognized login attempts? 

I think it would more confuse the user than it helps. Or do you have an example of a service that displays the ISP and and AS Number in their logs?

kmendell:
I think this is a more advanced user feature, maybe if implemented, it can be a toggleable option vs being on by default? that way users that want it can have it but it wont confuse users that don't know or don't want that info.

cdanis:
I think only showing an IP address can be just as confusing to users in some cases.

The GeoLite2 Country/City data is also only medium-accuracy outside the USA and Western Europe, whereas ASN data is accurate everywhere.

Anyway I'd be totally fine putting it behind a toggle or a per-user preference of some sort, or if you don't want the feature that's fine too.

stonith404:
I mean it's not a bad idea but I think there is just no need to add this to the audit log. 

If I look for example at the audit log table I recognize a suspicious event by the city and maybe the IP (if I know my IPs by hard). I don't really think that the the AS Number would help me to recognize a suspicious login attempt.

The goal of the audit log is really only to recognize suspicious events. If the user needs more information about the IP he can just lookup the IP online and get more information like the AS Number. 

Or do you disagree with me?

How is the ASN more accurate than the city? I mean you're looking up the ASN by the IP. When you then try to get the location from the ASN you probably get the same location, or do I miss something?

cdanis:
> How is the ASN more accurate than the city? I mean you're looking up the ASN by the IP. When you then try to get the location from the ASN you probably get the same location, or do I miss something?

IP address blocks are assigned to networks in a centralized way by [RIR organizations](https://www.ripe.net/community/internet-governance/internet-technical-community/the-rir-system/) like RIPE, ARIN, LACNIC, etc.  These registrations are public and searchable, and "what's the AS number associated with this IP address at this time" is a question where there is public data that provides an authoritative answer.

But the "city" of a specific IP address is something that Maxmind makes a proprietary guess at with heuristics and data aggregation, and sometimes hallucinates.  They publish city-level accuracy stats for GeoIP but not GeoLite: https://www.maxmind.com/en/geoip-accuracy-comparison

In the case of a large mobile network, that city might have not much to do with the physical location of a user, and much more to do with how and where that mobile network interconnects with other networks.  Their own stats bear this out in most regions, even if you accept a 100km error radius: https://www.maxmind.com/en/geoip-accuracy-comparison?country=&resolution=100&connection_type=only_cellular&ip_family=both#results

In the case of satellite Internet, Maxmind's location may not even match the country of the user, but instead will often be the city and country of a ground station site.

cdanis:
How about this from a UI perspective: collapse the name and number columns into a single one with just the network name, and have it link out to a site like bgptools or bgp.he.net or peeringdb or another similar site (via the AS number).

stonith404:
@cdanis Okay, I like this idea. Could you elaborate what the "network name" column would contain, just the IP or the city and country?

gngz:
ASN could be useful to get abuse email address

stonith404:
Are you still planning to work on this PR or can I close this?

PR #173, opened 2025-01-24
fix: use EnvConfig.AppURL instead EnvConfig.Host for smtp ehlo

kmendell:
Related to: https://github.com/stonith404/pocket-id/issues/172#issue-2806909214

### DIFF
--- a/backend/internal/service/email_service.go
+++ b/backend/internal/service/email_service.go
@@ -4,10 +4,6 @@ import (
 	"bytes"
 	"crypto/tls"
 	"fmt"
-	"github.com/stonith404/pocket-id/backend/internal/common"
-	"github.com/stonith404/pocket-id/backend/internal/model"
-	"github.com/stonith404/pocket-id/backend/internal/utils/email"
-	"gorm.io/gorm"
 	htemplate "html/template"
 	"mime/multipart"
 	"mime/quotedprintable"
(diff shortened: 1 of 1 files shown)


stonith404:
The argument of the HELO/EHLO command can't be an URL. It must be a domain name or IP:

> The client sends this command to the SMTP server to identify itself and initiate the SMTP conversation. The domain name or IP address of the SMTP client is usually sent as an argument together with the command (e.g. “HELO client.example.com”). If a domain name is used as an argument with the HELO command, it must be a fully qualified domain name (also called [FQDN](https://www.samlogic.net/articles/smtp-commands-reference-fqdn.htm)).

https://www.samlogic.net/articles/smtp-commands-reference.htm

kmendell:
@stonith404 oh gotcha, i missed that part, i just was doing some testing and saw it was still saying localhost , so i thought of trying this one and my tests worked but im also not using gmail.

stonith404:
If your `PUBLIC_APP_URL` is set to `http://localhost` it will still use `localhost`, or did you use another `PUBLIC_APP_URL`?

kmendell:
For my testing setup it was localhost yeah so that makes sense now, This can be closed probably, I misunderstood , didn't know that it could not be a URL. My bad on that one :/

stonith404:
No worries, thanks for trying to fix it.

PR #177, opened 2025-01-25
feat: add oidc end-session endpoint

kmendell:
Related to: https://github.com/stonith404/pocket-id/issues/160

I tested and all seems to work, the endpoint i came up with was https://id.example.com/api/oidc/end-session

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -1,13 +1,15 @@
 package controller
 
 import (
+	"net/http"
+	"strings"
+
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


stonith404:
The endpoint should conform to the [OIDC spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html). I'm still not really sure which points we should take in account and which we should just ignore. The following points should be definitely implement though:

- Logout confirmation screen if no valid id_token_hint is given
- Redirect to the post_logout_redirect_uri if the URL matches the registered one and the id_token_hint is valid

I'll look into it but it's not that easy to implement.

PR #180, opened 2025-01-25
fix: smtp hello for tls connections

kmendell:
This will fix: https://github.com/stonith404/pocket-id/issues/179#issue-2810943894

The StartTLS function was being called before the Hello function. To fix this i created a helper function and called the Hello function in all the functions that return the client. Tested locally with my postfix relay with TLS and running on Port 587.


stonith404:
Great, thank you :)

PR #186, opened 2025-01-27
docs: add version label to navbar

kmendell:
This will fix: https://github.com/stonith404/pocket-id/issues/185#issue-2813499361


kmendell:
@stonith404  In addition to this change, We may want to setup a pipeline to deploy the docs when anything in the docs is updated. Im looking at the latest version and it doesn't seem like it updated with your node js version 22 changes.

docs/docusaurus.config.ts
```diff
@@ -47,6 +47,11 @@ const config: Config = {
         src: "img/pocket-id.png",
       },
       items: [
+        {
+          to: "#version", // Use a hash link to trigger the onClick
+          label: "Version",
```
stonith404, line 52: I would leave the label empty while the version is loading.
```suggestion
          label: " ",
```

docs/versionlabel.js
```diff
@@ -0,0 +1,20 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
```
stonith404, line 1: Could you use typescript instead of javascript and put the file into the `src ` folder?

docs/versionlabel.js
```diff
@@ -0,0 +1,20 @@
...
+
+    function readVersionFile() {
+        return fetch('https://raw.githubusercontent.com/stonith404/pocket-id/refs/heads/main/.version')
+          .then(response => response.text())
+          .catch(error => `Error reading version file: ${error}`);
+      }
+
+    function getVersion() {
+        readVersionFile().then(version => {
+          const navbarItem = document.querySelector('.navbar__item[href="#version"]');
```
stonith404, line 13: This doesn't work on the navbar on mobile.

PR #189, opened 2025-01-30
fix(caddy): trusted_proxies for IPv6 enabled hosts

apearson:
## Problem
Current the Caddyfile.trust-proxy config only includes 0.0.0.0/0 which only includes IPv4 proxies.

## Proposed Solution
Adding ::/0 (the equivalent of 0.0.0.0/0 for IPv6) allows for IPv6 proxies to be trusted. 

## Testing
1) Ran docker image without modification on a IPv6 host with network_mode:host and a caddy server as a reverse proxy in front of the docker container [1].  
2) Attempt login and check audit logs for IP Address
3) Modify running docker container with `trusted_proxies ::/0` and reload caddy with changed Caddyfile
4) Attempt login and check audit logs for IP Address (see screenshot)

[1] Caddy site block on IPv6 enabled host
```
auth.example.com {
	reverse_proxy localhost:8389
}
```

Screenshot 2025-01-29 at 10 15 21 PM

### DIFF
--- a/reverse-proxy/Caddyfile.trust-proxy
+++ b/reverse-proxy/Caddyfile.trust-proxy
@@ -1,11 +1,14 @@
 :{$CADDY_PORT:80} {
 	reverse_proxy /api/* http://localhost:{$BACKEND_PORT:8080} {
 		trusted_proxies 0.0.0.0/0
+		trusted_proxies ::/0
 	}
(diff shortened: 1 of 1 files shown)


stonith404:
Great, thanks for your contribution!

PR #191, opened 2025-01-31
docs: Add Immich and Headscale client examples

jeffreygarc:
Added documentation for setting up Immich and Headscale.


stonith404:
Thanks :)

PR #194, opened 2025-02-01
docs: Added Gitea and Memos example

PrtmPhlp:
Added examples for Gitea and Memos. 
Your project is great!


stonith404:
Thanks :)

PR #196, opened 2025-02-01
docs: add fly.io setup guide

kmendell:
This will close: https://github.com/stonith404/pocket-id/discussions/181#discussion-7875463


kmendell:
converting to a draft as when i was testing earlier it didn't give me the option for a hobby plan only pay as you go.

kmendell:
this is no longer possible as the hobby plan does not exist

PR #200, opened 2025-02-02
Docs for adding Pocket ID to FreshRSS

UncleArya:
FreshRSS client documentation. I had come across an issue with setting up Pocket ID with FreshRSS due to case sensitivity with usernames. I added that warning to these docs in case anyone else has a similar issue.


kmendell:
@UncleArya Can you add this entry in the sidebar.ts file under the client-examples section and in the correct order (a-z) as well? Otherwise that looks good to me :)

UncleArya:
Added FreshRSS to sidebar.ts :)

stonith404:
Thank you :)

PR #203, opened 2025-02-03
docs: add landing page 

kmendell:
This will need more tweaking as we will need to update the docs to https://pocket-id.org/docs and what not but this is a good starting point to see if you like it or not. 

We also would have to update the index.tsx file to just redirect to this page instead of introduction


kmendell:
This is mostly ready, the only remaining things are below.

Change the Logo URL once we update the /docs url.
Change over all of the docs to use the new /docs url instead of docs.pocket-id.org

PR #211, opened 2025-02-05
docs: add the "HOST" environment variable

jonasclaes:
I've added the HOST environment variable to the documentation, as when you disable Caddy, the backend will only listen on localhost inside of the container, whilst the user is probably trying to reverse proxy it from their own reverse proxy.


stonith404:
Thanks. I think it would make sense to run it on `0.0.0.0` per default because the frontend is running on `0.0.0.0` already. What do you think?

jonasclaes:
That could work too! I was thinking about the potential security impact of doing that, but I suppose there isn't any as the paths are available anyways through `/api` and `/.well-known`. I'll update this PR to change that to `0.0.0.0`, and I'll open a new one over at the `pocket-id/website` repository to mention this variable in the docs, and change the default there to `0.0.0.0` as well. That okay with you, @stonith404?

kmendell:
@jonasclaes  That sounds like a plan!

PR #212, opened 2025-02-05
docs: fix freshrss callback url

RobinMicek:
Fixed the callback url for FreshRSS in the docs. It was missing '/i/oidc', as referenced here https://freshrss.github.io/FreshRSS/en/admins/16_OpenID-Connect.html.


stonith404:
Thanks!

PR #218, opened 2025-02-07
feat: update host configuration to allow external access

jonasclaes:
As discussed in #211, the backend now runs on 0.0.0.0 by default, and is overrideable through the `HOST` environment variable.


kmendell (approved):

stonith404:
Thank you!

PR #225, opened 2025-02-10
feat: user/group source location badge

kmendell:
This adds a simple badge that shows the source of the user or group in the list table.

image


kmendell:
Commit: https://github.com/pocket-id/pocket-id/pull/225/commits/59775f66a1fab53939ee96147afa7725ec5676f4 , provides the source column only if ldap is enabled, but the columns are all messed up....

Issue #111, opened 2025-01-14
🚀 Feature: Allow custom maxmind database download url.


wargio:
### Feature description

For networking reasons the files that pocket-id tries to pull from might not be reachable. it would be nice to be able to set a custom url like `http://internal.domain/mirror/latest-GeoLite2-City.mmdb`.

### Pitch

This should be configurable like for the license key

kmendell:
@wargio  you can manually mount a GeoLite2 DB into the container I believe, Does this not work for you?

wargio:
Well this will not update the dB every x hours. And to update it I will need to restart the app

kmendell:
Understood, Just wanted to clarify to make sure I understood.

wargio:
I think, given how the license is used, you can replace entirely the license key env with just the URL, since this is how the license is used.

stonith404:
I think it makes more sense if you mount the database to the container like @kmendell said.  Updating the database doesn't require a restart and the update can easily be automated with a cron job. 

In my view, implementing environment-specific features like this directly in the app isn’t necessary, especially when a simple workaround is available. Adding such features would only complicate Pocket ID unnecessarily.

wargio:
The issue is if i update the DB it needs to be done when the DB is not open.

**This is a race condition.**

It is true that based on the code (see below) the db is opened and closed each time you call `GetLocationByIP` but still this falls into a race condition which may crash the application.

- https://github.com/stonith404/pocket-id/blob/main/backend/internal/service/geolite_service.go#L46

stonith404:
Fair enough I didn't think about that. In that case it would make sense to add a custom database URL.

PR #226, opened 2025-02-11
Add GEOLITE_DB_URL and handle an empty MAXMIND_LICENSE_KEY

wargio:
Adds `GEOLITE_DB_URL` and handles the edge case where `MAXMIND_LICENSE_KEY` is not defined, avoiding to make a useless http request to the service.

Fix https://github.com/pocket-id/pocket-id/issues/111


backend/internal/common/env_config.go
```diff
@@ -38,6 +39,7 @@ var EnvConfig = &EnvConfigSchema{
 	Host:                     "0.0.0.0",
 	MaxMindLicenseKey:        "",
 	GeoLiteDBPath:            "data/GeoLite2-City.mmdb",
+	GeoLiteDBUrl:             "",
```
stonith404, line 42:
```suggestion
	GeoLiteDBUrl:             "https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-City&license_key=%s&suffix=tar.gz",
```

I would save the default value here.

backend/internal/common/env_config.go
```diff
@@ -56,4 +58,12 @@ func init() {
...
 		log.Fatal("Missing SQLITE_DB_PATH environment variable")
 	}
+
+	if EnvConfig.GeoLiteDBPath == "" {
+		log.Fatal("GEOLITE_DB_PATH environment variable cannot be empty")
+	}
+
+	if EnvConfig.GeoLiteDBUrl == "" && EnvConfig.MaxMindLicenseKey == "" {
+		log.Println("MAXMIND_LICENSE_KEY and GEOLITE_DB_URL are not defined. Without one, the GeoLite City database can't be updated.")
+	}
```
stonith404, line 68:
```suggestion
```

Then we could remove this step because the variables can't be empty as they have a default value.

backend/internal/service/geolite_service.go
```diff
@@ -109,13 +109,21 @@ func (s *GeoLiteService) updateDatabase() error {
 		return nil
 	}
 
-	log.Println("Updating GeoLite2 City database...")
+	downloadUrl := common.EnvConfig.GeoLiteDBUrl
```
stonith404, line 112: This could then also be simplified as `common.EnvConfig.MaxMindLicenseKey` and `common.EnvConfig.GeoLiteDBUrl` are always defined

backend/internal/common/env_config.go
wargio: They actually can if you define them empty.

backend/internal/common/env_config.go
```diff
@@ -38,22 +40,33 @@ var EnvConfig = &EnvConfigSchema{
...
+		log.Fatal("Missing SQLITE_DB_PATH environment variable.")
+	}
+
+	if EnvConfig.GeoLiteDBPath == "" {
+		log.Fatal("GEOLITE_DB_PATH environment variable cannot be empty.")
+	}
+
+	if EnvConfig.GeoLiteDBUrl == "" {
+		// Warn the user, but do not fail.
+		log.Println("GEOLITE_DB_URL environment variable is empty; without it, the GeoLite City database can't be updated.")
```
stonith404, line 70: It can't be empty because it has a default value? Or do I miss something?

wargio: You can set it empty, as long as it is defined in the environment

kmendell: @wargio  im not sure if setting the environment var  to blank will work since there is default value set, it will just treat it as if the var wasnt defined at all , i havent tested that theory just saying what im observing.

stonith404: Yeah okay I'm not sure if it takes the default value if it is set to an empty string. I'll look into it tomorrow.

wargio: Seems that the behaviour is correct as long as it doesn't have a default, so i'm going to remove that check.

https://github.com/caarlos0/env/blob/main/env_test.go#L165

kmendell:
@wargio Can you also update the .env.example with the new Env vars you had added?

kmendell (commented):
This latest version looks good to me, Ill leave it to to stonith though for final approval.

wargio:
> Can you also update the .env.example with the new Env vars you had added?

i was going to modify this, as you asked, but i don't think there is a need; a normal user should not use this.

kmendell:
Fair enough, I think the documentation will be enough as long its there, Ill get a PR opened there to add in this new env var.

stonith404:
Alright LGTM now. Thank you :)

PR #227, opened 2025-02-12
feat: Added Helm chart

hobbit44:
I primarily use k8s and i noticed you didnt have one yet so threw together somthing to get started.

Please let me know if you want anything changed inline with any design parameters you have and im happy to keep maintaining this.

This uses the OCI standard to push charts to the ghcr.io pakages repo alongside your docker image. I would also recommend publishing this to [ArtifactHub](https://artifacthub.io/).

This can be tested by using:
```
helm template test oci://ghcr.io/hobbit44/pocket-id/charts/pocket-id
```

If you want to include this im happy to update the docs in you other repo to add an installation guide and useful values examples.


kmendell:
Great Idea, my only reservation is that, as Sontih404 or I ont use Kubernetes (at least i dont i dont think he does either) , So this would be hard to support, I think this is great a comunity supported project, but im not sure how @stonith404 feels about it. Ill talk to him and see.

stonith404:
@hobbit44 Thanks for your contribution, I really appreciate it. 

Pocket ID is meant to be simple, so I don’t think we should officially support Kubernetes. If you come to the point where you have to scale Pocket ID, I’d actually recommend using something like Keycloak instead, as it’s better for more complex setups.

That said, I understand that if you run everything on Kubernetes, it would be nice to have Pocket ID there too.

Since I don’t have much experience with Kubernetes, adding it would make things more complicated especially when updating configs or debugging issues.

Because of that, I suggest you create your own repo. You could set up an action to automatically build and push the chart packages, and we could then reference your project in the docs, just like we do for Proxmox and Unraid.

hobbit44:
@stonith404 @kmendell Not a problem. It is just a case of everything i have is in K8s so i was building the configuration anyway. I'll do as you suggest then and publish it under my user and maintain it there. If you decide later you want to include it then I'm happy to do so.

kmendell:
@hobbit44  as stonith mentioned if you setup your own repo, We can reference that in the documentation for others that also use kubernetes, but thats up to you if you want to do that, just let us know.

hobbit44:
@kmendell @stonith404 I've added the helm chart as a standalone repo that syncs with the version updates you put out. I'm more than happy to take feedback on it and your welcome to reference it as you see fit. 

https://github.com/hobbit44/pocket-id-helm

kmendell:
Ill add this to the Unofficial Install methods in the docs :) Thank you!

Issue #257, opened 2025-02-22
🐛 Bug Report: When the Public URL has a trailing slash, discovery URLs are incorrect


nikdoof:
### Reproduction steps

When the Public URL, defined by env vars, has a trailing slash, the OIDC discovery endpoint generates incorrect URLs:

```shell-session
PUBLIC_APP_URL=https://id.doofnet.uk/
```

### Expected behavior

Valid URLs should be generated.

### Actual Behavior

Resulting OIDC discovery:

**https://id.doofnet.uk/.well-known/openid-configuration**
```json
{
   "authorization_endpoint":"https://id.doofnet.uk//authorize",
   "claims_supported":[
      "sub",
      "given_name",
      "family_name",
      "name",
      "email",
      "email_verified",
      "preferred_username",
      "picture"
   ],
   "end_session_endpoint":"https://id.doofnet.uk//api/oidc/end-session",
   "id_token_signing_alg_values_supported":[
      "RS256"
   ],
   "issuer":"https://id.doofnet.uk/",
   "jwks_uri":"https://id.doofnet.uk//.well-known/jwks.json",
   "response_types_supported":[
      "code",
      "id_token"
   ],
   "scopes_supported":[
      "openid",
      "profile",
      "email"
   ],
   "subject_types_supported":[
      "public"
   ],
   "token_endpoint":"https://id.doofnet.uk//api/oidc/token",
   "userinfo_endpoint":"https://id.doofnet.uk//api/oidc/userinfo"
}
```

Pocket ID doesn't correct obvious URL errors as such, so returns a 404 for each URL provided.

### Version and Environment

v0.35.0, Container ran in Kubernetes, fronted by Traefik, using Caddy in-image.

nikdoof:
Looks like its down to: https://github.com/pocket-id/pocket-id/blob/7446f853fcb94303d3f5ef253d6056a0607a3f02/backend/internal/controller/well_known_controller.go#L31

It's joining the strings together rather using a URL path join.

stonith404:
Fixed in `v0.35.1`.

PR #258, opened 2025-02-22
fix: generate valid urls when app url contains a trailing slash (#257)

nikdoof:
A potential solution to #257, I am by no means a Go developer, so feel free to rewrite/close as needed.

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -1,12 +1,13 @@
 package controller
 
 import (
-	"github.com/pocket-id/pocket-id/backend/internal/common"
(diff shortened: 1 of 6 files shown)


stonith404:
Oh, I've created a fix a few minutes ago. This should be fixed in `v0.35.1`. But thanks for trying to fix the issue, I appreciate it.

I've decided to throw an error if the `PUBLIC_APP_URL` contains a path or trailing slash because it would cause issues on other parts of the application if the URL contains a path. Additionally, Pocket ID doesn't support to be run an a path so it doesn't make sense that a user uses a `PUBLIC_APP_URL` that contains a path or trailing slash.

Issue #264, opened 2025-02-24
🐛 Bug Report: Pocket ID uses STARTTLS used by default if SMPT port is not 465


savely-krasovsky:
### Reproduction steps

I think we need to provide a proper way to select TLS and STARTTLS. In my case provider blocks all SMTP ports, but provides a proxy which can be used with any arbitrary port (e.g. 443). I tried to enable TLS, but it tries STARTTLS in that case, while proxy supports only TLS.

Also I find it a little bit confusing that TLS option could activate STARTTLS at all.

### Expected behavior

An option to select TLS or STARTTLS implicitly. 

### Actual Behavior

One TLS option which can lead to either TLS or STARTTLS

### Version and Environment

v0.35.1

kmendell:
@savely-krasovsky Are you wanting a dropdown box to select the TLS option? right now the email ports automatcially select TLS if the port is 465, and StartTLS if the port is 587, and no TLS if the port is 25

savely-krasovsky:
@kmendell yep, in my case 25, 465 and 587 are blocked. But Fastmail has SMTP proxy which can be used with any port, but it requires TLS support (not STARTTLS). Dropdown would be nice.

savely-krasovsky:
I prepared PR #265, will test it today.

kmendell:
@savely-krasovsky Im assuming that was a typo you meant to put that in the other issue? Either way i will try to work on this a bit later today.

savely-krasovsky:
Yep, sorry!

kmendell:
@savely-krasovsky I made PR #268 , but this is what i came up with , just as a fyi for you 

<img width="1099" alt="Image" src="https://github.com/user-attachments/assets/9755409f-d3b6-4547-acb2-7c4fc901788a" />

stonith404:
Released in `v0.35.3`. Thanks @kmendell.

PR #268, opened 2025-02-25
fix: add option to manually select SMTP TLS method

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/264

I tested and this works, but could maybe be cleaner, or some stuff could be changed a bit.


savely-krasovsky:
@kmendell have just tested. It works! Used 443/TLS test email has sent. One problem: in UI case doesn't preserve, you pick `TLS`, click Save and it became `tls`. Same then you load page for the first time. I use Firefox 135 if that important.

kmendell:
@savely-krasovsky yes the case of the characters doesn't matter, im not sure why that's happening , its writing the value vs the label, so its not the end of the world, but ill see if i can fix it still.

stonith404:
Thanks :)

Issue #112, opened 2025-01-14
🚀 Feature:  Support Device Authorization Flow


wh450g:
### Feature description

Ability to authenticate devices similar to users
(most other oidc solutions offer an option to use  a "device authentication endpoint" like 
.../protocol/openid-connect/auth/device.)

### Pitch

While I'm able to use pocket id to authenticate me against the web ui of an virtualization platform called Incus (similar to proxmox), I can't authenticate via cli, as there is no "device authorization endpoint" set.

kmendell:
For anyone wondering see my PR https://github.com/pocket-id/pocket-id/pull/270 as this is the WIP for the Device Code authorization.

PR #270, opened 2025-02-25
feat: device authorization endpoint

kmendell:
This took me awhile to figure out but it does work as far as i can see. More of a base layout for the feature as this was my first attempt. It follows the OIDC Spec as well. 

Fixes: https://github.com/pocket-id/pocket-id/issues/112


kmendell:
It seems i missed the setp to add the endpoint to the .well-known controller, I will try to add that here tonight

stonith404:
Thank you that looks nice. I haven't had the time to review to code yet but I have noticed a few things:
1.  If the user hasn't authorized the client before he should be prompted if he wants to share the requested scopes with the client. Like when you authorize a client with the authorization code flow:
  > The authorization server should then inform
     the user about the action they are undertaking and ask them to
     approve or deny the request.  Once the user interaction is complete,
     the server instructs the user to return to their device.
2. If the user tries to authorize a client a second time, a `duplicated identifier` error is returned. This is because the backend tries to create an entry in the `user_authorized_oidc_clients` table even though it already exists. You can take [this](https://github.com/pocket-id/pocket-id/blob/e46471cc2d6eef37c43b43ccd6dba1bc3a75e898/backend/internal/service/oidc_service.go#L70) as a reference how it was solved in the authorization code flow. 
3. `VerifyDeviceCode` doesn't do some required checks that are present in the `Authorize` function. For example the user groups don't get checked. You might want to extract some shared logic from the `Authorize` function that you can reuse in the `VerifyDeviceCode` function. This would fix 2. automatically.
4. As this is just an additional flow to the authorization code flow, I don't think that's necessary that an admin has to enable the flow for a client.

kmendell:
I will work on these and try to refactor my stuff to fit better as well. Ill keep you in the loop :)

kmendell:
The return to device message looks like below: 
image

kmendell:
@stonith404 I think i got everything you mentioned. Let me know what you think!

kmendell:
Verify Device Code page if no code is passed in the url: 
image

kmendell:
If a test client is needed i was testing using incus.

kmendell:
Yes it will, but it has to be reviewed against the OIDC spec , which takes time, and elias has been busy and hasn't had the time to sit and review this in full yet.

stonith404 (approved):
Looks good to me now, @kmendell. Feel free to test it, and when you're ready you can merge it.

kmendell:
Looks good to me as well. :)

Issue #267, opened 2025-02-24
🚀 Feature: QR Code / 2nd Device / Assisted Sign-In


acidRain-burns:
### Feature description

I tried to look up the name for this process, but couldn't find what it would be called beyond just a QR Code sign in, so bear with me:

The pocket-id admin could enable this globally for all OIDC clients, or opt-in individual clients. The user would, when authenticating with approved clients, have the option to complete the request on another device. When that option is selected by the user, pocket-id would present the user with a QR code where the request would already be populated. If the user was already signed in to pocket-id, they would be prompted to approve or deny the sign-in. If they were not yet signed in, they would be requested to sign in first. Ideally, the QR Code URL is just a normal URL that could be typed in as a fallback, maybe something short like pocket-id.example.com/approvals/shortlivedcodehere)

Examples of this:
 - Discord
 - Steam

### Pitch

For my use-cases, when using a device that does not support passkeys (TVs, XR Headsets, VMs), it would be appreciated to have the ability to complete the sign-in on your phone (via qr code) or another device (with say, a short-lived code).

penalte:
Duplicated #112

acidRain-burns:
I looked at #112; At the time the scope seemed different enough that this seemed warranted, but maybe I am mistaken. Feel free to close if this is the case. I can also add my thoughts to #112

kmendell:
I will close this issue and we can continue the conversation in #112. Just upvote that issue and add your comments. Thank you!

kmendell:
@stonith404 My mistake i thought they were the same thing, sorry for closing that out by accident.

acidRain-burns:
@stonith404 I can think of a few examples right off the bat; I would like to preface this by saying I am probably not representative of a large part of you users BUT I do think this feature addresses a lot of edge cases, of which mine are a sample. _I also acknowledge that there may be ways to technically accomplish a sign in, though they are not at all convenient._  Many of my users are passkey-only: for many of the services pocket-id protects, they made an account by signing in to the service via pocket-id, and some have the option to sign in via passwords disabled or the service makes you choose between passkeys and passwords.

1.  TVs: I have multiple android TVs, with poor support for  hardware passkeys or bitwarden, and connecting via phone is a pain as well, as I am not the only user signing into these devices. There are quite a few services that are very useful to access for configuration and updates to the device -- I use pingvin-share to store/update apks! Other services include jellyfin (has the quick connect though!), immich (very long api key, hard to switch users), and others.
[...]

PR #271, opened 2025-02-26
feat(account): allow creation of own login codes

Pyxels:
## Feature description

Adds button to allow users to create their own One-time links.

## Motivation

As a user I might want to add new devices with their own passkey. Currently I need an admin to create that One-time link for me, this PR allows the users to create their own links.

(See the relevant discussion #95)

---

I tried to implement this as minimally invasive as possible, thus I completely reused the admin component and moved it into the lib. Beware, I also slightly changed the info text on the popup, to be universally correct for admins and users creating their own links.

Please let me know if you'd like this handled differently!


stonith404:
Nice, thank you! I think this would also fix #267. 

Two things:
- The `/users/:id/one-time-access-token"` endpoint can only be requested by admins. So currently your implementation only works if an admin creates a one-time-access link. See https://github.com/pocket-id/pocket-id/blob/e46471cc2d6eef37c43b43ccd6dba1bc3a75e898/backend/internal/controller/user_controller.go#L38 This would need to be changed. You could add a `/users/me/one-time-access-token"` endpoint similarly to the profile picture endpoints.
- Please also add an integration test to [one-time-access-token.spec.ts](https://github.com/pocket-id/pocket-id/blob/main/frontend/tests/one-time-access-token.spec.ts). This should be pretty easy with Playwright.

Pyxels:
Thanks for the quick feedback!
I didn't check that indeed.

It should now work. I also added a new test, but to the account tests at `frontend/tests/account-settings.spec.ts`, which I thought more fitting.

Let me know if you'd like something done differently!

kmendell:
This looks good to me after the changes, i cant approve teh WF run to make sure tests work, but ill try to test  this all locally here in a bit as well.

kmendell:
> I made some UI changes to the login code creation, and I've also added the option for a user to enter the code manually. Could you please test this and let me know if everything works well?
> 
>  Screenshot.2025-03-03.at.16.08.52.2.mp4

That looks great! Ill test later as well :)

kmendell (approved):
I tested and everything seems to work great! Should be good to merge. I think this could be integrate with https://github.com/pocket-id/pocket-id/pull/270 as well. We could even use the same code page?

PR #273, opened 2025-02-27
fix(account): call correct function to set own profile picture

Pyxels:
Very minor fix to allow setting ones own profile picture.

## Issue
When uploading a picture on `/settings/account`, a 500 is returned by the backend due to an invalid UUID.

## Fix
Call the correct function to populate the UUID with ones own.


kmendell (approved):
Looks good to me. :)

kmendell:
I just relaized when i tested this i was a admin user, I will need to retest with a non admin user to make sure it works, i have a few thing to attend to before i can review that again though.

stonith404:
Thanks!

PR #277, opened 2025-02-28
feat: enable sd_notify support

savely-krasovsky:
Podman supports it natively, that allows to manage dependencies without using healthchecks. For example oauth2-proxy automatically fetches OIDC configuration after every launch, in case there Pocket-ID is not yet ready it will fail.


savely-krasovsky:
I am using Podman Quadlets, they are embedded into Podman for almost two years now I think. Basically they allow to launch containers as systemd services. By default those systemd start just after process inside container launches. So I cannot run immediately `oauth2-proxy` for example, it will try to fetch `/.well-known/openid-configuration` and since Pocket-ID is still launching it will fail. `sd_notify` allows Pocket-ID to notify systemd that process is ready (to accept http-requests in our case).

In case of Quadlet I can enable this behavior by setting `Notify=true`. Traefik and many other support it.

kmendell:
Couldn't you use the depends_on: directive in the compose file? then oauth2-proxy wont even try to start till pocket id is started?

savely-krasovsky:
It won't help, since Podman doesn't understand when service is ready. `depend_on` requires either `sd_notify` support or healthcheck. That's why you see `depend_on` usually with databases and with their healthcheck.

kmendell:
Pocket id does have a health check endpoint: 

Im guess on what im getting at here is this could add extra code that is not needed if something already is in place. 

```yaml
    healthcheck:
      test: "curl -f http://localhost/health"
      interval: 1m30s
      timeout: 5s
      retries: 2
      start_period: 10s
```

savely-krasovsky:
Healthcheck is the second way, yes. But I don't think supporting `sd_notify` is a bad idea. If `NOTIFY_SOCKET` in unavailable (e.g. it's not passed by default in Docker) this will just harmlessly fail.

Podman is pretty popular in selfhosting community. Healthcheks are a bit slower, notifying systemd directly saves time.

kmendell:
Gotcha, i guess i've never used it, so if elias thinks it's beneficial then we can merge it, i just don't know how popular this is .

stonith404:
A check if the frontend is ready would also be required as the frontend runs independed from the backend. Do you have a suggestion how we could achieve that? The only solution that comes in my mind would be to fetch the frontend from the backend and check if the status code is 200 but in that case this wouldn't really be faster than a healthcheck, right?

savely-krasovsky:
@stonith404 it's a valid point about frontend, though usually if we depend on container readiness, backend is the only thing that matters. In case of Podman `ready` flag from sd_notify and `healthy` state are two distinct things. So we can support health checks later and people will have an option to choose. Docs can highlight that.

stonith404:
@savely-krasovsky Alright, that makes sense, thanks. 

What do you think about notifying the daemon without the library? The logic seems very simple so I think we could do this without library: https://github.com/coreos/go-systemd/blob/main/daemon/sdnotify.go

savely-krasovsky:
@stonith404 sounds reasonable! Please review, I didn't change logic much, just dropped unnecessary boolean and variable unsetting.

kmendell:
I cloned this and tested it and it seems to build and start, Though im on my macbook so im not entirley certain if the sd_notify check will work on here. Otherwise code looks good to me as well. 

Ran e2e tests locally as well and all 38 passed.

savely-krasovsky:
sd_notify will and should work only under Linux environment. Under other OSes it should just noop. In case Podman it will work automatically. In case of Docker you will need to pass systemd socket as dedicated volume.

backend/internal/utils/systemd/sdnotify.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"net"
+	"os"
+)
+
+type SdNotifyState string
+
+const (
+	// SdNotifyReady tells the service manager that service startup is finished
```
stonith404, line 11: As we are only notifying the daemon when Pocket ID is ready when can remove these states as they are not used. Or I'm missing something?

backend/internal/utils/systemd/sdnotify.go
```diff
@@ -0,0 +1,53 @@
...
+	// it completed reloading.
+	SdNotifyReloading SdNotifyState = "RELOADING=1"
+
+	// SdNotifyWatchdog tells the service manager to update the watchdog
+	// timestamp for the service.
+	SdNotifyWatchdog SdNotifyState = "WATCHDOG=1"
+)
+
+// SdNotify sends a message to the systemd daemon. It is common to ignore the error.
+func SdNotify(state SdNotifyState) error {
```
stonith404, line 30:
```suggestion
func SdNotifyReady() error {
```

Then we could rename the function and remove the state parameter.

stonith404 (approved):
Thank you :)

Issue #288, opened 2025-03-02
🐛 Bug Report: Some profile pictures end up sideways


jfro:
### Reproduction steps

1. Go to account settings
2. Click profile picture to upload a new photo 
3. Choose a portrait photo (like from an iPhone)

### Expected behavior

It should respect the orientation of the photo (not sure if this is in EXIF data?) 

### Actual Behavior

End up with sideways profile photo once I refresh/navigate around

### Version and Environment

traefik proxy, docker based, Pocket ID v0.35.4

kmendell:
@jfro Can you pull and test this image to see if it fixes the issue? `ghcr.io/kmendell/pocket-id:exif-photos`

kmendell:
You can still test that image, but stonith found a different library to fix this, vs the way i did in that specific image.

stonith404:
Thanks @kmendell for your attempt. This should be fixed in the latest version.

PR #290, opened 2025-03-03
fix: images with exif data not showing correctly

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/288

### DIFF
--- a/backend/internal/utils/image/profile_picture.go
+++ b/backend/internal/utils/image/profile_picture.go
@@ -3,26 +3,54 @@ package profilepicture
 import (
 	"bytes"
 	"fmt"
+	"image"
+	"image/color"
+	"io"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


kmendell:
I tested this locally and took a portrait mode picture from my phone and it's seemed to be the correct orientation. Original photos still work as well.

kmendell:
Waiting on comment in https://github.com/pocket-id/pocket-id/issues/288 to confirm this resolves the issue.

stonith404:
Thanks. I think it makes sense to rely on `github.com/disintegration/imageorient` in this case. It's a package from the same maintainer as `github.com/disintegration/imaging`, which we are using for image cropping. I feel safer using this library than implementing this logic ourselves. 

4a909d842d7284d584c08ac6296b6dd9d92fc484 should fix this issue because of that I'm closing this PR.

Issue #62, opened 2024-11-09
🚀 Feature: API Key authentication


fallenleavesgocrunch:
### Feature description

The ability for a backend application to create pocket-id users through a REST API.

### Pitch

I'd also love to integrate pocket-id in to software I'm writing; but for that to work I cannot manually administer adding users. I'd love it if there were some kind of REST api where I can submit a new user to pocket-id and have it send out the authenticate email with one-time setup link and a redirect back to my app when it's setup.

app.backend -> pocket-id.backend/add-user?username=xxxx?group=zzzzz&returnto=urlurlurl

pocket-id would then create the user, add them to the group, and email them out the one-time invite. When the user has authenticated their email and added a passkey it would then redirect them to the returnto url.

stonith404:
Adding users via the rest API is already possible but you would need your access token for authentication which is suboptimal for automations as the access token is only short lived. 

We would have to add support for API key authentication where the user can create and revoke them in the admin UI and then use the key for authentication.  Because of that I'll rename this feature request to "API Key authentication".

kamilkosek:
Hi,
I´d love to see a way to interact with the API. 
What I like to build is a "landing page" or call it "portal" where users can see all the apps and services which are available for their authenticated identity.

stonith404:
This is now available in `v0.39.0`. Thanks @kmendell.

PR #291, opened 2025-03-03
feat: api key authentication

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/62


kmendell:
Just a note, i marked this as ready so it can be reviewed overall, but i still plan on adding tests as well.

frontend/src/lib/services/api-key-service.ts
```diff
@@ -0,0 +1,27 @@
...
+			params: options
+		});
+		return res.data;
+	}
+
+	async create(apiKey: ApiKeyCreate): Promise<ApiKeyResponse> {
+		const payload = {
+			...apiKey,
+			expiresAt:
+				apiKey.expiresAt instanceof Date ? apiKey.expiresAt.toISOString() : apiKey.expiresAt
```
stonith404, line 17: Is the check really necessary? According to the type `expiresAt` is a `Date.`

frontend/src/routes/settings/admin/api-keys/api-key-list.svelte
```diff
@@ -0,0 +1,113 @@
...
+	});
+
+	let requestOptions: SearchPaginationSortRequest | undefined = $state();
+	const apiKeyService = new ApiKeyService();
+
+	async function loadApiKeys(options?: SearchPaginationSortRequest) {
+		try {
+			const keys = await apiKeyService.list(options);
+			// If the service doesn't support pagination yet, convert the response
+			if (Array.isArray(keys)) {
```
stonith404, line 34: It should support pagination so we can remove this,

frontend/src/routes/settings/admin/api-keys/api-key-list.svelte
```diff
@@ -0,0 +1,113 @@
...
+						toast.success('API key revoked successfully');
+					} catch (e) {
+						axiosErrorToast(e);
+					}
+				}
+			}
+		});
+	}
+
+	$effect(() => {
```
stonith404, line 78: The api keys should be fetched on server side in a `+page.server.ts` and then passed as a prop to this component. You can take other lists as a reference.

backend/internal/model/api_key.go
```diff
@@ -0,0 +1,18 @@
+package model
+
+import (
+	"time"
+)
+
+type ApiKey struct {
+	Base
+
+	Name        string
```
stonith404, line 10: For the columns that should sortable `sortable:"true` has to be added. See the other models as a reference.

backend/internal/service/api_key_service.go
```diff
@@ -0,0 +1,98 @@
...
+	if err := s.db.Where("user_id = ?", userID).Order("created_at desc").Find(&apiKeys).Error; err != nil {
+		return nil, err
+	}
+	return apiKeys, nil
+}
+
+func (s *ApiKeyService) CreateApiKey(userID string, input dto.ApiKeyCreateDto) (model.ApiKey, string, error) {
+	// Check if expiration is in the future
+	if !input.ExpiresAt.After(time.Now()) {
+		return model.ApiKey{}, "", errors.New("expiration time must be in the future")
```
stonith404, line 33: All custom errors should be defined in `errors.go`. This makes it easier to return the correct status code and message to the client. 

Please also check the other parts where you are using custom errors.

backend/internal/service/api_key_service.go
```diff
@@ -0,0 +1,98 @@
...
+	hashedKey := utils.CreateSha256Hash(apiKey)
+
+	if err := s.db.Preload("User").Where("key = ? AND enabled = ? AND expires_at > ?",
+		hashedKey, true, time.Now()).Preload("User").First(&key).Error; err != nil {
+
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return model.User{}, errors.New("invalid API key")
+		}
+
+		log.Printf("Database error when validating API key: %v", err)
```
stonith404, line 86: Errors get logged automatically when you return an error,

frontend/src/routes/settings/admin/api-keys/api-key-form.svelte
```diff
@@ -0,0 +1,78 @@
...
+	import FormInput from '$lib/components/form/form-input.svelte';
+	import { Button } from '$lib/components/ui/button';
+	import type { ApiKeyCreate } from '$lib/types/api-key.type';
+	import { createForm } from '$lib/utils/form-util';
+	import { z } from 'zod';
+
+	// Add a function to ensure the date is in the future
+	function ensureFutureDate(date: Date): Date {
+		const now = new Date();
+		return date > now ? date : new Date(now.getTime() + 24 * 60 * 60 * 1000); // Default to tomorrow
```
stonith404, line 11: I would remove this and return an error message with Zod instead when the date is in the past.

stonith404 (changes requested):
Thanks :)

I have left some comments, additionally I noticed that the API keys do not refresh when you revoke or add an API key. 

Also there are some type errors, you can run `npm run check` to see them.

kmendell:
@stonith404 I fixed what you suggested i think. I also made it so the description is not needed, if its not provided it just is a blank string.

kmendell:
@stonith404 I have this https://github.com/pocket-id/website/pull/16 currently but its not using swagger, i can try to work on that later today, ive never used that before, unless you have expereince with it , i can let you do it. Otherwise ill take a look and see what i can come up with :)

kmendell:
Ahh swagger is easy to use, I think i can work on this in a bit :)

kmendell:
@stonith404 I added the swagger comments to the relevant controller files, I skipped the webauthn one as thats kinda of pointless to document in my head but maybe im wrong, i also did every api endpoint just for transparency for the relevant files. Let me know how that looks. Ill update the PR on the docs repo soon here as well.

stonith404 (approved):

PR #295, opened 2025-03-04
feat: add prometheus metrics

### LINKED ISSUE #262: 🚀 Feature: Prometheus Metrics for the Backend
### Feature description

Implement a Prometheus metrics endpoint in the backend service.

### Pitch

As a devops engineer / sysadmin, it would be nice to be able to get metrics from the backend to show some key values:

* Number of API calls
* Rate of calls
* Standard Go metrics

One example use would allow for users of Kubernetes to implement pod scaling on more factors than straight memory / cpu usage and increase observability off the application itself.

While off-the-shelf libraries exist for Go, I'm unsure if any library exist for easy integration into Gin.

nikdoof:
Resolves #262 

This uses the upstream supported mod of `github.com/prometheus/client_golang/prometheus/promhttp`. Additions consists of a middleware and a singular route to handle `/metrics`.

By using the standard `prometheus_client` library we also get a lot of free metrics for the Go internals, the additional middleware records any requests handled by Gin.


kmendell:
Initial glance at this it looks good. I will checkout this PR, and review it more in a bit, is there a specific way you were testing this? I've never used Prometheus , so if there's a simple way i can test im all for it!

kmendell:
Also, would we want this secured by a API key? https://github.com/pocket-id/pocket-id/pull/291? or do other services just have this endpoint open ?

nikdoof:
Metrics are produced to a `/metrics` endpoint, which is called by Prometheus, as long as the URL endpoint is working and returning some valid output then it'll work. As we're using the Prometheus' projects own client library we have some assurances that the format will always be compatible with any service ingesting the data.

In terms of API key security, nothing specifically would restrict its usage this way, but generally metrics endpoints are left unauthenticated as they don't contain any particularly sensitive information. The current Caddy configuration wouldn't route metrics requests by default, but this isn't an issue as people who are using Prometheus would generally arrange for their own route to access the endpoint.

For example, locally hosted they'd call the backend directly, for Docker/Kubernetes i'd ensure that I'd map the port of the backend directly so I can call the metrics endpoint rather than using the Caddy layer.

I understand that this PR/feature would have quite a niche use, but with it only consisting of a single Gin route and a middleware I hope that it wouldn't cause too much of a burden for future maintenance.

kmendell:
Sounds good i will pull and test this today to make sure i can get all the metrics, and then ill talk to stonith to make sure he agrees with the changes :)

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -52,6 +53,7 @@ func initRouter(db *gorm.DB, appConfigService *service.AppConfigService) {
 	r.Use(middleware.NewErrorHandlerMiddleware().Add())
 	r.Use(rateLimitMiddleware.Add(rate.Every(time.Second), 60))
 	r.Use(middleware.NewJwtAuthMiddleware(jwtService, true).Add(false))
+	r.Use(middleware.NewPrometheusMetricsMiddleware().Add())
```
kmendell, line 56: You will need to initialize the Prom Middleware like this: 

```go
	// Setup global middleware
	r.Use(middleware.NewCorsMiddleware().Add())
	r.Use(middleware.NewErrorHandlerMiddleware().Add())
	r.Use(rateLimitMiddleware.Add(rate.Every(time.Second), 60))
	r.Use(middleware.NewPrometheusMetricsMiddleware().Add())
	r.Use(middleware.NewJwtAuthMiddleware(jwtService, true).Add(false))

	job.RegisterLdapJobs(ldapService, appConfigService)
	job.RegisterDbCleanupJobs(db)

	// Initialize middleware for specific routes
	jwtAuthMiddleware := middleware.NewJwtAuthMiddleware(jwtService, false)
	fileSizeLimitMiddleware := middleware.NewFileSizeLimitMiddleware()

	//Regsiter /metrics endpoint for prometheus
	r.Use(jwtAuthMiddleware.Add(false))
	r.GET("/metrics", gin.WrapH(promhttp.Handler()))
```

This ensures the jwt service excludes the /mertics path.

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -79,6 +81,9 @@ func initRouter(db *gorm.DB, appConfigService *service.AppConfigService) {
 	baseGroup := r.Group("/")
 	controller.NewWellKnownController(baseGroup, jwtService)
 
+	// Set up metrics route
+	baseGroup.GET("/metrics", gin.WrapH(promhttp.Handler()))
```
kmendell, line 85: This is no longer needed, with the above changes.

kmendell (changes requested):
With these chnages it should work, in addition to these the caddy files will need to be updated to point /metrics to the backend:

reverse-proxy/Caddyfile
```env
:{$CADDY_PORT:80} {
	reverse_proxy /api/* http://localhost:{$BACKEND_PORT:8080}
	reverse_proxy /.well-known/* http://localhost:{$BACKEND_PORT:8080}
	reverse_proxy /metrics http://localhost:{$BACKEND_PORT:8080}
	reverse_proxy /* http://localhost:{$PORT:3000}
}
```


reverse-proxy/Caddyfile.trust-proxy
```env
:{$CADDY_PORT:80} {
	reverse_proxy /api/* http://localhost:{$BACKEND_PORT:8080} {
		trusted_proxies 0.0.0.0/0
		trusted_proxies ::/0
	}
	reverse_proxy /metrics http://localhost:{$BACKEND_PORT:8080} {
		trusted_proxies 0.0.0.0/0
		trusted_proxies ::/0
	}
	reverse_proxy /.well-known/* http://localhost:{$BACKEND_PORT:8080} {
		trusted_proxies 0.0.0.0/0
		trusted_proxies ::/0
	}
	reverse_proxy /* http://localhost:{$PORT:3000} {
		trusted_proxies 0.0.0.0/0
		trusted_proxies ::/0
	}
}
```

kmendell:
@nikdoof Do my suggestions make sense to you?

nikdoof:
Yes sorry, i've had a few commitments recently so i've not been able to update the PR, i'll get the changes done sometime this weekend.

stonith404:
@nikdoof No worries. 

Regarding the suggested changes by @kmendell with the `Caddyfile`, I wouldn't map the metrics path with Caddy, I would instead just set the metrics path to `/api/metrics`. With this change no additional reverse proxy change is needed.

When #291 gets merged, the metrics endpoint will be protected by an API key.

ItalyPaleAle:
I would suggest looking into support for OpenTelemetry, which is what the observability space is standardizing on. OTel supports metrics, as well as logs and traces. Nothing against Prometheus, but OTel (like it or not) is the likely direction where things are moving to, it works with pretty much every tool, and doesn't have issues such as unauthenticated endpoints.

stonith404:
@nikdoof I have no experience with Prometheus. Would Open Telemetry support replace the need for a Prometheus endpoint?

Because if that's the case I would recommend to use Open Telemetry as this is more generic.

ItalyPaleAle:
Prometheus is for metrics only and adopts a "pull" model. You will need an agent that periodically (usually, every 30-60s) scrapes the metrics from all apps by visiting the /metrics endpoint.

OpenTelemetry uses a "push" model. Apps have the [OTel SDK](https://github.com/open-telemetry/opentelemetry-go) installed and they establish a connection with an OTel collector or forwarder. OTel supports metrics and recently (for Go) also logs and traces.

OTel and Prometheus aren't necessarily mutually-exclusive. In some projects I created (one example: [revaulter](https://github.com/italyPaleAle/revaulter)) I support both Prometheus and OTel ([sample](https://github.com/ItalyPaleAle/revaulter/blob/main/pkg/metrics/metrics.go)).

Aside from how things are exposed, however, I'd suggest looking into using the OTel SDK for collecting metrics regardless. The OTel SDK can also expose a Prometheus-compatible endpoint (see my code linked above).

jfroy:
Definitely makes sense in 2025 to add OTel support rather than just Prometheus metrics, though it is more complex to integrate.

I would also caution putting the metrics endpoint behind `/api` and with possible authentication. It's very unusual for Prom metrics to require auth, and it's also unusual to use a different path than `/metrics`. There are exceptions of course, and it is possible to configure a prom scrape to authenticate, but it would be somewhat surprising to the average cluster/prom operator.

kmendell:
Im marking this PR as a draft till we decide/ get the changes made.

kmendell:
Are you still working on this? or should we close and work on implementing Otel instead?

Issue #63, opened 2024-11-09
🚀 Feature: if you edit as admin some users, show their groups 


PaddyPat:
### Feature description

If I edit a user (user is selected), it would be nice to see all groups and easy add the user to new / other groups .

currently I must search every group and add to every group in groups menu a user

### Pitch

Thanks

kmendell:
This issue is seems a bit redundant but If this feature request gets more likes (👍) and the demand for the feature is there, we will consider implementing it.

kmendell:
@PaddyPat Could you attach a screenshot with just a outline of where you were thinking you would like to see this implemented?

PaddyPat:
If I edit a existing member there is a page with

##  general

..
## profile picture

..
## claim
…

And I think a new headline 

## assigned groups

[ ] group mealie
[ ] group outline
[x] group pingvinshare
[ ] group …

Where you easy and fast see which group is assigned to one user

—
Currently I must open every group and check which user is inside..

kmendell:
This is a better version it uses the same table, as the other forms. 

![Image](https://github.com/user-attachments/assets/33b36305-2bb2-44ba-861a-31f980750d4c)

You can test by using this image: `ghcr.io/kmendell/pocket-id:user-group-form`

PaddyPat:
Thanks, looks fine.
please:

- "more than 10 Groups" Switch (1) doesn't work
- Allow us sorting "FriendlyName" over all tables
- Allow us to change default "10" items to 50? (on all pages also in groups view if you have more than 10 users)

<img width="1150" alt="Image" src="https://github.com/user-attachments/assets/5b219fb7-e3dd-46a9-b3b8-86dce26f75c0" />

kmendell:
Gotcha i'll try to fix the state values of the page nation and the sorting, i'll try to fix that today, and let you know when i have pushed a updated image

kmendell:
@PaddyPat Can you try again with the: `ghcr.io/kmendell/pocket-id:user-group-form` image? 

As far as the default sort count, the rest of the tables are at 10 by default, and i think this is by design. So i kept this one at 10 as well.

PaddyPat:
@kmendell awesome, it's fine, great job, no issues.

kmendell:
> [@kmendell](https://github.com/kmendell) awesome, it's fine, great job, no issues.
> 
> fyi, sorting is not fixed in:
> 
> * .../settings/admin/users
> * .../settings/admin/user-groups
[...]

PR #296, opened 2025-03-04
feat: display groups on the user form/account page

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/63


stonith404:
Your implementation had a lot of logic in the frontend that would be better handled in the backend. I have now moved most of the logic to the backend.

Would you mind adding a test for updating user groups on the users page?

kmendell:
@stonith404 I wasnt sure the best place to put it (backend/frontend) since it was a bit of "redundant" logic. But i see what you did now and that makes much more sense. Ill get working on the tests. :)

stonith404:
@kmendell Yeah, I also had a hard time finding an optimal solution, but I think this should now be a clean solution.

kmendell:
@stonith404 Test has been added, Let me know if that does meet your expectations for it. It pretty much mirrors the exsisting user-group one, just mirrored.

stonith404:
Thank you :)

Issue #298, opened 2025-03-04
Bug: Default Sorting not being honored


kmendell:
> If I open one of the other urls 
> 
> Https://pocketid.Domain.com/settings/admin/users
> 
> Or ..
> 
> .../settings/admin/user-groups
> .../settings/admin/user-groups/6a1asdf-asd-asd-asd-asdasd (users in group)
> .../settings/admin/oidc-clients
> 
> The sorting is wrong. I must first select sort by FriendlyName.
> It sorts fine on your fixed page but not on other urls.
> 
> Maybe it’s easy for you to fix also the other ‚tables‘ 

 _Originally posted by @PaddyPat in [#63](https://github.com/pocket-id/pocket-id/issues/63#issuecomment-2698696533)_

The default sorting is not being honored on all tables. See lines 43-58 (https://github.com/pocket-id/pocket-id/blob/114e31e60e6ad7cf31c9e0c415b83ccd77037e32/frontend/src/routes/settings/admin/users/group-selection.svelte) on how i fixed this in the user page.

PaddyPat:
Sorting is fine in 0.36 on page 
* https://pocketid.domain.com/settings/admin/users 
but if I edit a user on https://pocketid.domain.com/settings/admin/users, the new „User Groups“ is not sorted (AtoZ) https://pocketid.domain.com/settings/admin/users/1b945c73-bad1-4350-…..

--

Sorting is fine on page 
* https://pocketid.domain.com/settings/admin/user-groups
but if I edit a Group to display users, these users are not sorted (AtoZ)
https://pocketid.domain.com/settings/admin/user-groups/45de59ce-6fde-…….
On this page clicking on column "Name"does also not sort

Br

kmendell:
Can you attach a picture of what you are seeing?

PaddyPat:
1) Edit User

<img width="1145" alt="Image" src="https://github.com/user-attachments/assets/aa8da76d-4941-411b-94ae-eda9648a93a8" />
<img width="1125" alt="Image" src="https://github.com/user-attachments/assets/6782f1ca-b606-4d54-bb4d-4c3b9f555cad" />

2) Edit Groups (sorting also ignored by click on "name" row)

<img width="1343" alt="Image" src="https://github.com/user-attachments/assets/83321335-64a8-402d-bdcf-10765cb763fb" />
[...]

PR #303, opened 2025-03-09
fix: sorting not applying on page load

kmendell:
Missed Bug form the latest update sorting not working for the user groups on the user form. 

See comments here: https://github.com/pocket-id/pocket-id/issues/298#issuecomment-2708723234


stonith404:
I'm closing this in favor of 8e344f1151628581b637692a1de0e48e7235a22d as I've rewritten most of the sorting logic.

PR #313, opened 2025-03-11
chore: add Dev Container

nebula-it:
Idea is that this will simplify + standardize dev environment setup. 
Additionally, it'll lower the barrier to entry for new contributors with easy setup and debug options.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,13 +15,16 @@
         "svelte.svelte-vscode"
       ]
     }
-  }
+  },
+  // Use 'postCreateCommand' to run commands after the container is created.
+  // Install npm dependencies for the frontend.
+  "postCreateCommand": "npm install --prefix frontend"
+
+
   // Features to add to the dev container. More info: https://containers.dev/features.
   // "features": {},
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   // "forwardPorts": [],
-  // Use 'postCreateCommand' to run commands after the container is created.
-  // "postCreateCommand": "yarn install",
   // Configure tool-specific properties.
   // "customizations": {},
   // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,29 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+  "name": "pocket-id",
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+  "features": {
+    "ghcr.io/devcontainers/features/go:1": {},
+    "ghcr.io/devcontainers-extra/features/caddy:1": {}
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "golang.go",
+        "svelte.svelte-vscode"
+      ]
+    }
+  }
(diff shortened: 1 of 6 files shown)


kmendell:
Never actually used dev containers before. However, first few things I'm noticing 

1. Add `backend/cmd/__debug_*` to .gitignore, so it doesn't commit those debug files. 

2. How does this get the .env files? When i just tested the backend is starting but the frontend will not start, (i haven't dug into it as im going based on the new comer approach to see if it was easy setup)

nebula-it:
> Add backend/cmd/__debug_* to .gitignore, so it doesn't commit those debug files.

Done

> How does this get the .env files? When i just tested the backend is starting but the frontend will not start, (i haven't dug into it as I'm going based on the new comer approach to see if it was easy setup)

Thanks for catching that, I had the env files copied during my initial testing so didnt realize this. [This](https://github.com/pocket-id/pocket-id/pull/313/commits/b72920cf450914b36725c23621567845dfe8b38c) should take care of that.

> I'm assuming the container does not install the npm dependencies? If not, then we will probably want to add that to the contributing guide as well so people know all the steps to setup the environment.

Yep good catch, fixed this as well now.

Give it a shot and let me know :)

kmendell:
Okay i got it working, however caddy does not bind to port 80 (for obvious reasons since its a privileged port) mine bound to port 51147 so if you have a workaround for that i think that would be good to include in the docs. @stonith404 I will leave the final approval for this to you.

nebula-it:
Is that inside the container? So you mind sharing a screenshot of Ports tab (next to Terminal/Debug Console). I use WSL (Windows Subsystem for Linux) and it works fine for me, inside the container Caddy always binds on 80, then on host if 80 is available VS Code forward that to container, if not then it forwards 81 (or next available port)

kmendell:
Systems can only bind to ports under 1024 if its running elevated (root or admin) so in theory it would work in WSL if you are running as root, or sudo but since i didnt launch vs code or docker as root on my macbook it cant bind to it outside the container. Im logging off for the night, but stonith may have some more feedback on this as well.

nebula-it:
Yeap, docker runs as root by default so thats why I was curious. If its not running as root, that makes sense.
Sounds good and thank you :)

stonith404 (approved):
Thanks, that's a good idea :)

kmendell:
@nebula-it When i try this on windows, Im not able to access it at all, even though services are running. And caddy is bound to port 80. Inside the container it says caddy is only listening on tcp6 not that it matters i dont think... Any ideas on why? im just doing final testing before  i merge it in, and in the event someone is on windows id like it to work out of the box. :)

nebula-it:
Can you please double check what the port number is in `Forwarded Address`. Try clicking on the globe icon there.
image

kmendell:
image

Screenshot attached. I know the container works, as it does on macOS , but just wan to sort out the windows side, and update the docs to make sure stuff is caught.

nebula-it:
Yeah that looks good. From powershell on Windows can you please run `Test-Connection 127.0.0.1 -tc 80`

kmendell:
It replied back with `True`

nebula-it:
So Caddy is listening on that port, what do you get when you open the browser?

kmendell:
It just sits and spins, like its loading, but never does, i dont see logs of it even hitting caddy or the front or backend either. I ran that command for ports 80, 3000, and 8080 and they all say true. So im not sure if its a me thing, or what :D I belive it works on windows i just wanted to test to make sure...

nebula-it:
Hah yeah that is interesting. Anything in the Backend debug logs? FYI: It does it take few seconds initially for me, but thats probably just app startup and once its up its quick (2-3s).
Happy to jump on zoom etc to troubleshoot together.

nebula-it:
Alrighty, I just tested on another Windows box and it worked as expected for me.

kmendell:
Yeah i think something is just messy with mine. thank you for this though!

PR #318, opened 2025-03-11
Add Let's Encrypt Certs with Cloudflare DNS

taslabs-net:
## Features

- **Automatic SSL Certificates**: Uses Let's Encrypt via Cloudflare DNS validation
- **Modern Security Configuration**:
  - TLS 1.2/1.3 only
  - Strong cipher suites
  - OCSP stapling
  - HTTP/2 support
- **WebSocket Ready**: Built-in proxy configuration for WebSocket support
- **Automatic Redirects**: Forces HTTPS and handles port redirection
- **Cloudflare Integration**: Securely stores API credentials

### DIFF
--- a/installwithpocketid.sh
+++ b/installwithpocketid.sh
@@ -1,0 +1,195 @@
+#!/bin/bash
+# CloudNginx Installer with WebSocket Support and Persistence
+
+# Configuration
+BLUE='\033[0;34m'
+NC='\033[0m'
+DEFAULT_PORT="443"
+CLOUDFLARE_CRED_PATH="/etc/letsencrypt/cloudflare.ini"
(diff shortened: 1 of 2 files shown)


taslabs-net:
As an added point, if you take out TLS1.2 you have a FIPS 140-2 Compliant Certificate.  If you asked during install "Do you want to link to a Cloudflare Domain?" You could build this directly into the install.  It uses Certbot which automatically renews the cert.  I hope it's useful, I love Pocket ID !

taslabs-net:
> @taslabs-net Instead of a PR on this repo. Can you create it here: https://github.com/pocket-id/website , but no need to add teh script i was more over referring to just the doumentaion that you outline to get it working.

I suck.  Will do

Issue #319, opened 2025-03-11
🚀 Feature: Refresh Tokens


savely-krasovsky:
### Feature description

OpenID Connect Refresh Tokens from the spec

### Pitch

While I like Pocket ID a lot, I see that a lot of services/apps are actively trying to issue refresh tokens, but they can't since Pocket ID does not support them. Many SPA backends cannot issue cookie with own lifetime, but rather fully rely on access/refresh token mechanics, since they are basically advised to follow native applications best security practices. Source: https://www.ietf.org/archive/id/draft-ietf-oauth-browser-based-apps-24.html#name-browser-based-oauth-20-clie

I can probably try to contribute this feature, but I am not sure the project is ready/want to accept it. Maybe this is a principled position to keep project as simple as possible.

kmendell:
@stonith404 and i have talked about this from another issue (https://github.com/pocket-id/pocket-id/issues/250#issuecomment-2709808617). I think we were waiting on clarification for it. Though we will look into it more. It not a 'mandatory' item in the OIDC spec which if i had to guess is why it was not implemented. Ill see stonith has to say about it.

savely-krasovsky:
Yep, I understand that refresh tokens are fully optional by spec, but they are largely used in the wild. I see that Passkeys are much easier way to authorize, but it's still not convenient when for example every OAuth2 Proxy resources asks me literally every hour to authorize again.

PR #325, opened 2025-03-13
feat: add OIDC refresh_token support

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/319


backend/internal/service/oidc_service.go
```diff
@@ -145,60 +145,112 @@ func (s *OidcService) IsUserGroupAllowedToAuthorize(user model.User, client mode
...
+		// Verify that the refresh token belongs to the provided client
+		if storedRefreshToken.ClientID != clientID {
+			return "", "", "", 0, &common.OidcInvalidRefreshTokenError{}
+		}
+
+		// Get user claims
+		// Note: We're intentionally not generating a new ID token here
+		// The OAuth 2.0 spec typically doesn't include refreshing ID tokens
+
+		// userClaims, err := s.GetUserClaimsForClient(storedRefreshToken.UserID, clientID)
```
stonith404, line 230: I would remove this commented out code.

stonith404 (changes requested):
Thank you, it generally looks good to me. Here some things I have noticed:

---
https://openid.net/specs/openid-connect-core-1_0.html#RefreshingAccessToken

> the Client MUST authenticate to the Token Endpoint using the authentication method registered for its client_id

Which means that the client secret has to be validated if the client isn't a public client. 

---

Please make sure to also update the docs when you update a handler, in this case you would have to update the docs of the `createTokensHandler`.

---

Please add a job that clears the expired refresh tokens.

kmendell:
@stonith404 Let me know if that looks better :)

stonith404 (approved):
Looks good to me, thanks :)

kmendell:
Oh good catch! I forgot as well! Yeah i could get it working it kept failing to resolve the response or something like that , Ill take another stab at it as well...

stonith404 (changes requested):
This should be ready to merge as soon as we have the tests.

backend/internal/service/oidc_service.go
```diff
@@ -145,60 +145,121 @@ func (s *OidcService) IsUserGroupAllowedToAuthorize(user model.User, client mode
...
+
+			err := bcrypt.CompareHashAndPassword([]byte(client.Secret), []byte(clientSecret))
+			if err != nil {
+				return "", "", "", 0, &common.OidcClientSecretInvalidError{}
+			}
+		}
+
+		// Verify refresh token
+		var storedRefreshToken model.OidcRefreshToken
+		if err := s.db.Preload("User").Where("token = ? AND expires_at > ?", refreshToken, datatype.DateTime(time.Now())).First(&storedRefreshToken).Error; err != nil {
```
ItalyPaleAle, line 232: Should the refresh token be hashed when stored in the database? Since it's essentially equivalent to a password, at least for a short-ish period of time.

backend/internal/service/oidc_service.go
```diff
@@ -145,60 +145,121 @@ func (s *OidcService) IsUserGroupAllowedToAuthorize(user model.User, client mode
...
+
+		accessToken, err := s.jwtService.GenerateOauthAccessToken(authorizationCodeMetaData.User, clientID)
+
+		s.db.Delete(&authorizationCodeMetaData)
+
+		return idToken, accessToken, refreshToken, 3600, nil
 	}
 
-	accessToken, err := s.jwtService.GenerateOauthAccessToken(authorizationCodeMetaData.User, clientID)
+	if grantType == "refresh_token" {
```
ItalyPaleAle, line 207: Nit: suggest splitting this into 2 separate functions for the 2 grant types. With these 2 `if`s, it's harder to read

backend/internal/service/oidc_service.go
```diff
@@ -145,60 +145,121 @@ func (s *OidcService) IsUserGroupAllowedToAuthorize(user model.User, client mode
...
-	s.db.Delete(&authorizationCodeMetaData)
+		// Generate a new refresh token and invalidate the old one
+		newRefreshToken, err := s.createRefreshToken(clientID, storedRefreshToken.UserID, storedRefreshToken.Scope)
+		if err != nil {
+			return "", "", "", 0, err
+		}
 
-	return idToken, accessToken, nil
+		// Delete the used refresh token
+		s.db.Delete(&storedRefreshToken)
```
ItalyPaleAle, line 257: I think it's good practice to not invalidate the refresh token right away.

I have personally seen situations where there can be concurrent requests to get a new access token with a refresh token. For example, a user could have multiple tabs open for the same app, which would all, in background, try and refresh the token. Or the user could be on an unstable link, where sometimes requests don't succeed (but still hit the server).

For example, Okta allows a grace period before refresh tokens are rotated: https://developer.okta.com/docs/guides/refresh-tokens/main/#grace-period-for-token-rotation Even just a 30s grace period may help

ItalyPaleAle: The OIDC spec, AFAIK, doesn't say anything about refresh tokens being single-used.

This is a suggestion, but I have had bugs in web apps caused by multiple, parallel requests to refresh the access token at the same time. The first one succeeds, and the other ones fail.

The link from Okta above explains this better than I can!

stonith404 (approved):

PR #327, opened 2025-03-13
feat: Allow setting path where keys are stored + only store private key

ItalyPaleAle:
Currently, it's possible to configure paths used by pocket-id to store the SQLite DB, uploads, and the GeoDB data. However, keys were stored in the hardcoded path `data/keys`.

This PR introduces the env var `KEYS_PATH` which allows overriding where keys are stored. Its default value is `data/keys`.

Additionally, as I was implementing this, I realized that pocket-id was writing to disk both the private and public keys, and loading them independently. This was both unnecessary, since the public key can always be derived from the private one, and potentially a bit dangerous, as users could modify the public key on disk and pocket-id was not checking to make sure the two keys were related to each other.

PS: Also included a performance optimization where the ID of the key is pre-computed once and saved in the object, rather than re-computed every time, since the computation is expensive and the value is static


kmendell:
While i see the benefits of some of this change, what is the real purpose of needed to define a place to store the keys? This is not something technically that should be accessible that easily, and this could introduce the ability to mount a volume to this and just access the private key? id just like to understand it better. As this was the first thing that came to mind when i saw the env config part.

ItalyPaleAle:
Yes it's precisely about volumes.

I want to run pocket-id with `securityContext.readOnlyRootFileSystem` (`--read-only` if using Docker) enabled for security reasons (so for example the JS scripts can't be overwritten). To do that, I needed to mount a few volumes (I'm documenting that and will share that later), and one of them was for where the keys were stored.

ItalyPaleAle:
I'm not sure I agree with your last comment. The official docker-compose.yaml already includes this:

https://github.com/pocket-id/pocket-id/blob/f65997e85b9543e098ad6a728be1a475e217718d/docker-compose.yml#L9

This mounts the entire folder `/app/backend/data` as a volume, so keys are by default exposed on the host. If anything, being able to move them to a separate volume (which may not even be mounted on the host) like this PR allows should increase security in this case.

kmendell:
I forgot the whole data folder is mounted. It been a long day :D So with that being posed, why wouldn't that volume work for what you need? in theory doesn't that do what you are wanting? Again just trying to understand more i dont use podman, one of the core concepts of pocket-id is to keep it simple but keep security intact so hence why om just trying to get a better idea if this will help all users, or only a small subset if that makes sense.

ItalyPaleAle:
It's nice to be able to have different mountpoints. Going back to your previous point, this would allow mounting the `keys` subfolder inside a Docker/Podman volume that is not mounted on the host.

backend/internal/service/jwt_service.go
```diff
@@ -252,26 +241,24 @@ func (s *JwtService) generateKeyID(publicKey *rsa.PublicKey) (string, error) {
...
-		return errors.New("failed to generate private key: " + err.Error())
+		return fmt.Errorf("failed to generate private key: %w", err)
 	}
-	s.PrivateKey = privateKey
 
-	if err := s.savePEMKey(privateKeyPath, x509.MarshalPKCS1PrivateKey(privateKey), "RSA PRIVATE KEY"); err != nil {
-		return err
+	err = s.SetKey(privateKey)
+	if err != nil {
+		return fmt.Errorf("failed to set private key: %w", err)
```
stonith404, line 257:
```suggestion
```

This shouldn't be necessary as you're already setting the private key on line 81.

stonith404 (changes requested):
Thanks for your contribution :) 

Please not my suggestions.

ItalyPaleAle:
Thanks for the review, addressed your comments!

stonith404:
Great, thanks!

stonith404 (approved):

stonith404:
@ItalyPaleAle Would you mind to add the env variable `KEYS_PATH ` to the docs website?

ItalyPaleAle:
> @ItalyPaleAle Would you mind to add the env variable `KEYS_PATH ` to the docs website?

Yes I can do it tonight, thanks for the merge!

PR #332, opened 2025-03-14
feat(passkeys): name new passkeys based on agguids

kmendell:
This is a QoL feature, but instead of just "New Passkey" always being the default name, this PR introduces a AAGUID list that will name the passkey according to the aaguid for the password manager, or platform authenticator. 

I forked the list i found: https://github.com/kmendell/passkey-authenticator-aaguids/tree/main, there may be a better way to implement the list, but i didn't think we would want to load that big of a json file all at once, so i just picked some common ones.

### DIFF
--- a/backend/internal/utils/aaguid_util.go
+++ b/backend/internal/utils/aaguid_util.go
@@ -1,0 +1,119 @@
+package utils
+
+import "fmt"
+
+var AAGUIDMap = map[string]string{
+	// Bitwarden
+	"d548826e-79b4-db40-a3d8-11116f7e8349": "Bitwarden",
+
+	// Google
+	"ea9b8d66-4d01-1d21-3ce4-b6b48cb575d4": "Google Password Manager",
+	"a5921718-1094-d74c-93df-5582eea98e5c": "Google Password Manager",
+	"8e0bf8a9-3a4a-4b0b-966e-2daa78c05f19": "Google Password Manager",
+
+	// Windows Hello
+	"6028b017-b1d4-4c02-b4b3-afcdafc96bb2": "Windows Hello",
+	"9ddd1817-af5a-4672-a2b9-3e3dd95000a9": "Windows Hello",
+	"08987058-cadc-4b81-b6e1-30de50dcbe96": "Windows Hello",
+	"6e96969e-a5cf-440d-acea-c73edf8372ef": "Windows Hello",
+	"73019266-b7c3-46dd-ba3f-3c1f769a8af3": "Windows Hello",
+	"f8a011f3-8c0a-4d15-8006-17111f9edc7d": "Windows Hello",
+	"c1f9a0bc-1dd2-404a-b27f-8e29047a43fd": "Windows Hello",
+	"53414d53-554e-4700-0000-000000000000": "Windows Hello",
+	"4cb9c8df-743a-4767-a70f-a0be4375ee88": "Windows Hello",
+
+	// 1Password
+	"98b62f56-8a06-425b-85dd-49d6ca4eaab2": "1Password",
+
+	// Dashlane
+	"1e1ef5e6-cf33-4949-88c0-2ee55b2d8024": "Dashlane",
+
+	// Apple
+	"2fc0579f-8113-47ea-b116-bb5a8db9202a": "Apple",
+	"c5ef55ff-ad9a-4b9f-b580-adebafe026d0": "Apple",
+
+	// Safari
+	"adce0002-35bc-c60a-648b-0b25f1f05503": "Safari",
+
+	// Passkeys.com
+	"cf03be56-5b65-4228-9b38-3a023eb81d4a": "Passkeys.com",
+
+	// LastPass
+	"7c0903ab-d6c7-46dd-a25c-08d44c756f7c": "LastPass",
+
+	// Chrome
+	"a9139ba2-ad26-4367-9e67-3efebeabd979": "Chrome",
+	"aba949cc-b70e-486a-87d9-d677b4c71a53": "Chrome",
+
+	// Firefox
+	"08987058-cadc-4b81-b6e1-30de50dcbe95": "Firefox",
+	"cd443203-676d-45a5-a8a2-a21bd2f69e59": "Firefox",
+
+	// YubiKey
+	"dd4ec289-e01d-41c9-bb89-70fa845d4bf2": "YubiKey",
+	"ee041b16-0296-4cd7-9a47-83c29b1a0d48": "YubiKey",
+	"2fc0579f-8113-47ea-b116-bb5a8db9202d": "YubiKey",
+	"cb69481e-8ff7-4039-93ec-0a2729a154a8": "YubiKey",
+	"5dcd53ef-5e17-4cd3-8296-37d48ce35f57": "YubiKey",
+	"fa2b99dc-9e39-4257-8f92-4a30d23c4118": "YubiKey",
+	"ee882879-721c-4913-9775-3dfcce97072a": "YubiKey",
+	"c1f9a0bc-1dd2-404a-b27f-8e29047a43f3": "YubiKey",
+
+	// Microsoft Authenticator
+	"73bb0cd4-e502-49b8-9c6f-b59445bf720b": "Microsoft Authenticator",
+
+	// Duo
+	"ee041b16-0296-4cd7-9a47-83c29b1a0d49": "Duo",
+
+	// Keeper
+	"9c7dd906-68fd-4b76-8efd-da19f6c8c5ff": "Keeper",
+	"b0458839-dd3a-4458-a516-6c30c84955b6": "Keeper",
+	"b93fd961-f2e6-462f-b122-82002247de78": "Keeper",
+
+	// Enpass
+	"154b9b58-9a88-466b-851e-7ca3b65169a4": "Enpass",
+
+	// Brave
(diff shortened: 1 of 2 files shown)


savely-krasovsky:
FIDO MDS is the source of truth in that case. I would cache it and use for that exact case and some future ones (for example allowing to use only FIDO certified authenticators).

savely-krasovsky:
There is a useful MDS explorer btw: https://opotonniee.github.io/fido-mds-explorer/

kmendell:
@savely-krasovsky  yes i found that as well. However i dont know if caching every single aaguid is feasible, as that adds time and effort, im not sure but im waiting for stoniths thoughts on this to see what he thinks :)

.github/workflows/update-aaguids.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Fetch JSON data
+        run: |
+          curl -o data.json https://raw.githubusercontent.com/passkeydeveloper/passkey-authenticator-aaguids/refs/heads/main/combined_aaguid.json
+
+      - name: Process JSON data
+        run: |
+          mkdir -p backend/resources
+          jq 'map_values(del(.icon_light, .icon_dark))' data.json > backend/resources/aaguids.json
+
+      - name: Commit changes
```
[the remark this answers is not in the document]
ItalyPaleAle: See also the warning here: https://github.com/passkeydeveloper/passkey-authenticator-aaguids?tab=readme-ov-file#passkey-provider-aaguids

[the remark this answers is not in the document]
kmendell: @ItalyPaleAle Yes. We are aware of the warning. I marked it as ready, as i want to make sure he likes this logic. Tweaks will be made before merging most likely, I have been talking to elias on this, we decide to use this list as it combines everything, the FIDO one doesn't have password manages (not all at least). We have been having messages back and forth on this. What test are you referring to? I tested it locally and the output looks like this: 
image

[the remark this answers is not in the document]
ItalyPaleAle: Was suggesting creating a unit test for the utils/aaguid package and then running "go test" on it before committing. This helps validating the JSON won't break builds.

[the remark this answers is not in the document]
kmendell: What do you see that you think would break? Im not against it but the logic is fairly simple, and not sure if it justify tests in my current mindset.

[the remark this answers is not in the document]
kmendell: I added a test file for it.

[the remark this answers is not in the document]
ItalyPaleAle: The concern was that the JSON file fetched from the remote repo may be corrupted. I've seen things like that happen :)

.github/workflows/update-aaguids.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install jq
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
```
ItalyPaleAle, line 19: jq should be pre-installed on the Actions runtime: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

.github/workflows/update-aaguids.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Fetch JSON data
+        run: |
+          curl -o data.json https://raw.githubusercontent.com/passkeydeveloper/passkey-authenticator-aaguids/refs/heads/main/combined_aaguid.json
+
+      - name: Process JSON data
+        run: |
+          mkdir -p backend/resources
+          jq 'map_values(del(.icon_light, .icon_dark))' data.json > backend/resources/aaguids.json
+
```
ItalyPaleAle, line 29: Add unit tests. Because GitHub doesn't run Actions in response to a commit triggered by another action, this should be done before the commit


```suggestion

      - uses: actions/setup-go@v5
        with:
          go-version-file: 'backend/go.mod'
          cache-dependency-path: 'backend/go.sum'
      - name: Install dependencies
        working-directory: backend
        run: |
          go get ./...
      - name: Run backend unit tests
        working-directory: backend
        run: |
          go test -v ./internal/utils/...

```

backend/internal/utils/aaguid_util.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// If exactly 16 bytes, format as UUID
+	if len(aaguid) == 16 {
+		return fmt.Sprintf("%x-%x-%x-%x-%x",
+			aaguid[0:4], aaguid[4:6], aaguid[6:8], aaguid[8:10], aaguid[10:16])
+	}
+
+	// Otherwise just return as hex
+	return fmt.Sprintf("%x", aaguid)
```
ItalyPaleAle, line 36: [`hex.EncodeToString`](https://pkg.go.dev/encoding/hex#EncodeToString) causes less allocations on the heap than fmt.Sprintf

backend/internal/utils/aaguid_util.go
```diff
@@ -0,0 +1,85 @@
...
+func GetAuthenticatorName(aaguid []byte) string {
+	aaguidStr := FormatAAGUID(aaguid)
+	if aaguidStr == "" {
+		return ""
+	}
+
+	// Then check JSON-sourced map
+	aaguidMapOnce.Do(loadAAGUIDsFromFile)
+
+	aaguidMapMu.RLock()
```
ItalyPaleAle, line 49: I don't think you need a mutex if you're using `sync.Once`, because `sync.Once` blocks until it's done

BTW consider whether loading the contents of the JSON into its own variable, rather than `fs.FS`, may be beneficial. They could be embedded directly inside a variable, then are loaded once the app starts, and you don't need to worry with synchronization or reading from a FS

.github/workflows/update-aaguids.yml
```diff
@@ -0,0 +1,36 @@
...
+          sudo apt-get install -y jq
+
+      - name: Fetch JSON data
+        run: |
+          curl -o data.json https://raw.githubusercontent.com/passkeydeveloper/passkey-authenticator-aaguids/refs/heads/main/combined_aaguid.json
+
+      - name: Process JSON data
+        run: |
+          mkdir -p backend/resources
+          jq 'map_values(del(.icon_light, .icon_dark))' data.json > backend/resources/aaguids.json
```
ItalyPaleAle, line 28: Use `-c` to minify the output. Since the JSON will be embedded in the binary, we don't care if it's pretty-printed

```suggestion
          jq -c 'map_values(del(.icon_light, .icon_dark))' data.json > backend/resources/aaguids.json
```

ItalyPaleAle (commented):
Some comments from me :)

backend/internal/utils/aaguid_util.go
kmendell: The file needs to use FS , as the github action is updatng the embedded file. I dont think loading the 700 line file into a variable is a good idea.

.github/workflows/update-aaguids.yml
kmendell: I mean sure inside the binary but the file still exists in the code base, so in the event we ever need to reference it i think leaving it un-minified is fine.

.github/workflows/update-aaguids.yml
kmendell: I dont think we need to run the unit tests here. What purpose does it have? i plan on either adding a check to see if its null, or i am waiting to hear from elias about making our own fork of it.

backend/internal/utils/aaguid_util.go
ItalyPaleAle: It is the same result, since in both cases the embedded content is built into the binary. The difference is just how they are accessed: using `embed.FS` uses a virtual FS, while otherwise it's available in a variable. The kernel will take care of loading/unloading the pages from memory as needed.

https://forum.golangbridge.org/t/performance-boost-from-embed/29992/2

.github/workflows/update-aaguids.yml
ItalyPaleAle: But you are loading additional, unnecessary bytes into the executable, which are kept inside memory pages. So this does have an impact at runtime too.

backend/internal/utils/aaguid_util.go
ItalyPaleAle: Thinking about it, I agree that loading a 700-line JSON file in-memory may not be great (with or without virtual FS, it's still loaded in memory).

A better approach could be to use [Go source generation](https://go.dev/blog/generate) to convert the JSON into a Go map **at compile time**. This should use far less memory, since JSON is very verbose.

.github/workflows/update-aaguids.yml
kmendell: I think have a more robust check would be better, or having our own list.

backend/internal/utils/aaguid_util.go
kmendell: Cool, i didnt know that, ill look into updating that as well.

.github/workflows/update-aaguids.yml
kmendell: This is no longer a issue, i createds a custom repo with the same logic as the aaguid repo but this way we can make sure the list does not go away.

stonith404 (approved):

PR #337, opened 2025-03-14
fix: pocket-id notify emails are considered as medium spam by rspamd

### LINKED ISSUE #330: 🐛 Bug Report: pocket-id notify emails are considered as medium spam by rspamd
### Reproduction steps

I am receving my notification emails on a domain that uses rspamd and the mails end up with a score of 7 due to missing standard headers.

To reproduce the issue, create an user with a target email address that uses rspamd and check the spam level of the received mail, it should not be more than 0.x, but is about 7

### Expected behavior

The email should contain the following headers
MIME-Version
Date
Message-ID

Also it should not QP-Encode email names if they consist of only ASCII characters

### Actual Behavior

The email is missing the headers, most mail servers do not add Date or Message-ID when the mail is not generated by the local UI or a local tool

### Version and Environment

version v39
local mail server is postfix, the recipient uses postfix+rspamd

### Log Output

_No response_

alexlehm:
add additional mail headers and do not encode realnames when they are only 7bit ascii chars

Fixes: #330

### DIFF
--- a/backend/internal/service/email_service.go
+++ b/backend/internal/service/email_service.go
@@ -92,14 +92,20 @@ func SendEmail[V any](srv *EmailService, toEmail email.Address, template email.T
 
 	// to create a message-id, we need the FQDN of the sending server, but that may be a docker hostname or localhost
 	// so we use the domain of the from address instead (the same as Thunderbird does)
+	// if the address does not have an @ (which would be unusual), we use hostname
 
 	from_address := srv.appConfigService.DbConfig.SmtpFrom.Value
 	domain := ""
 	if strings.Contains(from_address, "@") {
 		domain = strings.Split(from_address, "@")[1]
 	} else {
-		// this is completely wrong, but what else could we do?
-		domain = from_address
+		hostname, err := os.Hostname()
+		if err != nil {
+			// can that happen? we just give up
+			return fmt.Errorf("failed to get own hostname: %w", err)
+		} else {
+			domain = hostname
+		}
 	}
 	c.AddHeader("Message-ID", "<" + uuid.New().String() + "@" + domain + ">")
 
--- a/backend/internal/service/email_service.go
+++ b/backend/internal/service/email_service.go
@@ -18,6 +18,8 @@ import (
 	"os"
 	ttemplate "text/template"
 	"time"
+	"github.com/google/uuid"
+	"strings"
 )
 
 type EmailService struct {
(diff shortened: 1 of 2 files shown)


kmendell:
After analyzing both the before and after email messages, Both have the message-id header, and mime-version, so that leads me to believe this is something the mail server should/can do? i don't know much about mail servers this deep though.

alexlehm:
the problem is that the mail server that forwards the mail can add the header but does not have to
e.g. gmail cleans up the excess quoted printable fields so that this is redudant, but postfix does not and the message-id and date is optional as well

kmendell:
Ahh gotcha.

kmendell (approved):
Looks good to me. @stonith404 Let me know if you're okay with this as well.

backend/internal/service/email_service.go
```diff
@@ -84,6 +86,23 @@ func SendEmail[V any](srv *EmailService, toEmail email.Address, template email.T
...
+	// to create a message-id, we need the FQDN of the sending server, but that may be a docker hostname or localhost
+	// so we use the domain of the from address instead (the same as Thunderbird does)
+
+	from_address := srv.appConfigService.DbConfig.SmtpFrom.Value
+	domain := ""
+	if strings.Contains(from_address, "@") {
+		domain = strings.Split(from_address, "@")[1]
+	} else {
+		// this is completely wrong, but what else could we do?
+		domain = from_address
```
stonith404, line 102: I would just not add the header if the domain can't be discovered.

stonith404 (changes requested):

backend/internal/service/email_service.go
alexlehm: I have considered that as well, however it is a common practice in some mail sending services to use a local hostname instead, e.g. Javamail does that when the fqdn cannot be discovered, so it ends up being xxx@webserver17, i would guess a random incorrect hostname is better than no message-id

alexlehm: or we could use the hostname returned by the Go function gethostname, that would be more consistent, i will implement that later

alexlehm:
when the email does not have a domain, it uses something like Message-ID: <e5f9face-8fb1-46bd-b855-db9acca45db4@Windows10>

stonith404 (approved):

stonith404:
Thanks!

Issue #322, opened 2025-03-13
🚀 Feature: Make key type & size configurable, and support EC keys


ItalyPaleAle:
### Feature description

Currently, pocket-id signs JWTs with RSA 2048 keys, which are not configurable. While RSA with a key size of 2048 is still acceptable until 2030 [per NIST's guidelines](https://www.keylength.com/en/4/), there are other algorithms that are considered better options.

It would be great if pocket-id allowed configuring the type (algorithm) and size of the key. For example, allowing using RSA keys with 3072 bits, or allowing elliptic curve signing algorithms including ECDSA (JWT alg `ES256`) or EdDSA (JWT alg `EdDSA`, e.g. with `crv: Ed25519`).

For more information, this is a good read: https://www.scottbrady.io/jose/jwts-which-signing-algorithm-should-i-use

### Pitch

See above :)

kmendell:
@stonith404  Ill have to defer to you on this one.

kmendell:
Until stonith404 replies about his thoughts on this, i am adding the Need more upvotes label.

ItalyPaleAle:
Thanks, I understand your point.

May I suggest an option that would allow supporting EC without adding user-facing complexity?

Pocket ID could detect the algorithm of the private key and use the corresponding JWS algorithm. This would only impact people who update the private key manually.

There are quite significant benefits to using EC, such as much smaller signatures (so much smaller JWTs).

stonith404:
Thanks, I actually didn't know that.

Is EC widely supported by JWT libraries? If yes,  we could just replace the RSA with ECDSA, right?

ItalyPaleAle:
I don't think it's as universally supported by JWT libraries. Going by what I see here: https://jwt.io/libraries some libraries only support RSA.

Solutions like Auth0 let you choose if you want RSA or ECDSA for each app (OAuth client). That may be much more complexity than you'd be willing to accept here.

stonith404:
Okay, I see. If the changes are only small (which I believe they are), you can create a PR if you would like to.

ItalyPaleAle:
@stonith404 how would you feel about me changing the format of the (private) key file to a JWK (JSON) instead of PCKS#1 (PEM)? This would allow being more flexible with different kinds of keys. It would also contain additional info such as the key ID, so it can be computed just once.

Note this is just about the format used to encode the private key on disk, using JWK (so JSON) instead of PKCS#1

[...]

PR #339, opened 2025-03-15
feat: store keys as JWK on disk

ItalyPaleAle:
Part of #322

This converts the format of keys stored on disk to JWK and it allows reading keys containing RSA, ECDSA, or EdDSA keys (note: while non-RSA keys can be read, at this stage I don't believe they can be used to sign/verify JWTs... that will be a subsequent PR)

If a PEM-encoded key is found on disk, it is converted to JWK upon startup.

Lastly, completes a perf improvement started with #327: the JWKS endpoint now returns data that is always cached, making it significantly faster (and reducing the amount of heap allocations on every invocation)


backend/internal/service/jwt_service.go
```diff
@@ -6,92 +6,164 @@ import (
...
+	"github.com/pocket-id/pocket-id/backend/internal/utils"
 )
 
 const (
-	privateKeyFile = "jwt_private_key.pem"
+	privateKeyFilePem = "jwt_private_key.pem"
+	privateKeyFileJwk = "jwt_private_key.json"
+
+	// Size, in bits, of the RSA key to generate if none is found
+	rsaKeySize = 2048
```
ItalyPaleAle, line 34: Kept it at 2048 for now. It can be increased to 3072 in a subsequent PR if you wish.

backend/go.mod
```diff
@@ -18,9 +18,11 @@ require (
 	github.com/golang-migrate/migrate/v4 v4.18.2
 	github.com/google/uuid v1.6.0
 	github.com/joho/godotenv v1.5.1
+	github.com/lestrrat-go/jwx/v3 v3.0.0-alpha3
```
ItalyPaleAle, line 21: Using github.com/lestrrat-go/jwx which is a very complete JWx library, used by other large projects too (we had implemented in in Dapr too). In here I'm using it only for JWK support.

backend/internal/service/test_service.go
```diff
@@ -304,38 +304,9 @@ func (s *TestService) ResetAppConfig() error {
...
...b9dfA-WkhSWY7lURQP4yIBTMU0PF_rEATuS7lRciYk1SOx5fqXZd3m_LP0vpBC4Ujlq6NAq6CIjCnms","dq":"TtUVGCCkPNgfOLmkYXu7dxxUCV5kB01-xAEK2OY0n0pG8vfDophH4_D_ZC7nvJ8J9uDhs_3JStexq1lIvaWtG99RNTChIEDzpdn6GH9yaVcb_eB4uJjrNm64FhF8PGCCwxA-xMCZMaARKwhMB2_IOMkxUbWboL3gnhJ2rDO_QO0","e":"AQAB","kid":"8uHDw3M6rf8","kty":"RSA","n":"yaeEL0VKoPBXIAaWXsUgmu05lAvEIIdJn0FX9lHh4JE5UY9B83C5sCNdhs9iSWzpeP11EVjWp8i3Yv2CF7c7u50BXnVBGtxpZpFC-585UXacoJ0chUmarL9GRFJcM1nPHBTFu68aRrn1rIKNHUkNaaxFo0NFGl_4EDDTO8HwawTjwkPoQlRzeByhlvGPVvwgB3Fn93B8QJ_cZhXKxJvjjrC_8Pk76heC_ntEMru71Ix77BoC3j2TuyiN7m9RNBW8BU5q6lKoIdvIeZfTFLzi37iufyfvMrJTixp9zhNB1NxlLCeOZl2MXegtiGqd2H3cbAyqoOiv9ihUWTfXj7SxJw","p":"_Yylc9e07CKdqNRD2EosMC2mrhrEa9j5oY_l00Qyy4-jmCA59Q9viyqvveRo0U7cRvFA5BWgWN6GGLh1DG3X-QBqVr0dnk3uzbobb55RYUXyPLuBZI2q6w2oasbiDwPdY7KpkVv_H-bpITQlyDvO8hhucA6rUV7F6KTQVz8M3Ms","q":"y5p3hch-7jJ21TkAhp_Vk1fLCAuD4tbErwQs2of9ja8sB4iJOs5Wn6HD3P7Mc8Plye7qaLHvzc8I5g0tPKWvC0DPd_FLPXiWwMVAzee3NUX_oGeJNOQp11y1w_KqdO9qZqHSEPZ3NcFL_SZMFgggxhM1uzRiPzsVN0lnD_6prZU","qi":"2Grt6uXHm61ji3xSdkBWNtUnj19vS1-7rFJp5SoYztVQVThf_W52BAiXKBdYZDRVoItC_VS2NvAOjeJjhYO_xQ_q3hK7MdtuXfEPpLnyXKkmWo3lrJ26wbeF6l05LexCkI7ShsOuSt-dsyaTJTszuKDIA6YOfWvfo3aVZmlWRaI","use":"sig"}`
```
ItalyPaleAle, line 307: Note this key is the same as the old one, but re-encoded in a JWK instead of PEM. This makes writing the tests a lot easier, or I'd need to also expose methods to add the key ID, etc

kmendell:
The tests are failing see below for the error if you havent alread ydebugged this: 

```
2025/03/15 00:44:20 Failed to initialize jwt service: failed to save private key file at path 'data/keys/jwt_private_key.json': failed to create key file: open data/keys/jwt_private_key.json: no such file or directory
{"level":"error","ts":1741999479.3790631,"logger":"http.log.error","msg":"dial tcp [::1]:8080: connect: connection refused","request":{"remote_ip":"192.168.65.1","remote_port":"33184","client_ip":"192.168.65.1","proto":"HTTP/1.1","method":"POST","host":"localhost","uri":"/api/test/reset","headers":{"Accept":["application/json, text/plain, */*"],"Content-Type":["application/x-www-form-urlencoded"],"User-Agent":["axios/1.8.2"],"Accept-Encoding":["gzip, compress, deflate, br"],"Connection":["keep-alive"],"Content-Length":["0"]}},"duration":0.003169458,"status":502,"err_id":"xywts2mt5","err_trace":"reverseproxy.statusError (reverseproxy.go:1269)"}
```

kmendell:
I am looking to find a fix for that locally midst other things, if i do ill push it.

backend/internal/service/jwt_service.go
```diff
@@ -204,78 +276,194 @@ func (s *JwtService) VerifyIdToken(tokenString string) (*jwt.RegisteredClaims, e
...
+	// Set other required fields
+	_ = key.Set(jwk.KeyUsageKey, keyUsageSigning)
+	ensureAlgInKey(key)
+
+	return key, err
+}
+
+// saveKeyJWK saves a JWK to a file
+func saveKeyJWK(key jwk.Key, path string) error {
 	keyFile, err := os.Create(path)
```
kmendell, line 412: To fix the tests, I got this to work by adding the below code in: 

```go
	dir := filepath.Dir(path)
	if err := os.MkdirAll(dir, 0755); err != nil {
		return fmt.Errorf("failed to create directory for key file: %w", err)
	}
```

however there were more errors after that. 

```
[GIN] 2025/03/15 - 00:55:12 | 500 |    3.616209ms |    192.168.65.1 | POST     "/api/webauthn/login/finish"
Error #01: key is of invalid type: RSA sign expects *rsa.PrivateKey
[GIN] 2025/03/15 - 00:55:12 | 500 |    3.647167ms |    192.168.65.1 | POST     "/api/webauthn/login/finish"
Error #01: key is of invalid type: RSA sign expects *rsa.PrivateKey
```

kmendell (changes requested):
See Comment above

backend/internal/service/jwt_service.go
ItalyPaleAle: Thanks. I saw the test failure but didn't have time to debug. I'll apply this tomorrow and investigate more.

ItalyPaleAle: Tests are green! 🟢

stonith404:
Thank you for your contribution. To be honest, I don't think adding a library solely for saving the keys as JWK instead of the default supported PEM format makes much sense.

If this were a small change, it might be justifiable, but with this implementation, we would need to add a library and convert the keys from JWK to PEM every time we sign or verify a JWT.

The only real advantage would be the ability to use other key algorithms. While this would be beneficial, as I mentioned, adding a library and increasing complexity doesn't seem worthwhile for such a minor advantage.

ItalyPaleAle:
> we would need to add a library and convert the keys from JWK to PEM every time we sign or verify a JWT.

That is not quite how it works.

Operation on keys are performed on objects that implement [`crypto.PrivateKey`](https://pkg.go.dev/crypto#PrivateKey) or [`crypto.PublicKey`](https://pkg.go.dev/crypto#PublicKey). These structures in memory contain just a few properties: in the case of RSA, it's [3 numbers](https://pkg.go.dev/crypto/rsa#PrivateKey).

When keys are persisted to disk, they need to be serialized using some format. The most common one is [ASN.1](https://en.wikipedia.org/wiki/ASN.1), using formats such as PKCS#1 or PKCS#8 for private keys. The ASN.1-encoded data is binary, so it's normally encoded to PEM for ease of transmission.

JWK is an alternative format to encoding keys. Rather than using ASN.1 and PEM, it uses JSON.

So it is not true that keys must be converted from JWK to PEM to perform operations. This PR implements support for both, so when a key is read from JWK **or** PEM, it is loaded in memory in a `crypto.PrivateKey` object. That object happens to be wrapped in a `jwk.PrivateKey`, which is exposed by the `jwx` library.

**That said...** I agree that adding a 2nd library for this can add some more work. It is possible to do all of the above without the jwx library, just specifying a few JSON structures, but the jwx library adds some convenience methods.

One option that I could implement in the next PR is to use the jwx library for signing/verifying JWTs too, which would allow using a single library for all JWx things? (I'd do this in a next PR as this is already fairly large) FWIW, jwx is a pretty solid library, well-tested, and a few years ago I consolidated all JWx operations to that for [Dapr](https://github.com/dapr/dapr) too, and it's been great.

stonith404:
@ItalyPaleAle Okay thanks for the detailed explanation, I didn't know how the different key formats are handled. 

I think it would be a good idea to migrate all the JWT logic to jwx in a future PR.

Regarding the migration of the old key format. I think this would make more sense if we add a new package called `migration` or add a new file called `migration.go` to the `bootstrap` package for handling the conversion from PEM to JWK. With that we could decouple migration logic from the actual logic. In my opinion it makes the code complicated if we have migration logic inside the services. What do you think?

ItalyPaleAle:
I can do that, but not sure it'd make the code simpler. The current migration logic is here:

https://github.com/pocket-id/pocket-id/pull/339/files#diff-d9fba0ccb17d9425693d601a78bbf45326ef7c887954aab3a5601b492a2d6d17R96-R115

We hit that code when there's no JWK file on disk.

1. If there's a PEM file on disk, reads that
2. If not, generates a new key
3. In both cases, it then saves the key to disk as JWK

If I moved that to a separate package I'd have to have the logic that writes the JWK to disk twice, one for migrations and one for newly-generated keys.

Maybe I can move the entire code that reads from disk (JWKs and PEMs) and generates keys to a separate service?

stonith404:
This is a suggestion for the migration function. It's just an example, it might contain some errors.
```go
func migratePEMToJWK() error {
	pemPath := filepath.Join(common.EnvConfig.KeysPath, "jwt_private_key.pem")
	
	// If the file does not exist, we don't need to do anything
	if _, err := os.Stat(pemPath); os.IsNotExist(err) {
		return nil
	}
	
	// Load the key from disk and parse it
	data, err := os.ReadFile(pemPath)
	if err != nil {
		return fmt.Errorf("failed to read key data: %w", err)
	}

	key, err := jwk.ParseKey(data, jwk.WithPEM(true))
	if err != nil {
		return fmt.Errorf("failed to parse key: %w", err)
	}

	// Populate the key ID using the "legacy" algorithm
	pubKey, err := key.PublicKey()
	if err != nil {
		return fmt.Errorf("failed to get public key: %w", err)
	}
	var pubKeyRaw any
	err = jwk.Export(pubKey, &pubKeyRaw)
	if err != nil {
		return fmt.Errorf("failed to export public key: %w", err)
	}
	pubASN1, err := x509.MarshalPKIXPublicKey(pubKeyRaw)
	if err != nil {
		return fmt.Errorf("failed to marshal public key: %w", err)
	}

	hash := sha256.New()
	hash.Write(pubASN1)
	hashed := hash.Sum(nil)

	shortHash := hashed[:8]

	// Set the key ID
	key.Set(jwk.KeyIDKey, shortHash)

	// Populate other required fields
	_ = key.Set(jwk.KeyUsageKey, service.KeyUsageSigning)
	
	// Save the key to disk (yeah that's redundant)
	keyFile, err := os.OpenFile(service.PrivateKeyFileJwk, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
	if err != nil {
		return fmt.Errorf("failed to create key file: %w", err)
	}
	defer keyFile.Close()
	
	enc := json.NewEncoder(keyFile)
	enc.SetEscapeHTML(false)
	err = enc.Encode(key)
	if err != nil {
		return fmt.Errorf("failed to write key file: %w", err)
	}
	
	// Remove the old key
	err = os.Remove(pemPath)
	if err != nil {
		return fmt.Errorf("failed to remove old key: %w", err)
	}

	return nil
}

```

You're right the JWT saving logic is redundant but I think adding this redundancy is a good compromise when we can completely remove all PEM logic from the JWT service. In my opinion this makes the future maintaining much easier because else we have legacy code in there that is only used for migration (`privateKeyFilePem`, `loadKeyPEM`, `generatePemKeyID` and the migration part).

ItalyPaleAle:
I've pushed the change, will add some additional tests later today

ItalyPaleAle:
All tests are present and passing!

stonith404:
Thanks :)

stonith404 (approved):

kmendell:
@stonith404 This one didnt merge for some reason. I only mention it as i saw you release 0.41.0 im assuming you wanted this to be in it :P

stonith404:
Yeah, I have to create a second release. The merging was blocked because your review state was "request changes", I didn't notice that.

kmendell:
Whoops, My bad! I assumed since you approved it would have gone through...

stonith404:
Yeah I thought the same too 😅

ItalyPaleAle:
If you're not in a rush do you want me to open the next PR (support for non-RSA keys) before the release? I should have it by tomorrow.

kmendell:
> If you're not in a rush do you want me to open the next PR (support for non-RSA keys) before the release? I should have it by tomorrow.

This was released in 0.41.0, but feel free to open up a PR for your next steps.

savely-krasovsky:
I am not sure if it's this PR, but after updating to 0.42.0 from 0.40.1 I cannot authorize on some resources. Example of error from Hoarder: `no valid key found in issuer's jwks_uri for key parameters {"kid":"","alg":"RS256"}`. I just did `podman pull` and restarted pod.

savely-krasovsky:
I rolled back to v0.41.0, everything started to work. Something is definitely wrong. I would revert that MR or hotfix it ASAP.

@ItalyPaleAle

ItalyPaleAle:
I can investigate tonight.

If you still have it running, what do you see on the `/.well-known/jwks.json` endpoint? (you can post that here since it's non-secret)

backend/internal/bootstrap/jwk_migration.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// Save the key as JWK
+	err = service.SaveKeyJWK(key, jwkPath)
+	if err != nil {
+		return fmt.Errorf("failed to save private key file at path '%s': %w", jwkPath, err)
+	}
+
+	// Finally, delete the PEM file
+	err = os.Remove(pemPath)
```
savely-krasovsky, line 70: I also would delete anything tbh. No one will die from storing extra 2 kilobytes, but if this migration will fail, user will lose his key.

ItalyPaleAle:
@savely-krasovsky I cannot repro. Upgrading to 0.42.0 worked for me. Key was upgraded correctly.

backend/internal/bootstrap/jwk_migration.go
ItalyPaleAle: This was requested by @stonith404 , but I wouldn't be too concerned.

1. The key is deleted at the end only after the migration succeeded.
2. Even if the key is lost, it's no biggie. Tokens that were issued would be invalidated, but pocket-id would generate a new key right away so new tokens will be issued without problem.

savely-krasovsky:
@ItalyPaleAle I can stably reproduce it with Hoarder and Open WebUI. Tailscale SSO and OAuth2 Proxy works fine.

ItalyPaleAle:
@savely-krasovsky can you post your `jwks.json` (from `/.well-known/jwks.json`) **before and after** the upgrade?

backend/internal/bootstrap/jwk_migration.go
ItalyPaleAle: However, there's a valid point that deleting the PEM makes downgrading harder

ItalyPaleAle:
Also, try restarting hoarder and open webui. does that fix the issue?

savely-krasovsky:
@ItalyPaleAle I tried to restart them, unfortunately they still cannot authorize me. I think they could store kid somewhere. Does pre-42 PocketID stores key somewhere internally inside container? Can I manually convert it into JSON? I didn't dig Pocket ID internals a lot.

ItalyPaleAle:
If you use the official docker-compose.yaml, the key is stored in a volume mounted on the host. should be in the `data/keys` folder. Same folder is used by 0.42.

They shouldn't store the kid but they may cache the JWKS. Restarting them should make sure the JWKS is not cached.

However, not sure what happened in your case, but in my system the key was migrated **with the same kid** so even if values were cached bye clients, it should have not been an issue.

savely-krasovsky:
Unfortunately I cannot easily restore keys from backup, since I backup entire zvol. But at the first glance it seems pem private key is gone, but I don't know how it works if I just rolled it back.

My volume mount seems fine: https://github.com/savely-krasovsky/homelab/blob/17584ce598ee4420e8ddfc35e567315d0f00751f/configs/containers/systemd/pocket-id.container.tftpl#L32

ItalyPaleAle:
(I use Podman + Quadlet  too :) )

Delete the existing keys:

```sh
podman exec -it pocket-id sh
```

In the container, delete all files `.pem` and `.json` inside `/app/backend/data`.

Then, exit the container. Upgrade to 0.42.0 and restart Pocket ID.

It will generate a new key for you, in JSON format.

Clients (Hoarder etc) may need to be restarted since the key is new.

savely-krasovsky:
Logs from Immich: `[31m[Nest] 12  - [39m03/18/2025, 10:28:47 PM [31m  ERROR[39m [33m[Api:ErrorInterceptor~rmifv7ds][39m [31mUnknown error: RPError: no valid key found in issuer's jwks_uri for key parameters {"kid":"","alg":"RS256"}[39m`

savely-krasovsky:
@ItalyPaleAle I rolled back to v0.41.0, keys war created again, another kid, but it works without the need to restart anything (Immich, Open WebUI, Hoarder).

ItalyPaleAle:
try deleting your cookies for the Pocket ID site. it's possible maybe you're presenting a token that was issued for the old key when it got deleted.

stonith404:
This should be fixed in the latest version with f7e36a422ea6b5327360c9a13308ae408ff7fffe.

savely-krasovsky:
Thank you, indeed it fixed, have just tested.

Issue #342, opened 2025-03-15
🐛 Bug Report: Cache-Control for `/api/users/me/profile-picture.png` is incorrect


cdemi:
### Reproduction steps

I have noticed that `/api/users/me/profile-picture.png` has `cache-control: max-age=14400`

Since this URL is the same for all Pocket ID users, whoever logs in first after the cache is empty, gets his picture set for all users

### Expected behavior

I think either the URL should have a unique identifier per user or else the Cache-Control configuration should be `no-store`

### Actual Behavior

Users are seeing pictures of other users

### Version and Environment

v0.40.0

kmendell:
You can get the profile picture per user by using a similar url to the below one: 

`/api/users<user-uuid-here>/profile-picture.png`

cdemi:
I understand that, but in the login screen and in the user account manager at the top right is a profile picture with that URL. This is why users are seeing other users' profile pictures.


________________________________

cdemi:
I understand that, but in the login screen and in the user account manager at the top right is a profile picture with that URL. I should have made my report clearer.

kmendell:
Ahh okay i think i understand now, I will see what i can find/fix for this, but it may not be till tomorrow or early next week.

stonith404:
Fixed in `v0.40.1`.

PR #351, opened 2025-03-16
fix: avatar in header fix

savely-krasovsky:
~~#331~~ https://github.com/pocket-id/pocket-id/commit/e45d9e970d327a5120ff9fb0c8d42df8af69bb38 introduced a bug with missing avatar in the header, this PR fixes it

PR #353, opened 2025-03-17
fix: update profile picture URL to use dynamic user ID (#342)

cdemi:
This should fix a small bug created by a fix for #342.

The header profile picture now links to the User's ID instead of `me`


kmendell:
Thanks, However https://github.com/pocket-id/pocket-id/pull/351 has already been opened for this :)

PR #359, opened 2025-03-19
Add support for ECDSA and EdDSA keys

### LINKED ISSUE #322: 🚀 Feature: Make key type & size configurable, and support EC keys

ItalyPaleAle:
> In draft until I can complete some additional manual testing

Fixes #322

1. Adds support for using ECDSA (e.g. P-256), EdDSA (e.g. Curve25519) keys, as well as RSA keys of arbitrary size.
   - The important caveat is that Pocket ID **will not** generate these keys. Users will need to provide their own JWK file containing a key in the desired format.
     - TODO @ItalyPaleAle: Create docs for this
1. Completes the switch from golang-jwt to jwx for generating and validating JWTs, to use the same library. This simplifies the code a lot too.

I've added unit tests that validate that the keys work.

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -160,22 +160,28 @@ func (oc *OidcController) createTokensHandler(c *gin.Context) {
 // @Security OAuth2AccessToken
 // @Router /oidc/userinfo [get]
 func (oc *OidcController) userInfoHandler(c *gin.Context) {
-	authHeaderSplit := strings.Split(c.GetHeader("Authorization"), " ")
-	if len(authHeaderSplit) != 2 {
+	_, authToken, ok := strings.Cut(c.GetHeader("Authorization"), " ")
+	if !ok || authToken == "" {
 		c.Error(&common.MissingAccessToken{})
 		return
 	}
 
-	token := authHeaderSplit[1]
-
-	jwtClaims, err := oc.jwtService.VerifyOauthAccessToken(token)
+	token, err := oc.jwtService.VerifyOauthAccessToken(authToken)
 	if err != nil {
 		c.Error(err)
 		return
 	}
-	userID := jwtClaims.Subject
-	clientId := jwtClaims.Audience[0]
-	claims, err := oc.oidcService.GetUserClaimsForClient(userID, clientId)
+	userID, ok := token.Subject()
+	if !ok {
+		c.Error(&common.TokenInvalidError{})
+		return
+	}
+	clientID, ok := token.Audience()
+	if !ok || len(clientID) != 1 {
+		c.Error(&common.TokenInvalidError{})
+		return
+	}
+	claims, err := oc.oidcService.GetUserClaimsForClient(userID, clientID[0])
 	if err != nil {
 		c.Error(err)
 		return
--- a/backend/internal/service/jwt_service.go
+++ b/backend/internal/service/jwt_service.go
@@ -11,17 +11,17 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"slices"
 	"strconv"
 	"time"
 
-	"github.com/golang-jwt/jwt/v5"
 	"github.com/lestrrat-go/jwx/v3/jwa"
 	"github.com/lestrrat-go/jwx/v3/jwk"
+	"github.com/lestrrat-go/jwx/v3/jwt"
 
 	"github.com/pocket-id/pocket-id/backend/internal/common"
 	"github.com/pocket-id/pocket-id/backend/internal/model"
 	"github.com/pocket-id/pocket-id/backend/internal/utils"
+	"github.com/pocket-id/pocket-id/backend/internal/utils/tokenutils"
 )
 
 const (
@@ -37,10 +37,10 @@ const (
 )
 
 type JwtService struct {
-	privateKey       jwk.Key
-	keyId            string
-	appConfigService *AppConfigService
-	jwksEncoded      []byte
+	privateKey      jwk.Key
+	keyId           string
+	sessionDuration time.Duration
+	jwksEncoded     []byte
 }
 
 func NewJwtService(appConfigService *AppConfigService) *JwtService {
@@ -55,15 +55,14 @@ func NewJwtService(appConfigService *AppConfigService) *JwtService {
 }
 
 func (s *JwtService) init(appConfigService *AppConfigService, keysPath string) error {
-	s.appConfigService = appConfigService
+	sessionDurationInMinutes, err := strconv.Atoi(appConfigService.DbConfig.SessionDuration.Value)
+	if err != nil {
+		return fmt.Errorf("invalid value for 'SessionDuration' configuration option: %w", err)
+	}
+	s.sessionDuration = time.Duration(sessionDurationInMinutes) * time.Minute
 
 	// Ensure keys are generated or loaded
 	return s.loadOrGenerateKey(keysPath)
-}
-
-type AccessTokenJWTClaims struct {
-	jwt.RegisteredClaims
-	IsAdmin bool `json:"isAdmin,omitempty"`
 }
 
 // loadOrGenerateKey loads the private key from the given path or generates it if not existing.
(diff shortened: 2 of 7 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -67,6 +67,8 @@ jobs:
           -e APP_ENV=test \
           pocket-id/pocket-id:test
 
+          docker logs -f pocket-id-sqlite &> /tmp/backend.log &
```
ItalyPaleAle, line 70: Chanes to this file allow getting logs from the backend container too, since playwright logs don't have all info sometimes

backend/internal/controller/oidc_controller.go
```diff
@@ -160,22 +160,28 @@ func (oc *OidcController) createTokensHandler(c *gin.Context) {
 // @Security OAuth2AccessToken
 // @Router /oidc/userinfo [get]
 func (oc *OidcController) userInfoHandler(c *gin.Context) {
-	authHeaderSplit := strings.Split(c.GetHeader("Authorization"), " ")
-	if len(authHeaderSplit) != 2 {
+	_, authToken, ok := strings.Cut(c.GetHeader("Authorization"), " ")
```
ItalyPaleAle, line 163: `strings.Split` allocates a slice in the heap and scans the entire string. `strings.Cut` is more efficient when we know there's at most 2 items, since it doesn't cause heap allocations and it stops at the first match

backend/internal/service/jwt_service.go
```diff
@@ -170,133 +169,157 @@ func (s *JwtService) SetKey(privateKey jwk.Key) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse token: %w", err)
 	}
 
-	claims, isValid := token.Claims.(*AccessTokenJWTClaims)
-	if !isValid {
-		return nil, errors.New("can't parse claims")
+	err = jwt.Validate(
+		token,
+		jwt.WithAcceptableSkew(5*time.Minute),
```
ItalyPaleAle, line 211: 5 mins is generally considered industry-standard clock skew. If pocket-id is scaled horizontally, the clock may be slightly off between nodes

frontend/tests/oidc.spec.ts
```diff
@@ -117,7 +117,7 @@ test('End session without id token hint shows confirmation page', async ({ page
...
+		'eyJhbGciOiJSUzI1NiIsImtpZCI6Ijh1SER3M002cmY4IiwidHlwIjoiSldUIn0.eyJhdWQiOiIzNjU0YTc0Ni0zNWQ0LTQzMjEtYWM2MS0wYmRjZmYyYjQwNTUiLCJlbWFpbCI6InRpbS5jb29rQHRlc3QuY29tIiwiZW1haWxfdmVyaWZpZWQiOnRydWUsImV4cCI6MzY5MDAwMDAwMSwiZmFtaWx5X25hbWUiOiJUaW0iLCJnaXZlbl9uYW1lIjoiQ29vayIsImlhdCI6MTY5MDAwMDAwMCwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdCIsIm5hbWUiOiJUaW0gQ29vayIsIm5vbmNlIjoib1cxQTFPNzhHUTE1RDczT3NIRXg3V1FLajdacXZITFp1XzM3bWRYSXFBUSIsInN1YiI6IjRiODlkYzItNjJmYi00NmJmLTlmNWYtYzM0ZjRlYWZlOTNlIn0.eKLkaOjSnMK67jpyQOXTztro7-8p0N1F3rCfmT9PGzAG1YhV6IXp5T9DEHjBAV_roPHY-ibv_afFEau18oI2dTY_yZP8CXZ0AJbUOjACZ_q5NA5UzEF5l2oysa6ATe3ilET7dMQqf6FUBOse5K4Ttg0oTzc-_1WP-Us1cucSeqrNDz8mfM5HvjrPFKeQJq-olQj1b9lL_L8b71LxPY7ItZyWZkjFrU2kt7A1n09ZxbGZszyptJpOHhn4d7_hSWhdkG4-2eWkbeJj64rIJXFLedezZoz8cTZInkArIXeTjfS6Wl8jsxszAtNkQM96T3dQD79Yx93UM-ke-m1hjN2htw';
```
ItalyPaleAle, line 120: This token was expired, so had to update it. New token will last  till 2086

backend/internal/utils/tokenutils/claims.go
kmendell, line 1: Wouldnt this fit better inside of the jwt service? that way any logic from the jwt token is all container in the service? As the files that use this either are the jwt service, or have a jwt service from injection.

backend/internal/service/jwt_service.go
```diff
@@ -37,10 +37,10 @@ const (
 )
 
 type JwtService struct {
-	privateKey       jwk.Key
-	keyId            string
-	appConfigService *AppConfigService
-	jwksEncoded      []byte
+	privateKey      jwk.Key
+	keyId           string
+	sessionDuration time.Duration
```
kmendell, line 42: we would need the appConfigService added back for dependency injection no?

backend/internal/controller/oidc_controller.go
```diff
@@ -160,22 +160,28 @@ func (oc *OidcController) createTokensHandler(c *gin.Context) {
...
 	}
-	userID := jwtClaims.Subject
-	clientId := jwtClaims.Audience[0]
-	claims, err := oc.oidcService.GetUserClaimsForClient(userID, clientId)
+	userID, ok := token.Subject()
+	if !ok {
+		c.Error(&common.TokenInvalidError{})
+		return
+	}
+	clientID, ok := token.Audience()
```
kmendell, line 179: dont we need to select the specific audience? just comparing to the previous logic, i noticed in my oidc test app, the audience is an array vs a string, with your new logic.

kmendell (commented):
Just a glance review so far.

kmendell:
You also mentioned users have to provide there own key? I think the defeats the purpose of the simplicity of setup of pocket id. Is that just how this algo works? or did you just not implment that part? Im no expert with JWT stuff so im more over just thinking out loud.

savely-krasovsky:
At this point we need full-fledged JWK storage. Some services could not support EdDSA validation, that's why things like Keycloak generate a pool of keys, send an array in `id_token_signing_alg_values_supported` field and allow to configure it for every client.

kmendell:
So this key type is not supported for every client? If so we may want to hold off on this maybe? Until all clients would support the protcol , last thing we would want is to not be able to have users sign int o certain services.

savely-krasovsky:
@kmendell we could wait it forever. Some JSON Web stack libraries will never receive support due to various reasons (no maintainer, project doesn't want to spend time on it, app cannot update library easily, etc). I think this PR can be merged, since it's optional and people will need to manually generate those keys. But in future I would generate a set of keys and allowed to specify what kid to use for exact OIDC client. But I am not sure it's in scope of project.

kmendell:
Yeah that would be out of scope of the project i think , though its not 100% my call. We ll see what elias has to say once he is done with his thesis. I havent fully tested this yet, just a overview last night.

ItalyPaleAle:
> You also mentioned users have to provide there own key? I think the defeats the purpose of the simplicity of setup of pocket id. Is that just how this algo works? or did you just not implment that part?

I agree with you, but I am going with what I was told here: https://github.com/pocket-id/pocket-id/issues/322#issuecomment-2721655220

> But in future I would generate a set of keys and allow to specify what kid to use for exact OIDC client. But I am not sure it's in scope of project.

Exactly.

The thinking is that for now this can offer a bridge so people who know all their clients can support ECDSA or EdDSA have an option.

Users could also run multiple, separate instances of Pocket ID, if they need different key types. This is added complexity, but it's on the users.

backend/internal/utils/tokenutils/claims.go
ItalyPaleAle: I'll move it

backend/internal/service/jwt_service.go
ItalyPaleAle: The only property of `appConfigService` that was in use was the session duration. Since that had to be computed from string to time.Duration, it's simpler to do it once during initialization, than on every call

backend/internal/controller/oidc_controller.go
ItalyPaleAle: It's still an array. `clientID` indeed will be an array and I'm checking it below.

Since we generate all tokens, we know audience can have at most 1 element in the array.

kmendell:
Okay , got it. Ultimatly elias will have to review. I dont think adding a selecting of keys is in the scope of the project , as that adds confusion and complexity, where pocket-id aims to be simple.

backend/internal/controller/oidc_controller.go
kmendell: Yeah i see that, all iwas saying is its showing as this:

`["xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"]` instead of just the raw string:  xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx

ItalyPaleAle: I'm not following, sorry.

You're saying that previously tokens (what kind? access tokens? ID tokens?) had the `aud` claim as an array, and now it's just a raw string?

backend/internal/service/jwt_service.go
kmendell: Pocket ID isnt designed to be muti-node. Does this chnage only effect that or is there another purpose?

backend/internal/controller/oidc_controller.go
kmendell: The other way around , your logic returns an array (first example) the original logic returns just the raw string from the array

backend/internal/service/jwt_service.go
ItalyPaleAle: Primarily for multi-node, but clocks can get out of sync even on the same node, especially in virtual environments. I can reduce it to 1 min?

Btw it does work as multi-node, as long as you use Postgres :)

backend/internal/controller/oidc_controller.go
ItalyPaleAle: I see. It must be a difference in the how the jwx library works.

The RFC states that the `aud` claim is _normally_ an array, but _MAY_ be a string if there's a single value: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3

I will look into changing it back to a string. An array is likely more compliant with the RFC, but don't want to break apps that were used to it being a string

kmendell: It works either way , i was just pointing out what i noticed :)

backend/internal/service/jwt_service.go
ItalyPaleAle: Reduced to 1 min

backend/internal/controller/oidc_controller.go
ItalyPaleAle: Changed it so "aud" is a string now in the tokens

stonith404:
Thanks, I'll look into this as soon as possible. It might take a bit longer because I'm currently very busy with school.

backend/internal/service/jwt_service.go
```diff
@@ -55,17 +61,16 @@ func NewJwtService(appConfigService *AppConfigService) *JwtService {
 }
 
 func (s *JwtService) init(appConfigService *AppConfigService, keysPath string) error {
-	s.appConfigService = appConfigService
+	sessionDurationInMinutes, err := strconv.Atoi(appConfigService.DbConfig.SessionDuration.Value)
```
stonith404, line 64: As the session duration can be changed at runtime, this can't be done only on startup.

backend/internal/service/jwt_service.go
```diff
@@ -170,133 +175,169 @@ func (s *JwtService) SetKey(privateKey jwk.Key) error {
...
+		jwt.WithKey(alg, s.privateKey),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse token: %w", err)
 	}
 
-	claims, isValid := token.Claims.(*jwt.RegisteredClaims)
-	if !isValid {
-		return nil, errors.New("can't parse claims")
+	err = jwt.Validate(
```
stonith404, line 282: This function should return true if the token is expired because this function is used in the end session handler and the OIDC spec states that the ID token can be accepted if it's expired.

I think it would make sense to add a `ignoreExpiration` parameter though to not mistakenly expect that the expiration gets validated when we use this function on another place in the future.

frontend/tests/oidc.spec.ts
stonith404: The id token can be expired so I would keep the old one.

stonith404 (changes requested):

backend/internal/service/jwt_service.go
ItalyPaleAle: Gotcha, I will change it

backend/internal/service/jwt_service.go
ItalyPaleAle: Got it, I'll have to special-case this

ItalyPaleAle: Done

backend/internal/service/jwt_service.go
ItalyPaleAle: Done

backend/go.sum
```diff
@@ -78,8 +78,8 @@ github.com/goccy/go-json v0.10.4 h1:JSwxQzIqKfmFX1swYPpUThQZp/Ka4wzJdK0LWVytLPM=
 github.com/goccy/go-json v0.10.4/go.mod h1:oq7eo15ShAhp70Anwd5lgX2pLfOS3QCiwU/PULtXL6M=
 github.com/gogo/protobuf v1.3.2 h1:Ov1cvc58UF3b5XjBnZv7+opcTcQFZebYjWzi34vdm4Q=
 github.com/gogo/protobuf v1.3.2/go.mod h1:P1XiOD3dCwIKUDQYPy72D8LYyHL2YPYrpS2s69NZV8Q=
-github.com/golang-jwt/jwt/v5 v5.2.2 h1:Rl4B7itRWVtYIHFrSNd7vhTiz9UpLdi6gZhZ3wEeDy8=
-github.com/golang-jwt/jwt/v5 v5.2.2/go.mod h1:pqrtFR0X4osieyHYxtmOUWsAWrfe1Q5UVIyoH402zdk=
+github.com/golang-jwt/jwt/v5 v5.2.1 h1:OuVbFODueb089Lh128TAcimifWaLhJwVflnrgM17wHk=
```
kmendell, line 81: this should be version 5.2.2 as its patched a security issue: e2f74e5687efe11246111c3b2d45d1899abd41d7

ItalyPaleAle: Thanks, that happened by accident while fixing a merge conflict.

stonith404 (approved):
Great, thank you!

PR #365, opened 2025-03-20
fix: Use source generation for managing AAGUID dataset

ItalyPaleAle:
Small improvement over #332 as discussed here: https://github.com/pocket-id/pocket-id/pull/332#discussion_r2002362424

Instead of bundling the JSON file into the compiled binary, for it to be parsed at runtime, this uses source generation ([`go generate`](https://go.dev/blog/generate)) to parse it at build-time and generate a Go file

Makes the code simpler (there's a Go map embedded in the source) as well as faster since there's no need to parse a JSON file at runtime (and synchronize concurrent access).

CC @kmendell


kmendell:
Not trying to be "that guy", but how much time does it save? Not opposed to it but i think only if its a major/decent performance increase.

I cant test benchmarks right now, but this is adding alot of extra code, and im not sure if the performance increase is worth it (again i have not tested).

For example if this goes from 0.7 ns to 0.6 ns is there a point ? Go already is fast, and most of these operations are already running at the Nanosecond level, (theres some exceptions of course).

Hopefully you get where im coming from.

ItalyPaleAle:
It's not so much about speed (although it will be faster because there's no locks used anymore - e.g. sync.Once -, and the first call will be a lot faster since there's no parsing needed), but more about memory usage.

Keeping a JSON in an embedded FS means that the data is kept in the binary as-is, and kept in-memory in a string. Then, when it gets parsed, it gets _also_ added to a map and kept in-memory that way too.

The code is a lot simpler now too, since there's no need to perform synchronizations, nor there's the need to parse a JSON file at runtime.

Thinking about it... I think I could make it even better by changing it to use a function with a `switch` inside, rather than keeping a map, however. This way there won't even be a data structure in the heap.

kmendell:
Sure i get your point, but this code is maybe called on intial setup, and maybe every so often as its only called when adding passkeys, if it was a major consistantly called function i can see this making sense more. 

Does that make sense?

kmendell:
I discussed with @stonith404, We feel this is unnecessary as while it may add some performance increases, it makes the code as a whole alot less manageable. 

Thank you though!

ItalyPaleAle:
You didn't even let me push the change to use a function with a lookup table :)

> while it may add some performance increases, it makes the code as a whole alot less manageable.

Hard disagree on that. Easier to have a Go file with a map or lookup table, which allows using the language server to navigate through that, than an abstract structure that exists only at runtime.
Plus, memory usage should matter.

savely-krasovsky:
I think we need some kind of chat for active contributors. I understand that @stonith404 has his own vision regarding the direction of Pocket ID development, so it would be nice to have a space to preliminarily discuss certain concepts. I understand that we can just create an issue, but that would take more time, especially if the answer ultimately turns out to be simply "no."

kmendell:
@savely-krasovsky There is one on discord, i just have to grant you the role once you join the server

savely-krasovsky:
Where I can find a link to join?

stonith404:
Yeah, I would recommend creating an issue here on GitHub first before creating a pull request. This helps to prevent creating pull requests that could be rejected. 

Small refactorings could also be discussed on Discord and don't need a dedicated issue.

ItalyPaleAle:
Btw this was discussed: https://github.com/pocket-id/pocket-id/pull/332#discussion_r2002362424

ItalyPaleAle:
My branch has been updated with the change to generate a function. I do think there's value in this to reduce memory... https://github.com/ItalyPaleAle/pocket-id/tree/aaguid-source-gen

PR #372, opened 2025-03-21
refactor: use atomic renames for uploaded files

ItalyPaleAle:
This PR fixes a bug in how uploaded files are handled, by using temporary files that are then atomically renamed to the final destination.

This ensures that in case of errors, there is no corrupted file saved on disk. For example, imagine scenarios such as pocket id crashing while writing the file, or a power failure, or more simply interrupted uploads... Because the stream was opened to the original file, only partial data would be written to the file.

The typical solution to this problem is to write to a temporary file first, and then perform a rename. In most OS's, renaming is an atomic operation.

Note that renaming must be done in the same filesystem, so the temporary file needs to be in the same directory. We can't use `/tmp` because that's often on a separate filesystem

PS: Also updates `CreateProfilePicture` to avoid buffering the entire encoded image in-memory


backend/internal/service/user_service.go
```diff
@@ -199,7 +194,7 @@ func (s *UserService) UpdateUser(userID string, updatedUser dto.UserCreateDto, u
 }
 
 func (s *UserService) RequestOneTimeAccessEmail(emailAddress, redirectPath string) error {
-	isDisabled := s.appConfigService.DbConfig.EmailOneTimeAccessEnabled.Value != "true"
+	isDisabled := s.appConfigService.DbConfig.EmailOneTimeAccessEnabled.IsTrue()
```
kmendell, line 197: This should be != true

backend/internal/utils/file_util.go
```diff
@@ -69,14 +75,67 @@ func SaveFile(file *multipart.FileHeader, dst string) error {
...
+	return SaveFileStream(src, dst)
+}
+
+// SaveFileStream saves a stream to a file.
+func SaveFileStream(r io.Reader, dstFileName string) error {
+	// Our strategy is to save to a separate file and then rename it to override the original file
+	// First, get a temp file name that doesn't exist already
+	var tmpFileName string
+	for {
+		suffix := xxhash.Sum64String(dstFileName + strconv.FormatInt(time.Now().UnixMicro(), 10))
```
kmendell, line 87: couldn't we just generate a random string, or use the userID_tmp as a file name? i dont think we need to import a seperate hash library and hash a temp file that's going to be deleted anyways.

backend/internal/utils/image/profile_picture.go
```diff
@@ -3,36 +3,42 @@ package profilepicture
...
 	img = imaging.Fill(img, profilePictureSize, profilePictureSize, imaging.Center, imaging.Lanczos)
 
-	var buf bytes.Buffer
-	err = imaging.Encode(&buf, img, imaging.PNG)
-	if err != nil {
-		return nil, fmt.Errorf("failed to encode image: %v", err)
-	}
-
-	return &buf, nil
+	pr, pw := io.Pipe()
```
kmendell, line 31: This is a more general question, why the change form buffer to pipe?

backend/internal/service/user_service.go
```diff
@@ -376,7 +371,7 @@ func (s *UserService) ResetProfilePicture(userID string) error {
 	}
 
 	// Build path to profile picture
-	profilePicturePath := fmt.Sprintf("%s/profile-pictures/%s.png", common.EnvConfig.UploadPath, userID)
+	profilePicturePath := common.EnvConfig.UploadPath + "/profile-pictures/" + userID + ".png"
```
kmendell, line 374: Are all of these changes just for preference? I normal think Sprintf is cleaner to read then the latter (thats my preference of course) . Aka what im asking is: Is there a purpose to these or not?

backend/internal/model/app_config.go
```diff
@@ -9,6 +13,11 @@ type AppConfigVariable struct {
 	DefaultValue string
 }
 
+func (a *AppConfigVariable) IsTrue() bool {
```
kmendell, line 16: Is this just for convivence? I feel like this adds extra work when it was already checking true or false before. Maybe I'm missing something though.

kmendell (changes requested):
General Comments, there are a few thigs id like clarification on

backend/internal/service/user_service.go
ItalyPaleAle: Yes good catch
```suggestion
	isDisabled := !s.appConfigService.DbConfig.EmailOneTimeAccessEnabled.IsTrue()
```

backend/internal/utils/file_util.go
ItalyPaleAle: > couldn't we just generate a random string

Yes, but that is quite a bit slower than using xxhash, which is a very fast hashing algorithm. Generating a random string requires reading from `/dev/random`, which involves I/O). I can maybe replace xxhash with crc32 / crc64, which is built into the stdlib?

> or use the userID_tmp as a file name

No, because if there are 2 requests in parallel to write to the same file (e.g. user clicks "upload" twice), then they will conflict with each other. Now they will both write to the temp file, and then since renaming is atomic, the last one will win.

backend/internal/utils/image/profile_picture.go
ItalyPaleAle: That is to avoid buffering the entire generated image in-memory.

The current implementation encodes the entire image into a `bytes.Buffer`, which keeps it all in-memory. Then it returns from the function.

The new implementation performs the encoding asynchronously and doesn't buffer the image in-memory. You can see that the encoding is now performed in a goroutine, and the function returns a PipeReader, which is an `io.Reader`. The  goroutine will start writing the encoded image to the pipe only when a reader is attached to the other side. This generates [backpressure](https://medium.com/@jayphelps/backpressure-explained-the-flow-of-data-through-software-2350b3e77ce7) that prevents buffering in-memory.

backend/internal/service/user_service.go
ItalyPaleAle: Yes, in Go, using fmt.Sprintf is slower and most importantly causes allocations on the heap. Best to avoid it when it's for simple string concatenation. https://www.dolthub.com/blog/2024-11-08-sprintf-vs-concat/

backend/internal/model/app_config.go
ItalyPaleAle: Convenience is one part. Correctness another: checking for `== "true"` only matches the exact value. [`strconv.ParseBool`](https://pkg.go.dev/strconv#ParseBool) is more flexible as it also matches `T`, `yes`, `1`, etc, which are common ways to describe a boolean true.

backend/internal/utils/file_util.go
kmendell: I think using a built in hash library may be better i dont think we need to import a extra one just for this one function, if thats makes sense?

kmendell:
Okay i do understand this change, though i need to test it localy when i have time, and its up to @stonith404 if he wants to make the final approval.

kmendell (changes requested):
I tested it locally, and it works as expected. The only change i think should be done is the hash library, unless elias thinks its fine otherwise.

ItalyPaleAle:
Replaced with crc64 which is in the stdlib

kmendell (approved):
@stonith404 Can you do a final review on this?

stonith404 (approved):
Thanks!

Issue #362, opened 2025-03-20
🐛 Bug Report: LDAP: With wrong unique attribute for groups ldap sync creates _local_ groups. 


jose-d:
### Reproduction steps

I installed pocket-id using docker image and connected it to instance of freeIPA.

![Image](https://github.com/user-attachments/assets/537c8d4f-2845-4dde-86b9-e8e8951f7266)

### Expected behavior

Imported groups are shown with source `LDAP`.

### Actual Behavior

Groups imported from LDAP are shown as `local`

![Image](https://github.com/user-attachments/assets/3d80f387-4f13-4e37-9aa5-a0e6f80683ae)

### Version and Environment

0.42.1 in Docker behind haproxy

### Log Output

-

kmendell:
I have never used FreeIPA , Does it differ from LDAP in any way when it comes do search queries? 

Also do you get any error logs load this page?

jose-d:
> Does it differ from LDAP in any way when it comes do search queries?

it is solution around LDAP so should be not different. 
I can provide any `ldapsearch` if needed to verify it.

On every LDAP sync I see this kind of error for all LDAP groups:  ( c4 is ldap group and I never created it manually for sure )

```
...
2025/03/20 13:00:00 /app/backend/internal/service/user_group_service.go:74 duplicated key not allowed
[0.090ms] [rows:0] INSERT INTO `user_groups` (`id`,`created_at`,`friendly_name`,`name`,`ldap_id`) VALUES (?,?,?,?,?)
2025/03/20 13:00:00 Error syncing group c4: name is already in use
...
```

so cannot be that the groups were somehow created as `local` during initial sync?

kmendell:
If its uses the same Logic as LDAP it should work. 

My guess is something is not setting the ldapId value on the groups. If you try to edit a group thats from ldap (dont save it), does it let you or are the values and options greyed out? 

Same thing for users?

jose-d:
>  If you try to edit a group thats from ldap (dont save it), does it let you or are the values and options greyed out?

yes, I am able to check/uncheck the check-boxes with (possible) members in group and type into textbox with groupname

> Same thing for users?

no, when I open particular user in /settings/admin/users , then I can not edit name, it's grayed out.


when I check the `user_groups` sqlite table, it looks like this (i redacted some more sensitive groupnames)


```
sqlite> .schema user_groups
CREATE TABLE user_groups
(
    id           TEXT NOT NULL PRIMARY KEY,
[...]

PR #376, opened 2025-03-21
fix: skip ldap objects without a valid unique id

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/362


stonith404 (approved):
Thanks!

PR #379, opened 2025-03-23
fix: hash the refresh token in the DB (security)

ItalyPaleAle:
Fixup from #325

Makes sure that refresh tokens are stored as hashes in the database for security, as they are comparable to passwords.

Since refresh tokens are random strings of 40 characters (alphanumeric), they have enough entropy and a short-enough lifespan that a simple SHA-256 hash should suffice.


kmendell:
Can you adapt the tests to make sure they work with your new logic?

kmendell (approved):

stonith404 (approved):
Thanks for the refactoring and adding hashing to the refresh token :)

PR #381, opened 2025-03-24
feat: modernize ui

kmendell:
So far this is what I've implemented for this:

- Added Animations for each page. 
- Made the Sidebar sticky and updated the buttons hover state.
- Updated the My Account page to have a more modern yet separated design.
- Updated the Passkey List item to be glassy.
- Added Link back to `/account/settings` for the header logo.
- Slide in animation for Login Wrapper
- Combined Profile Picture into Account Details Card **_(Left old component and code till we confirm we like this look)_**
- Moved the Glassy Row Item to Separate Svelte Component
- Updated All Cards to use the new 'Depth' Design look.
- Updated the version of lucide-svelte to 0.483.0

### Known Issues
- Scrollbars popup on the sub-components during the animations... I fought with this all night and could not figure it out.


kmendell:
@stonith404 Can you take a look at this in its current state and see what you think and provide some feedback? :)

frontend/src/lib/components/login-wrapper.svelte
```diff
@@ -11,59 +12,84 @@
 		children: Snippet;
 		showAlternativeSignInMethodButton?: boolean;
 	} = $props();
+
+	let mounted = $state(false);
+
+	// Determine if we're on the alternative login page
+	let isAlternativePage = $derived(page.url.pathname.includes('/alternative'));
```
stonith404, line 19: I would add a `animate` parameter to the login wrapper instead of relying on the path.

stonith404 (changes requested):
Thanks, I really like it! 

A few notes, let me know if you don't agree:
- I would only show the slide animation of the auth pages on first page load. If you click for example on "Don't have access to your passkey?" the animation plays again which could be a bit annoying.
- I would keep the default shadows, the newly added, more prominent shadows confuse a bit
- A few cards have two borders at the bottom which makes the bottom border thicker than the others
Screenshot 2025-03-25 at 15 51 39@2x
- Although the "New" badge on the passkeys look nice, I think it could confuse the users and I don't really think that it adds a value to see on the first view that the passkey has been added recently.
- I would keep the old spacing between the cards:
Screenshot 2025-03-25 at 15 55 52@2x
- I've noticed on a few places that you only return the UI if the component is mounted. What's the reason for this? Ideally we should avoid this because with that we opt out of server-side rendering.
- On a few places for example on the `Card` component you're assigning classes that you assign everywhere. Instead of adding them everywhere, I would recommend to add them directly to the `Card` component.

Just a idea, I'm not really sure if this is possible. Maybe I can also tackle this if you don't want to try it:
It would be nice if we could add a wrapper for all the `animate-fade-in` animations. This would remove the need of adding the `animate-fade-in` to every Card on every page. Ideally we could then use this component in the layout for the settings pages, then every Card would automatically be animated. Here is an example how it could look like:
```
<FadeInWrapper>
<Card>First item that fades in with a delay of 150ms</Card>
<Card>Second item that fades in with a delay of 200ms</Card>
<Card>Second item that fades in with a delay of 250ms</Card>
</FadeInWrapper>
```

kmendell:
@stonith404 Thank you for the feedback! I will get working on some of those notes.

As far as the onMount , that should only be for the animations to note when the animations shoould stop and are visible, though maybe i implemented it wrong , ill take a look.

The FadeInWrapper, i can for sure try to do this, ill try to fix the rest first. 

And as far as the classes go, I applied them everywhere because i wanst sure if shadcn-svelte gets updated for the card coomponent if it would overwrite the custom changes made or not. If it wont ill apply directly to the card :)

stonith404:
Great, thank you :)

I tried to remove `onMount` on a page and everything worked as expected, but I didn't test it on every page.

> And as far as the classes go, I applied them everywhere because i wanst sure if shadcn-svelte gets updated for the card coomponent if it would overwrite the custom changes made or not. If it wont ill apply directly to the card :)

Yes, you're right. If we update the components, especially when shadcn-svelte 1.0.0 is released, we will have to manually check the changes we've made to the components. However, I think it's worth it because having the same classes on every page can be annoying to maintain. I have already made some changes to the Shadcn components in the past, for example, I added the `isLoading` property to the Button.

kmendell:
I will get most of the fixed and then have you re-review :) Im not sure about the fadeinwrapper though, ill fix everything else then try that:)

kmendell:
@stonith404  In regards to the shawdows, you meant just remove the customs class ones  i added right? and use the default component ones ?

kmendell:
@stonith404 Can you take a look at this now and see if i caught all the changes? I think i got everything, but want to make sure :)

kmendell:
@stonith404 I think with your commit d4abe77f115e6b2835b0dcf8e2bfdcd5622d9cc4 , it may have broke the links for the admin to see, At least they are not showing up when I just pulled this, I will try to look into it but you may beat me too it

kmendell:
Correction, Its only broke on initial load, If the page is refreshed the all the links show up .

stonith404 (approved):

PR #388, opened 2025-03-27
fix: use WAL for SQLite by default and set busy_timeout

ItalyPaleAle:
When using SQLite in a web application, it's common to [enable WAL](https://www.sqlite.org/wal.html) which makes it possible to get better performance especially for concurrent access. WAL is supported by Gorm since under the hood it does use the C sqlite library, so it should be enabled by default.

Additionally, this PR makes sure to set the explicit `_txlock=immediate` in the connection string unless the user set another value (not recommended). It also sets a `busy_timeout` to 2500ms (unless another value is set by the user) to make SQLite wait if there's already a transaction/query that has locked the database - otherwise, SQLite would just error immediately, showing a failure to the user (not a good experimence). (Note that when WAL is enabled, SQLite creates 2 additional files next to the database called `pocket-id.db-shm` and `pocket-id.db-wal`)

One thing to note is that Pocket ID's configuration uses the terminology "database path" for SQLite, assuming that only paths to files can be passed. However, the underlying driver can accept either a path, or a connection string (starting with `file:`), which allows passing additional options as query string args. This is what we rely on.

Most users will not need to change a thing in how they use Pocket ID.

The only situation where one would want to change anything is to disable the WAL if the SQLite DB is running on a networked filesystem, such as NFS or SMB. However, even with the WAL disabled, running SQLite over a networked file system is [strongly discouraged](https://www.sqlite.org/faq.html#q5) by the SQLite developers themselves, as it's risky in all cases.


backend/internal/bootstrap/db_bootstrap.go
```diff
@@ -101,6 +105,84 @@ func connectDatabase() (db *gorm.DB, err error) {
 	return db, err
 }
 
+func parseSqliteConnString(connString string) string {
```
kmendell, line 108: I think we can use GORM for this, We can just add the journal mode to the connection string, and this way we keeping the trend of the dependencies were using vs setting the params manually:

```go
db, err := gorm.Open(sqlite.Open("your_database.db?_journal_mode=WAL"), &gorm.Config{})
```

I think this is easier to read and makes more sense in the context of this project no?

kmendell (changes requested):
I see the point of this , but see my comment. I think that makes more sense to use it that way.

backend/internal/bootstrap/db_bootstrap.go
ItalyPaleAle: I don't recommend that.

First, it is already possible today to pass additional query string parameters, and there are situations where users may want to customize the parameters. Your suggested change would break it.

Second, is a concrete example of a situation where user would want to customize the parameters: when using SQLite on an NFS/SMB share. While that is **not recommended**, it *can be attempted* but it needs the journal to be disabled.

stonith404:
Wouldn't it make more sense to set the default value of `SQLITE_DB_PATH` to a connection string that includes those parameters?

Like you said it would then make sense to rename the environment variable `SQLITE_DB_PATH` to `SQLITE_CONNECTION_STRING`. Or better, we could merge `SQLITE_DB_PATH` and `POSTGRES_CONNECTION_STRING` to `DB_CONNECTION_STRING`.

ItalyPaleAle:
> Wouldn't it make more sense to set the default value of SQLITE_DB_PATH to a connection string that includes those parameters?

The issue I have with that is that if users set their own DB path, they will _also_ need to remember to add some "required" (or at least very, very recommended) args themselves, including `_busy_timeout` and `_journal_mode=WAL`. It may become cause of user error. Ideally, we'd set the defaults as the simplest option.

stonith404:
Yes you're right the user could forget to add these parameters. 

But I think if we have `"file:data/pocket-id.db?_journal_mode=WAL&_busy_timeout=2500"` as the default value the user won't randomly remove those parameters. And if he does, he just wouldn't profit from WAL, which isn't the end of the world.

ItalyPaleAle:
Alright

To make sure we don't surprise users while they upgrade, my suggestion is:

- Keep `SQLITE_DB_PATH` as **deprecated** (remove it from docs but allow it be used if set)
   - If `SQLITE_DB_PATH` is set, use `file:$SQLITE_DB_PATH?_journal_mode=WAL&_busy_timeout=2500` as connection string
- Keep `POSTGRES_CONNECTION_STRING ` as **deprecated** alias for `DB_CONNECTION_STRING` (remove from docs, but keep working)
- Recommend using `DB_CONNECTION_STRING` for Postgres and SQLite

Alternatively, we can keep `POSTGRES_CONNECTION_STRING` and `SQLITE_CONNECTION_STRING` separate.

Does that work?

stonith404:
Yes that makes sense, thanks. 

I'll add to the release notes that `SQLITE_DB_PATH` and `POSTGRES_CONNECTION_STRING` are depreciated and might get removed in the future.

stonith404 (changes requested):

backend/internal/bootstrap/db_bootstrap.go
```diff
@@ -78,9 +80,31 @@ func connectDatabase() (db *gorm.DB, err error) {
 	// Choose the correct database provider
 	switch common.EnvConfig.DbProvider {
 	case common.DbProviderSqlite:
-		dialector = sqlite.Open(common.EnvConfig.SqliteDBPath)
+		// Check if we're using the deprecated SqliteDBPath env var
+		switch {
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

stonith404 (approved):
Thank you.

PR #392, opened 2025-03-27
fix: use transactions when operations involve multiple database queries

ItalyPaleAle:
Currently, the backend does not use database transactions, even when there are operations performed in sequence. For example: https://github.dev/pocket-id/pocket-id/blob/4d049bbe24aa0b714c6f97369f1562064076a3b4/backend/internal/service/user_service.go#L166-L194 where the user is loaded from the DB and then updates

This can cause a number of problems if multiple requests happen in parallel, and in some cases _could_ also have security implications.

This PR updates the backend's to add transactions where they are needed.


ItalyPaleAle:
No worries. Fixed now!

kmendell:
I introduced more after merging the latest linter update in... sorry... these should be easier though .

ItalyPaleAle:
Fixed

kmendell (approved):
From my overview and testing this looks good to me, @stonith404 Theres alot of chnages here so please review to make sure i didnt miss anything as well.

backend/internal/service/api_key_service.go
```diff
@@ -54,50 +58,62 @@ func (s *ApiKeyService) CreateApiKey(userID string, input dto.ApiKeyCreateDto) (
...
 	hashedKey := utils.CreateSha256Hash(apiKey)
 
-	if err := s.db.Preload("User").Where("key = ? AND expires_at > ?",
-		hashedKey, datatype.DateTime(time.Now())).Preload("User").First(&key).Error; err != nil {
-
+	var key model.ApiKey
+	err := s.db.
+		WithContext(ctx).
+		Model(&model.ApiKey{}).
+		Clauses(clause.Returning{}).
```
stonith404, line 102: Didn't know that this is possible, that's a nice change.

backend/internal/service/audit_log_service.go
```diff
@@ -41,33 +42,54 @@ func (s *AuditLogService) Create(event model.AuditLogEvent, ipAddress, userAgent
...
 // CreateNewSignInWithEmail creates a new audit log entry in the database and sends an email if the device hasn't been used before
-func (s *AuditLogService) CreateNewSignInWithEmail(ipAddress, userAgent, userID string) model.AuditLog {
-	createdAuditLog := s.Create(model.AuditLogEventSignIn, ipAddress, userAgent, userID, model.AuditLogData{})
+func (s *AuditLogService) CreateNewSignInWithEmail(ctx context.Context, ipAddress, userAgent, userID string, tx *gorm.DB) model.AuditLog {
+	createdAuditLog := s.Create(ctx, model.AuditLogEventSignIn, ipAddress, userAgent, userID, model.AuditLogData{}, tx)
 
 	// Count the number of times the user has logged in from the same device
 	var count int64
-	err := s.db.Model(&model.AuditLog{}).Where("user_id = ? AND ip_address = ? AND user_agent = ?", userID, ipAddress, userAgent).Count(&count).Error
+	err := tx.
```
stonith404, line 63: Is it really necessary to run this in the same transaction? Couldn't we just run this outside the transaction and then check if the count equals zero?

backend/internal/service/audit_log_service.go
```diff
@@ -41,33 +42,54 @@ func (s *AuditLogService) Create(event model.AuditLogEvent, ipAddress, userAgent
...
 		log.Printf("Failed to count audit logs: %v\n", err)
 		return createdAuditLog
 	}
 
 	// If the user hasn't logged in from the same device before and email notifications are enabled, send an email
 	if s.appConfigService.DbConfig.EmailLoginNotificationEnabled.IsTrue() && count <= 1 {
+		// We use a background context here as this is running in a goroutine
+		//nolint:contextcheck
 		go func() {
+			innerCtx := context.Background()
```
stonith404, line 79: What's the advantage of having a context here?

backend/internal/service/custom_claim_service.go
```diff
@@ -113,39 +128,61 @@ func (s *CustomClaimService) updateCustomClaims(idType idType, value string, cla
...
 	if err != nil {
 		return nil, err
 	}
 
 	return updatedClaims, nil
 }
 
-func (s *CustomClaimService) GetCustomClaimsForUser(userID string) ([]model.CustomClaim, error) {
+func (s *CustomClaimService) GetCustomClaimsForUser(ctx context.Context, userID string, tx *gorm.DB) ([]model.CustomClaim, error) {
 	var customClaims []model.CustomClaim
```
stonith404, line 162: Do we really have to run the query in the same transaction?

backend/internal/service/user_service.go
```diff
@@ -48,13 +49,23 @@ func (s *UserService) ListUsers(searchTerm string, sortedPaginationRequest utils
 	return users, pagination, err
 }
 
-func (s *UserService) GetUser(userID string) (model.User, error) {
+func (s *UserService) GetUser(ctx context.Context, userID string) (model.User, error) {
+	return s.getUserInternal(ctx, userID, s.db)
+}
+
+func (s *UserService) getUserInternal(ctx context.Context, userID string, tx *gorm.DB) (model.User, error) {
```
stonith404, line 56: I'm assuming you've created the internal function to pass the transaction from `UpdateUserGroups` into this function. Is this really necessary can't querying the user run inside another transaction?

stonith404 (commented):
Thanks!

I have some notes:
- I'm not really familiar with contexts in Go. I'm assuming the main advantage is this case is to cancel the query when the HTTP request gets cancelled? If yes, what's the advantage of creating a context in `bootstrap.go` and passing it down to the jobs which aren't called by HTTP?
- I saw that you've created multiple "internal" functions for functions that are reading from the database so that you can pass a transaction to the function. Is this really necessary because it seems like those function don't depend on the state of the transaction. Couldn't we just run these read operations outside the transaction?
- Couldn't we pass the context to the transaction instead of adding it to every query?

Like this
```
tx := s.db.WithContext(ctx).Begin()
...
err := tx.
	Find(...)
...
err := tx.
	Find(...)
...
```
instead of 
```
tx := s.db..Begin()
...
err := tx.
	WithContext(ctx)
	Find(...)
...
err := tx.
	WithContext(ctx)
	Find(...).
...
```

ItalyPaleAle:
> I'm not really familiar with contexts in Go. I'm assuming the main advantage is this case is to cancel the query when the HTTP request gets cancelled?

Correct. You can read more about it here: https://www.digitalocean.com/community/tutorials/how-to-use-contexts-in-go

> If yes, what's the advantage of creating a context in bootstrap.go and passing it down to the jobs which aren't called by HTTP?

It's best practice to make any call that makes external I/O, or does long-running work, have a context to be able to cancel it. Also, a bit of a WIP, but eventually it would be nice to ensure there's support for graceful shutdown, which does require having a context passed down from the `main()` method.

> I saw that you've created multiple "internal" functions for functions that are reading from the database so that you can pass a transaction to the function.

When I created an "internal" function it's because the function could be called either directly (by a REST handler) or by another function, which had its own transaction.

For example, take [`UserService.createUserInternal`](https://github.com/ItalyPaleAle/pocket-id/blob/3d5a05f16ae615a63916c6420589bd300e69622d/backend/internal/service/user_service.go#L225). This one can be called by:

- The handler for POST /api/user. In this case, it only [invokes `CreateUser`](https://github.com/ItalyPaleAle/pocket-id/blob/3d5a05f16ae615a63916c6420589bd300e69622d/backend/internal/controller/user_controller.go#L195) which needs to start its own transaction
- Or, by the LDAP service, which already has a larger transaction so it calls [`createUserInternal`](https://github.com/ItalyPaleAle/pocket-id/blob/3d5a05f16ae615a63916c6420589bd300e69622d/backend/internal/service/ldap_service.go#L287) only - otherwise, we'd be creating nested transactions, which are not what we want here

In the case of `createUserInternal`, however, I can probably remove the starting of the transaction from `CreateUser`. Originally I added it because I saw that `creaeteUserInternal` performs 2 queries, including checking for duplicate fields... But then noticed it does that only in case of errors, so the transaction would be rolled back anyways. So yes, I need to update that so `CreateUser` doesn't start a transaction (but we still need a `createUserInternal` method because the LDAP service has its own transaction)

Not all methods needed this, so I've only done it when strictly necessary. I wanted to avoid putting the logic to create transactions in the controllers.

> Couldn't we pass the context to the transaction instead of adding it to every query?

Good question. I looked at the docs for Gorm and it seems that [may be supported](https://gorm.io/docs/context.html). 

Both should work I think, but the reason why normally you pass a context to each individual query (at least, that's the pattern when not using Gorm) is that it allows setting a timeout on each query. I am not doing it today, but in the future it may be nice to do to ensure that single queries don't block for too long if there's something going on.

backend/internal/service/user_service.go
ItalyPaleAle: I responded in the comment below providing more details.

backend/internal/service/audit_log_service.go
ItalyPaleAle: I responded below. For now it's mostly to make sure all I/O calls have a context, as a good practice (so all queries and the `SendEmail` method have a context). In the future we may find this useful to be able to do graceful shutdown

backend/internal/service/custom_claim_service.go
ItalyPaleAle: This method is only called by `GetCustomClaimsForUserWithUserGroups` which does make multiple queries

backend/internal/service/audit_log_service.go
ItalyPaleAle: This is only called by `VerifyLogin` which does perform multiple queries

stonith404 (approved):

stonith404 (approved):
Looks good to me, thanks :)

Issue #396, opened 2025-03-27
🐛 Bug Report: removed ldap users aren't deleted in pocket id


tupcakes:
### Reproduction steps

I noticed when deleting a user in ldap (using lldap) that the pocket id sync will not delete the user.

### Expected behavior

I wasn't sure if this was intended, but it seems like deleting a user in ldap should remove from pocket id, or at least have a way in the pocket id gui to manually delete the user.

### Actual Behavior

User was not deleted.

### Version and Environment

0.44.0
Environment: k3s
Proxy: ingress-nginx

### Log Output

This is the only event in the log related to the operation:
`2025/03/27 19:43:43 Failed to delete user svcauthentik-sync with: LDAP users can't be updated`

kmendell:
I'll take a look at this. Thank you for reporting!

kmendell:
@tupcakes Can you try this image `ghcr.io/kmendell/pocket-id:delete-ldap-users` and see if the issue is fixed?

tupcakes:
Seems to work. Nice! Thank you.

kmendell:
Fixed in: https://github.com/pocket-id/pocket-id/commit/7e658276f04d08a1f5117796e55d45e310204dab

PR #398, opened 2025-03-27
fix: ldap users arent deleted if removed from ldap server

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/396


ItalyPaleAle (approved):

stonith404:
I think it makes more sense if we handle this similarly to how we handle the user update.

I'm closing this because of 7e658276f04d08a1f5117796e55d45e310204dab.

PR #349, opened 2025-03-16
🚀 Feature: Translation

jonasclaes:
This PR will add translation support using Paraglide v2 (beta).

The order in which a first language is chosen is as follows: (configurable in vite.config)

1. local storage
2. cookie
3. baseLocale (default locale set in `project.inlang/settings.json`, in this case, `en`)

I've currently added English and Dutch (I'm a native Dutch speaker)

Leaving this PR in draft untill I've got it translated. :D

Fixes: https://github.com/pocket-id/pocket-id/issues/174

PR #401, opened 2025-03-28
chore(deps): install inlang plugins from npm

gepbird:
Since https://github.com/pocket-id/pocket-id/pull/349, building the project fetched the inlang plugins at build time (every time you ran `npm run build`), rather than install time (with `npm install`), this PR puts inlang plugins at install time.

I'm packaging pocket-id for nixpkgs where more people will be able to use your project. This has caused me a [few hours of debugging](https://github.com/NixOS/nixpkgs/pull/381867#issuecomment-2746330493), as nix ensures reproducibility by cutting off the internet at build time, and there were no error messages coming from inlang that downloading the plugins have failed, only some warnings from vite.


kmendell (approved):
@stonith404 Can you do the final review on this? seems fine to me :) but up to you.

stonith404:
Strange, I'm not even sure why the default configuration uses `jsdelivr`. Thanks!

PR #402, opened 2025-03-28
refactor: do not include test controller in production builds

ItalyPaleAle:
Currently, the test controller is built into Pocket ID binaries and is enabled every time `APP_ENV` is any value but `production`.

The test controller is very powerful, as it includes a RESTful endpoint that allows nuking the entire database. While it is required for E2E tests, its existence outside of that can be considered a security risk.

This PR uses Go build tags to make sure that the test controller is only built when the application is compiled with `-tag "e2etest"`, which is not included in production builds


backend/internal/controller/e2etest_controller.go
kmendell, line 1: Were these files renamed on purpose? I think e2etest just looks odd, renming back to test_service and test_controller look cleaner and less bulky. 

The same can apply if we seperate the router bootstrpa that can just be test_router_bootstrap that just looks cleaner thats kinda a small thing but i digress.

kmendell (changes requested):

backend/internal/bootstrap/e2etest_router_bootstrap.go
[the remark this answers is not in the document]
ItalyPaleAle: It must be a separate file or we can't exclude it at built-time with this comment at the very start of the file:

https://github.com/pocket-id/pocket-id/pull/402/files#diff-740317ed25c41596830dbfbc6ed8c5b3b4766a6ae23b282f85fa85e097d74dd6R1

See https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags

backend/internal/controller/e2etest_controller.go
ItalyPaleAle: I can rename. While not strictly required, it's best practice to use the name of the build tag (e2etest) in the name of the file (and can't use `test` as build tag because that has a special meaning for unit test files)

kmendell: Oh okay then this is fine, if it needs to match.

backend/internal/bootstrap/e2etest_router_bootstrap.go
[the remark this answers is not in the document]
kmendell: Okay understood.

kmendell (approved):
@stonith404 Final review is up to you :)

stonith404 (approved):
Didn't know that this is possible, thanks.

Issue #395, opened 2025-03-27
🚀 Feature: Provide introspection endpoint


aksdb:
### Feature description

In OAuth2 flows, where the server needs to verify the validity of a token provided by a client, it may be of use to offer token introspection as defined in [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662) and advertise it as `introspection_endpoint` in `/.well-known/openid-configuration`.

### Pitch

An single-page-application could be a public OIDC app acquiring an access token. This is passed to a backend which has its own OIDC app. That backend can then make use of the introspection endpoint, to verify the validity of the token and extract information from it.

An application that uses it is for example [OpenTalk](https://opentalk.eu).

kmendell:
I added the needs-more-upvotes label for now. I want to research this more as well.

Also @stonith404 please provide your input on this.

PR #405, opened 2025-03-30
feat: implement token introspection

aksdb:
This PR implements the token introspection as requested in ticket #395.

It fulfills RFC 7662 as far as I can tell (at least the client that required it is happy with the result).

The endpoint requires a `client_id`/`client_secret` pair for authentication, so no random attackers can use it to fish for valid tokens (although that attack vector is pretty limited currently, since we don't offer any token revocation anyway). Potentially we could also allow API token auth for this endpoint, but for now it should be a good start with clientid/clientsecret already.

It also allows checking for the validity of a `refresh_token`.

Fixes: https://github.com/pocket-id/pocket-id/issues/395


backend/internal/service/oidc_service.go
```diff
@@ -277,6 +277,67 @@ func (s *OidcService) createTokenFromRefreshToken(refreshToken, clientID, client
...
+	}
+
+	if err := bcrypt.CompareHashAndPassword([]byte(client.Secret), []byte(clientSecret)); err != nil {
+		return introspectDto, &common.OidcClientSecretInvalidError{}
+	}
+
+	token, err := s.jwtService.VerifyIdToken(tokenString, false)
+	if err != nil {
+		introspectDto.Active = false
+		return introspectDto, nil
```
kmendell, line 303: This should be 

`return '', err` shouldn't it?

kmendell (changes requested):
I haven't looked fully, but that's one fix should be needed, also we will want to implement a playwright e2e test for this new endpoint as well.

backend/internal/service/oidc_service.go
aksdb: I currently consider the failed token validation to be an "inactive" token. If we ever implement revocation, we might consider separating between "no token" (= bad request), "expired token" (= inactive), "revoked token" (= inactive), otherwise undecodable (= bad request).

However I think practically for the caller it doesn't matter why we consider the token inactive, so I opted for just validating it using the normal means (i.e. checking its signature, expiration, valid-after, etc.).

aksdb:
> I haven't looked fully, but that's one fix should be needed, also we will want to implement a playwright e2e test for this new endpoint as well.

Ah I was wondering where the tests were. I was a bit "shocked" that there were no Go tests. I didn't consider it could be an external test suite. I'll look into this.

backend/internal/service/oidc_service.go
kmendell: Okay got it. I think we should still return the error though ? Im only pointing it out as the go linter is complaining, I haven't fully pulled and tested this either, but if the lint issue is not an issue, then we can ignore it.

backend/internal/service/oidc_service.go
```diff
@@ -277,6 +277,67 @@ func (s *OidcService) createTokenFromRefreshToken(refreshToken, clientID, client
 	return accessToken, newRefreshToken, 3600, nil
 }
 
+func (s *OidcService) IntrospectToken(clientID, clientSecret, tokenString string) (introspectDto dto.OidcIntrospectionResponseDto, err error) {
+	if clientID == "" || clientSecret == "" {
```
ItalyPaleAle, line 281: RFC 7662 says that the introspection endpoint should also support refresh tokens too?

backend/internal/dto/oidc_dto.go
```diff
@@ -73,3 +77,15 @@ type OidcTokenResponseDto struct {
...
+	Active     bool     `json:"active"`
+	Scope      string   `json:"scope,omitempty"`
+	Expiration int64    `json:"exp,omitempty"`
+	IssuedAt   int64    `json:"iat,omitempty"`
+	NotBefore  int64    `json:"nbf,omitempty"`
+	Subject    string   `json:"sub,omitempty"`
+	Audience   []string `json:"aud,omitempty"`
+	Issuer     string   `json:"iss,omitempty"`
+	Identifier string   `json:"jti,omitempty"`
+}
```
ItalyPaleAle, line 91: I wonder if this DTO struct is needed. Couldn't the introspection endpoint just return the list of claims in the JWT, with the added `active` claim? This would allow saving the awkward `if claim, hasClaim...` chain, and it would also include optional claims (which are allowed per the RFC)

aksdb: My intention was to have the DTO for better OpenAPI documentation. But in the end ... true, we could simply return a relatively anonymous map.

Once I have the E2E test in place, I'll refactor this.

backend/internal/service/oidc_service.go
aksdb: True. Actually even a good usecase for this endpoint, since there is no client-side ability to verify the validity of a refresh token.

backend/internal/dto/oidc_dto.go
kmendell: I would keep this if the endpoint is returning this type. That way it gets pulled into the documentation, when generated with swagger.

ItalyPaleAle: I wasn’t considering OpenAPI. Makes sense then

aksdb:
@kmendell I added tests in 410e91e7.

backend/internal/dto/oidc_dto.go
aksdb: I replaced it with a generic loop in 324981dc.

backend/internal/service/oidc_service.go
aksdb: Done in 324981dc.

backend/internal/dto/oidc_dto.go
aksdb: Ok, reverted and adjusted in: 3a8d1226

kmendell:
Ill check this out today at some point and try to do a full review for it. Initial glance looks okay so far though.

backend/internal/service/oidc_service.go
aksdb: I added a comment explaining the error (non-)handling. Conveniently I had to check for a parsing error to have and indication for a potential `refresh_token`, so `golangci-lint` also seems to be happy now.

backend/internal/controller/oidc_controller.go
```diff
@@ -290,6 +290,36 @@ func (oc *OidcController) EndSessionHandlerPost(c *gin.Context) {
...
+// @Summary Introspect OIDC tokens
+// @Description Pass an access_token to verify if it is considered valid.
+// @Tags OIDC
+// @Produce json
+// @Param token formData string true "The token to be introspected."
+// @Success 200 {object} dto.OidcIntrospectionResponseDto "Response with the introspection result."
+// @Router /api/oidc/introspect [post]
+func (oc *OidcController) introspectToken(c *gin.Context) {
+	// Disable cors for this endpoint
+	c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
```
ItalyPaleAle, line 303: I don't think this will work, as CORS normally is determined by preflight requests (HEAD method). You may need to modify the CORS middleware (see the `middlewares` folder)

backend/internal/controller/oidc_controller.go
```diff
@@ -290,6 +290,36 @@ func (oc *OidcController) EndSessionHandlerPost(c *gin.Context) {
...
+
+	var input dto.OidcIntrospectDto
+	if err := c.ShouldBind(&input); err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	// Client id and secret can also be passed over the Authorization header
```
ItalyPaleAle, line 311: can also, or must?

backend/internal/dto/oidc_dto.go
```diff
@@ -55,6 +55,10 @@ type OidcCreateTokensDto struct {
 	RefreshToken string `form:"refresh_token"`
 }
 
+type OidcIntrospectDto struct {
+	Token string `form:"token" binding:"required"`
```
ItalyPaleAle, line 59: Do we need token_hint too?

backend/internal/service/oidc_service.go
```diff
@@ -277,6 +278,93 @@ func (s *OidcService) createTokenFromRefreshToken(refreshToken, clientID, client
...
+
+	if err := bcrypt.CompareHashAndPassword([]byte(client.Secret), []byte(clientSecret)); err != nil {
+		return introspectDto, &common.OidcClientSecretInvalidError{}
+	}
+
+	token, err := s.jwtService.VerifyIdToken(tokenString, false)
+	if err != nil {
+		if errors.Is(err, jwt.ParseError()) {
+			// It's apparently not a valid JWT token, so we check if it's a valid refresh_token.
+			return s.introspectRefreshToken(tokenString)
```
ItalyPaleAle, line 305: I think you could do this in a more heuristic way.

Refresh tokens issues by Pocket ID are fixed at 40-chars and have no dots. So before attempting to parse this as access token, you could simply check if the length is 40 chars and has no dots (`!strings.ContainsRune`) and treat it as refresh token.

Also note the specs have a `token_hint` parameter which can be used by clients to hint if it's an access token or refresh token (not required)

backend/internal/controller/oidc_controller.go
aksdb: Copy paste ... dammit. Will adjust.

backend/internal/service/oidc_service.go
aksdb: True, however the other endpoints that handle refreshTokens also don't apply any checks besides the DB lookup. The only scenario I see where this could hurt us is, if someone deliberately fires random stuff against the `introspect` endpoint. If they wanted to do that, they could as well throw that random stuff against the `token` endpoint, though.

So basically: I opted for keeping the logic unified and didn't implement it differently than in other spots.

backend/internal/middleware/cors.go
```diff
@@ -18,6 +18,8 @@ func (m *CorsMiddleware) Add() gin.HandlerFunc {
 		// Allow all origins for the token endpoint
 		if c.FullPath() == "/api/oidc/token" {
 			c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
+		} else if c.FullPath() == "/api/oidc/introspect" {
+			c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
 		} else {
 			c.Writer.Header().Set("Access-Control-Allow-Origin", common.EnvConfig.AppURL)
 		}
```
ItalyPaleAle, line 25: Nit, for readability

```go
		// Allow all origins for the token and introspection endpoints
		switch c.FullPath() {
		case "/api/oidc/token", "/api/oidc/introspect":
			c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
		default:
			c.Writer.Header().Set("Access-Control-Allow-Origin", common.EnvConfig.AppURL)
		}
```

kmendell: I thought github repoen the one i resolved i was like what the heck. Sure not a huge deal , some people find switch readable other find if statements. I think this is fine to leave for now, as were looking into a revamped cors solution in the future hopefully.

kmendell: Nevermind its complaining, ill switch it

kmendell (approved):
Looks good to me, Thank you! 

@stonith404 As long as you dont see anything major wrong , this should be good to merge.

aksdb:
Almost forgot about the copy&paste error that I was made aware of yesterday; I fixed it in 5c03357a and expanded the comment to give a bit more context on why I implemented the authentication that way.

ItalyPaleAle (approved):

backend/internal/service/oidc_service.go
```diff
@@ -277,6 +278,93 @@ func (s *OidcService) createTokenFromRefreshToken(refreshToken, clientID, client
...
+	// Verify the client secret. This endpoint may not be used by public clients.
+	if client.IsPublic {
+		return introspectDto, &common.OidcClientSecretInvalidError{}
+	}
+
+	if err := bcrypt.CompareHashAndPassword([]byte(client.Secret), []byte(clientSecret)); err != nil {
+		return introspectDto, &common.OidcClientSecretInvalidError{}
+	}
+
+	token, err := s.jwtService.VerifyIdToken(tokenString, false)
```
stonith404, line 301:
```suggestion
	token, err := s.jwtService.VerifyOauthAccessToken(tokenString, false)
```

Shouldn't we verify if it is an access token instead?

aksdb: @stonith404 I adjusted it in commit c925831e. It's a bit confusing that there is `VerifyAccessToken`, `VerifyIdToken` and `VerifyOauthAccessToken`. Looks like `VerifyAccessToken` is strictly meant for access towards PocketId itself. The different between `VerifyIdToken` and `VerifyOauthAccessToken` seems to be mainly that the former allows for less strict validation. But at least semantically `VerifyOauthAccessToken` makes indeed much more sense.

stonith404: @aksdb Thanks. Yes, that's a bit concerning because at first glance, it seems possible that an `id_token` could be successfully validated by `VerifyOauthAccessToken`, which would be very bad. I'll look into it as soon as possible. After that, the PR should be ready to merge.

stonith404 (approved):

Issue #390, opened 2025-03-27
🐛 Bug Report: Default profile picture is not cached


ItalyPaleAle:
### Reproduction steps

When invoking `GET /api/users/{id}/profile-picture.png`, if the user doesn't have a profile picture, the default one is generated.

However, the image is generated on every request, and this is a costly operation.

### Expected behavior

The default profile picture should be cached. Ideally, it's created once when the user is created (or when another picture is deleted).

### Actual Behavior

The image is generated on every request, and this is a costly operation.

### Version and Environment

Current main branch

ItalyPaleAle:
Not a super-high priority, but creating this to make sure I don't forget about it either.

stonith404:
Yeah caching the profile picture probably would make sense. What it makes a bit difficult is that the profile picture needs to be updated when the name of a user changes.

We probably would have to add a column in the database to mark if a user has uploaded a custom profile picture. If not, the profile picture needs to be updated on every name change.

ItalyPaleAle:
There could be another way.

Every time a user is created, or there's a name change, generate the default profile image and store it in a separate folder (or instead of giving it the name of the UUID of the user, name it after the initials, such as "es.png"). We can optionally skip doing that if the user has a personal profile pic (but this isn't strictly necessary). If the image already exists (e.g. there's already "es.png" either because of Elias S. or because of Emanuela S.), you can skip re-generating this (this will need some coordination with locks etc to make sure we aren't generating the same image twice in parallel).

Those images are never deleted. Even if people change names, they can be kept around, so when another user with the same initials appear, the image already exists. This means that there may be a bit of waste, but it's negligible (in the absolute worst case, where you go AA..AB..AC..ZZ, you get 676 images, which is a relatively small number).

kmendell:
I like that approach but i think it may need some tweaking, to make it not overly engineered/complicated.

PR #406, opened 2025-03-30
fix: create reusable default profile pictures

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/390

This PR also include a fix for the Uploading and Resetting of profile pictures on the My Account page: ba6e04aa4a4c4d5602347b24e872e0a715451a0c


backend/internal/service/user_service.go
```diff
@@ -59,28 +60,54 @@ func (s *UserService) GetProfilePicture(userID string) (io.Reader, int64, error)
...
+		// Get the file size
+		fileInfo, err := file.Stat()
+		if err != nil {
+			file.Close()
+			return nil, 0, err
+		}
+		return file, fileInfo.Size(), nil
+	}
+
+	// If no cached default picture exists, create one and save it for future use
```
ItalyPaleAle, line 97: Should do the entire generation in a background goroutine, and ensure that no more than 1 (or GOMAXPROCS) can run at the same time. Otherwise you may have 2 goroutines generating the same image at the same time

ItalyPaleAle (changes requested):

backend/internal/service/user_service.go
kmendell: The migration steps should handle all of them technically, i have 40 users in my ldap server and it generated all of those on start with no issue, unless im missing something in what your saying. It should never try to generate it if it exsists already.

ItalyPaleAle: No, i'm talking about when there are multiple requests to generate images in parallel.

For example, when you create a new Pocket ID and add LDAP. If your directory contains "John Apple" and "Jake Anders", they both are JA. This means that you'll have 2 goroutines creating the same image in parallel. That's why I suggested creating a separate background worker that takes care of this.

I can send you a patch in a bit to show an example!

ItalyPaleAle: Here's a patch.

The idea is that this has a background worker which listens for requests for profile pictures to be generated if they don't exist. Because this uses a single queue, we can make sure that no more than 1 request for the same image is coming in at the same time. Otherwise, it would cause conflicts: for example, when multiple users are being imported from LDAP at the same time.

Commit here: https://github.com/pocket-id/pocket-id/commit/cfca59e2b36b2a784c7e8861d79d45eb6308b49d (built on top of your branch)

You can apply it to your branch with:

```sh
git checkout fix/pfp-cache
curl -LO https://github.com/pocket-id/pocket-id/commit/cfca59e2b36b2a784c7e8861d79d45eb6308b49d.patch
git apply cfca59e2b36b2a784c7e8861d79d45eb6308b49d.patch
```

PS: I also fixed a bug where the file descriptors were never closed after reading profile pictures from disk (e.g. custom profile pictures)

kmendell: I tried your logic and it seems to work the same but in a much more xomplex way. the gocoginit lint score was over 30. Your examples also seem to be over engineering this which is what i was trying to avoid this is a very simple task that shouldnt need its own service to run. 

That being said im not saying my version is perfect but it align better with aim of how we wanted to implement this feature.

stonith404 (approved):
Looks good to me. 

@kmendell FYI I have removed the migration because I think it doesn't make sense to create the default profile pictures for the user as it will get saved anyways if a user requests a profile picture that doesn't exist. Or do I miss something?

Feel free to merge it if you agree with removing the migration.

kmendell:
@stonith404 Yeah that makes sense, i was just trying to be proactive about it but overall your chnages make sense :)

Issue #391, opened 2025-03-27
🐛 Bug Report: Race conditions in App Config


ItalyPaleAle:
### Reproduction steps

Updating App Config using the UI could cause panics (exceptions) in Pocket ID.

The cause is that various handlers, such as `PUT /api/application-configuration` can update the config (in Go, each request is handled by a separate goroutine). Handlers like these call `UpdateAppConfig` which update the config in the DB before reloading it and populating local variables such as `DbConfig`.

Meanwhile, other goroutines (e.g. other handlers) can access values in `DbConfig`.

Because there are no locks or other mechanisms to ensure safe access from multiple goroutines, this can cause race conditions.

### Expected behavior

There should not be race conditions

### Actual Behavior

There can be race conditions

### Version and Environment

main branch

kmendell:
You say it "can" cause panics, have you reproduced this, or just something you noticed it could do? Im just curious.

ItalyPaleAle:
I have not experienced it. Race conditions can be very tough to reproduce manually since they tend to strike "at random".

A simpler way to reproduce data races like these would be to have a Go unit test created for the handlers, and then run it with `go test -race`

PR #407, opened 2025-03-31
fix: DbConfig race conditions

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/391


backend/internal/service/app_config_service.go
```diff
@@ -31,6 +34,14 @@ func NewAppConfigService(db *gorm.DB) *AppConfigService {
 	return service
 }
 
+// GetDbConfig returns a copy of the DbConfig in a thread-safe manner.
+func (s *AppConfigService) GetDbConfig() model.AppConfig {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	configCopy := *s.DbConfig // Create a copy of the struct
```
ItalyPaleAle, line 41: Rather than creating a copy when the data is accessed, which means a copy every time, just create a new object when you write it into `s.DbConfig`.

This is because when you do something like `s.DbConfig = &model.AppConfig{}` (or whatever it is), you are changing what's inside the pointer. The old object can still be accessed safely without problems, as long as you don't modify its content.

Even better, when you think in terms of pointers, you can get rid of mutexes (which are slow) and just use `atomic.Pointer` https://medium.com/better-programming/atomic-pointers-in-go-1-19-cad312f82d5b

backend/internal/service/app_config_service.go
```diff
@@ -17,12 +18,14 @@ import (
 type AppConfigService struct {
 	DbConfig *model.AppConfig
```
ItalyPaleAle, line 19: This would become `*atomic.Pointer[model.AppConfig]`

backend/internal/service/app_config_service.go
```diff
@@ -17,12 +18,14 @@ import (
...
 	db       *gorm.DB
+	mu       sync.RWMutex // Add a mutex
```
ItalyPaleAle, line 21: Remove locks when using atomic.Pointer

backend/internal/service/app_config_service.go
```diff
@@ -17,12 +18,14 @@ import (
...
 }
 
 func NewAppConfigService(db *gorm.DB) *AppConfigService {
 	service := &AppConfigService{
 		DbConfig: &defaultDbConfig,
 		db:       db,
+		mu:       sync.RWMutex{}, // Initialize the mutex
```
ItalyPaleAle, line 28:
```suggestion
		DbConfig: &atomic.Pointer[model.AppConfig]{}
		db:       db,
```

Then set `s.DbConfig.Store(defaultDbConfig)`

backend/internal/service/app_config_service.go
```diff
@@ -31,6 +34,14 @@ func NewAppConfigService(db *gorm.DB) *AppConfigService {
 	return service
 }
 
+// GetDbConfig returns a copy of the DbConfig in a thread-safe manner.
+func (s *AppConfigService) GetDbConfig() model.AppConfig {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	configCopy := *s.DbConfig // Create a copy of the struct
+	return configCopy
```
ItalyPaleAle, line 42:
```suggestion
	return s.DbConfig.Load()
```

backend/internal/service/app_config_service.go
```diff
@@ -362,6 +383,13 @@ func (s *AppConfigService) InitDbConfig() error {
...
 // LoadDbConfigFromDb loads the configuration values from the database into the DbConfig struct.
 func (s *AppConfigService) LoadDbConfigFromDb() error {
+	s.mu.Lock()         // Acquire write lock
+	defer s.mu.Unlock() // Release write lock
+	return s.loadDbConfigFromDbLocked()
+}
+
+// Add an unexported version that doesn't acquire the lock.
+func (s *AppConfigService) loadDbConfigFromDbLocked() error {
 	dbConfigReflectValue := reflect.ValueOf(s.DbConfig).Elem()
```
ItalyPaleAle, line 393: In here you initialize a new *model.AppConfig (maybe that's a clone of the value you load from the pointer), work on that, then when you're done you call `s.DbConfig.Store()`

backend/internal/service/app_config_service.go
kmendell: Okay ill look into it , ive never delt with these very much.

kmendell: I looked into this, and instroducing the atomic.Pointer will break alot of otehr files , not they they cant be fixed, but its a way bigger effort than i think this issue may need, when the preformence using mutexes does seem proficient. 

Unless there concrete eveidence that the mutex part of this PR is causing performance issues, i think this implementation will be fine.

backend/internal/service/app_config_service.go
```diff
@@ -202,30 +213,52 @@ func (s *AppConfigService) UpdateAppConfig(input dto.AppConfigUpdateDto) ([]mode
...
 		key := field.Tag.Get("json")
 		value := rv.FieldByName(field.Name).String()
 
-		// If the emailEnabled is set to false, disable the emailOneTimeAccessEnabled
-		if key == s.DbConfig.EmailOneTimeAccessEnabled.Key {
+		// Apply special case logic for email settings
+		emailOneTimeAccessKey := ""
+		s.mu.RLock() // Short read lock just to get the key
+		emailOneTimeAccessKey = s.DbConfig.EmailOneTimeAccessEnabled.Key
+		s.mu.RUnlock()
```
wargio, line 235: i strongly suggest instead of doing this, create setter/getter which locks/unlocks within the function.

If you patch this race condition like this, it might create in the future more bugs caused by deadlocks.

PR #413, opened 2025-04-03
fix: ensure file descriptors are closed + other bugs

ItalyPaleAle:
1. Ensures file descriptors are closed: changes in #406 meant that when reading the default picture, file descriptors would be left open
2. fixes ownership issue of the `defaultPicture` stream (previous code was potentially unsafe as the returned slice was valid only until the next buffer modification)
3. Fixes generation of initials when there are non-ASCII characters.

Fixup from #406


stonith404 (approved):
Thanks!

PR #415, opened 2025-04-04
fix: ensure indexes on audit_logs table

ItalyPaleAle:
The table had complex queries that in all cases were performing full-table scans, and in some cases were parsing each row's data as JSON. This ensures indexes are present on all columns that are used for querying, including on the `clientName` JSON property in the data column

Also added passing context to the audit log methods which allows interrupting the long-running aggregate queries if the request is canceled (that is a small spin-off from #392 )


backend/resources/migrations/postgres/20250403082322_audit_logs_indexes.down.sql
kmendell, line 1: I think there something else at play here, none of the other migrations define the indexes, and they all have them , so the real question is why is the audit log model not creating it.

kmendell (commented):
Overall indexes shouldn't really matter in this case cause the entire audit log table is basically just being dumped not manipulated so indexes while nice to have and may be more performant in big datasets (if being modified) wouldn't really be necessary. 

However it is interesting why they are not be created automatically like the rest of them are.

kmendell:
I think we should figure out why they aren't being created automatically is my point with those comments.

ItalyPaleAle:
It seems that indexes are created only when using Gorm for migrations, which I don't believe Pocket ID is: https://gorm.io/docs/indexes.html#Index-Tag

They do matter for this table. For example, loading the global audit log runs these queries among others:

```sql
SELECT * FROM `audit_logs` ORDER BY `created_at` DESC LIMIT 20;
SELECT DISTINCT json_extract(data, '$.clientName') as client_name FROM `audit_logs` WHERE json_extract(data, '$.clientName') IS NOT NULL
```

Adding a filter adds more conditions such as:

```sql
SELECT * FROM `audit_logs` WHERE event = "CLIENT_AUTHORIZATION" AND json_extract(data, '$.clientName') = "MyApp" ORDER BY `created_at` DESC LIMIT 20
```

Both queries can significantly benefit from having indexes:

- Each WHERE clause or column in ORDER BY benefits heavily from indexes, otherwise the DB needs to perform a full-table scan on every query.
- In the case of conditions like `WHERE ... json_extract(data, '$.clientName') = "MyApp"`, without this PR there's both a full-table scan AND the JSON parsing of each value in `data` (JSON parsing is quite expensive)
- The query that loads all client names is particularly expensive since it performs a FTS and parses each row. DISTINCT can be very expensive to compute too.

The indexes added will solve all problems above. Especially, the index on `json_extract(data, '$.clientName')` (and equivalent for Postgres) makes the otherwise expensive operation a simple index lookup (you can verify this with `EXPLAIN QUERY PLAN SELECT DISTINCT json_extract(data, '$.clientName') as client_name FROM `audit_logs` WHERE json_extract(data, '$.clientName') IS NOT NULL` with and without the index

backend/resources/migrations/postgres/20250403082322_audit_logs_indexes.down.sql
ItalyPaleAle: Some do. For example:

https://github.com/pocket-id/pocket-id/blob/b65e693e12be2e7e4cb75a74d6fd43bacb3f6a94/backend/resources/migrations/sqlite/20250323184520_oidc_refresh_tokens.up.sql#L11

https://github.com/pocket-id/pocket-id/blob/b65e693e12be2e7e4cb75a74d6fd43bacb3f6a94/backend/resources/migrations/sqlite/20250302220732_api_key_auth.up.sql#L12

Also note that all columns that are primary keys, unique constraints, or have foreign keys have an automated index.

ItalyPaleAle:
Not sure. I added them because the Gorm tag said they were sortable. I can remove (and remove the Gorm tag) if that's not true

stonith404:
Yes those fields are sortable but I assume that this functionality isn't used that often. Because of that I'm not sure if it's worth to index them all. Almost every column in every table is sortable through the UI. I'm not sure how to performance impact is when tables get over-indexed. Would you recommend to index them all even though this functionality isn't used that often?

ItalyPaleAle:
I removed them.

There's not a huge downside in having extra indexes, but they do use memory in the DB, so if there's zero chance they are used, best to remove them. Let's just remember to add them later if needed

kmendell:
> It seems that indexes are created only when using Gorm for migrations, which I don't believe Pocket ID is: https://gorm.io/docs/indexes.html#Index-Tag

Yeah im just blind and didnt see them being created... My mistake on those comments. Then again i read that The database should auto create the indexes still , if theres a primary key, maybe im missreading though..

stonith404:
@kmendell I think when they mention automatically creating an index, they mean that PostgreSQL or SQLite creates an index on the primary key but not on all other columns. Every table should have a primary key, and if the DBMS were to create an index on every column in a table that has a primary key, almost every column would be indexed.

stonith404 (approved):

kmendell:
Got it , Not a databse expert so i just was relaying what i had researched :D

PR #423, opened 2025-04-09
refactor: simplify app_config service and fix race conditions

### LINKED ISSUE #391: 🐛 Bug Report: Race conditions in App Config

ItalyPaleAle:
Fixes #391
Fixes #407 

As discussed with @stonith404 , this PR refactors the app_config service with the goals of greatly simplify the logic, and fixing race conditions.

There are significant performance improvements too, especially when updating config, as the number of database round-trips has been reduced to no more than 2 (currently, it's 2 per each key).

1. Stores the config object in-memory in an `atomic.Pointer`, which removes the risk of data races.
2. Ensures consistency when updating config in the database, by using a transaction and acquiring a table-level lock. This ensures that config is only written by a single goroutine.
3. Simplifies the data structures a lot. In particular, default values and attributes (such as marking fields as non-public) can are not stored in the database anymore, but only in the code. This reduces the complexity of keeping all data in-sync, since it wasn't necessary to have that info int eh DB.

PR includes extensive unit tests which use an in-memory SQLite database to validate the behavior of the service.

Note: I had to update to Go 1.24 to be able to use `testing.Context()`.

### DIFF
--- a/.github/workflows/backend-linter.yml
+++ b/.github/workflows/backend-linter.yml
@@ -27,7 +27,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version-file: backend/go.mod
 
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -1,6 +1,6 @@
 module github.com/pocket-id/pocket-id/backend
 
-go 1.23.7
+go 1.24
 
 require (
 	github.com/caarlos0/env/v11 v11.3.1
--- a/backend/internal/bootstrap/bootstrap.go
+++ b/backend/internal/bootstrap/bootstrap.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	_ "github.com/golang-migrate/migrate/v4/source/file"
+
 	"github.com/pocket-id/pocket-id/backend/internal/service"
 )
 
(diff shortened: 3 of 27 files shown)


kmendell (commented):
Looks good to me so far, im currently on vacation so I can't do a full review just due to lack of time, but something I noticed there seems to be some spots where your calling GetDbConfig() more than once in the same function.

Maybe that can be set to the dbConfig variable like in other places? 

Sort of a nit pick but im thinking more on the lines of consistency. 

Otherwise looks good at my initial glance :)

stonith404 (approved):
Looks good to me, thank you @ItalyPaleAle!

@kmendell I couldn't find a function where `GetDbConfig()` is called twice. Maybe I'm blind, could you tell me where you saw that?

kmendell:
@stonith404  I must be blind, and was seeing stuff , now I can't find them :D , ignore me this should be good to go now :)

PR #425, opened 2025-04-10
fix: improve LDAP error handling

ItalyPaleAle:
This PR revises how errors are handled in the LDAP syncing service. I went through line-by-line to make sure that we treat errors consistently, and in most cases we should abort the sync and roll-back the changes in the DB (with some exceptions).

Should hopefully fix errors that were happening with syncing from LDAP, as some errors were simply logged while others causes failures.

### DIFF
--- a/backend/internal/service/ldap_service.go
+++ b/backend/internal/service/ldap_service.go
@@ -146,22 +146,22 @@ func (s *LdapService) SyncGroups(ctx context.Context, tx *gorm.DB, client *ldap.
 		groupMembers := value.GetAttributeValues(dbConfig.LdapAttributeGroupMember.Value)
 		membersUserId := make([]string, 0, len(groupMembers))
 		for _, member := range groupMembers {
-			cn := getCN(member)
-			if cn == "" {
+			ldapId := getDNProperty("uid", member)
+			if ldapId == "" {
 				continue
 			}
 
 			var databaseUser model.User
 			err = tx.
 				WithContext(ctx).
-				Where("username = ? AND ldap_id IS NOT NULL", cn).
+				Where("username = ? AND ldap_id IS NOT NULL", ldapId).
 				First(&databaseUser).
 				Error
 			if errors.Is(err, gorm.ErrRecordNotFound) {
 				// The user collides with a non-LDAP user, so we skip it
 				continue
 			} else if err != nil {
-				return fmt.Errorf("failed to query for existing user '%s': %w", cn, err)
+				return fmt.Errorf("failed to query for existing user '%s': %w", ldapId, err)
 			}
 
 			membersUserId = append(membersUserId, databaseUser.ID)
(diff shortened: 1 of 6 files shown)


kmendell:
Just to confirm this is just aligning the errors right for consistency? nothing was actually "failing" with the ldap sync?

backend/internal/service/ldap_service.go
```diff
@@ -128,29 +128,40 @@ func (s *LdapService) SyncGroups(ctx context.Context, tx *gorm.DB) error {
...
+		}
 
 		// Get group members and add to the correct Group
 		groupMembers := value.GetAttributeValues(dbConfig.LdapAttributeGroupMember.Value)
+		membersUserId := make([]string, 0, len(groupMembers))
 		for _, member := range groupMembers {
-			// Normal output of this would be CN=username,ou=people,dc=example,dc=com
-			// Splitting at the "=" and "," then just grabbing the username for that string
-			singleMember := strings.Split(strings.Split(member, "=")[1], ",")[0]
+			ldapId := getDNProperty("uid", member)
```
kmendell, line 149: shouldn't this be using the unique attribute value instead of hard coding "uid"?

backend/internal/service/ldap_service.go
```diff
@@ -268,17 +270,26 @@ func (s *LdapService) SyncUsers(ctx context.Context, tx *gorm.DB) error {
...
+		if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			// This could error with ErrRecordNotFound and we want to ignore that here
+			return fmt.Errorf("failed to query for LDAP user ID '%s': %w", ldapId, err)
+		}
 
 		// Check if user is admin by checking if they are in the admin group
 		isAdmin := false
 		for _, group := range value.GetAttributeValues("memberOf") {
-			if strings.Contains(group, dbConfig.LdapAttributeAdminGroup.Value) {
+			if getDNProperty("cn", group) == dbConfig.LdapAttributeAdminGroup.Value {
```
kmendell, line 290: Same thing here I think we should pull the unique attribute from the config vs hardcoding "cn"?

backend/internal/service/ldap_service.go
```diff
@@ -367,9 +390,28 @@ func (s *LdapService) saveProfilePicture(parentCtx context.Context, userId strin
...
+	property = strings.ToLower(property)
+	l := len(property) + 1
+	for _, v := range strings.Split(str, ",") {
+		v = strings.TrimSpace(v)
+		if len(v) > l && strings.ToLower(v)[0:l] == property+"=" {
+			return v[l:]
+		}
+	}
+
+	// CN not found, return an empty string
```
kmendell, line 415: if a CN isn't found then there's some majorly wrong with the ldap user, shouldn't we return a error instead? instead of making it a blank string?

kmendell (changes requested):
From my testing everything looks good, (besides that bug with the profile picture). See my notes above as well.

stonith404 (approved):
Looks good to me. Thank you!

kmendell (approved):

Issue #424, opened 2025-04-09
🚀 Feature: Login by QRCode ?


paulgreg:
### Feature description

The purpose is to make it easier to log in on another device by adding a QR Code.
The idea is, when generating a temporary code, to add that QR Code in the popup showing the URL with the one time code.
You can then simply scan it with your other device to be logged in.

### Pitch

I’ve installed pocket-id for my home and installed passkey on each computer for all family member and then repeat the operation on their phone.
Instead of typing the code on a new device, it would be much easier to scan an URL to then create a passkey.

kmendell:
Im adding the needs more upvotes label to this for now, @stonith404 I see some value to this, but what are your thoughts?

mzgaljic:
I don't have a strong opinion on this feature request, but I am curious, isn't this already provided out of the box by most passkey providers? 

Examples:
- [Google Password manager](https://developers.google.com/identity/passkeys/supported-environments#google-password-manager) - scroll down to _Figure 3_.
- [Passkey's with iCloud Keychain](https://support.apple.com/guide/iphone/use-passkeys-to-sign-in-to-websites-and-apps-iphf538ea8d0/ios#:~:text=On%20the%20other%20device%2C%20go,to%20scan%20the%20QR%20code.)


Are you perhaps not syncing passwords via a cloud provider? That could definitely be one situation where QR code login on another device wouldn't work out of the box.

paulgreg:
I’m using Bitwarden which doesn’t have that behavior.

My idea here is to make it easier to login via code.

So, for example, when you create a new user (which doesn’t have any passkey configured yet), you could simply display the QR code which she/he can scan with its phone to then add a passkey.

mzgaljic:
> I’m using Bitwarden which doesn’t have that behavior.

Today I learned. This feature sounds like a good idea!

paulgreg:
I can try to open a pull request if that’s ok with you (I have no experience with svelte but I can try 😅)

mzgaljic:
oh I'm not a maintainer, just a curious pocket-id user.   :)

kmendell:
> I can try to open a pull request if that’s ok with you (I have no experience with svelte but I can try 😅)

Sure, We are open for PR's :)

paulgreg:
I’m having trouble setting up the project. Is there a space (like discord) do discuss about it ?

kmendell:
> I’m having trouble setting up the project. Is there a space (like discord) do discuss about it ?

[...]

PR #436, opened 2025-04-12
feat: add qrcode representation of one time link (#424)

paulgreg:
My proposal to display a QRCode for one time links as discussed in #424 

It’s my first PR on a Svelte project so feel free to tell me what can be enhanced. Sorry If that’s look ugly to a Svelte expert :sweat_smile: 

Fixes: https://github.com/pocket-id/pocket-id/issues/424


kmendell:
i will take a look at this in a bit, first glance looks good but i have not tested yet

kmendell (approved):
@stonith404 I made a few changes to styling so make it more sleek. Otherwise it looks good to me, If all looks good to you this should be good to merge :)

frontend/src/lib/components/qrcode/qrcode.svelte
```diff
@@ -0,0 +1,59 @@
...
+			});
+		}
+	});
+</script>
+
+<div
+	class="qrcode-container"
+	style="--bg-color: {backgroundColor === 'transparent' ? 'transparent' : backgroundColor};"
+>
+	<canvas id="qr-code-canvas" class="rounded-lg"></canvas>
```
paulgreg, line 45: I’ve just find out about `bind` (https://stackoverflow.com/questions/67334531/how-to-get-html-element-of-svelte-component-using-bindthis).

So it would be better to add `bind:this={canvasEl}/`` here and define a `let canvasEl` above. We could then remove `id="qr-code-canvas"` and avoid a `document.getElementById` above ;-)

kmendell: you can change it to that sure, it's been along time since i have messed with canvases i just wanted to get rid of the white light border on the qr code

kmendell:
the tests need to be fixed too i just did not have time to fix those, so if you want to take a stab at fixing those as well feel free

frontend/src/lib/components/qrcode/qrcode.svelte
paulgreg: I’ve replace `getElementById` by a `bind:this`.

kmendell:
you have to build it with the go build tags , see the workflow on a example on how to do that i don't have the exact command off hand

stonith404 (approved):
Looks good to me, thanks!

Issue #404, opened 2025-03-30
🚀 Feature: Ability to disable users instead of deletion (and keeping user data intact)


abno85:
### Feature description

I'd like to have the possibility to disable users and blocking them from logging in instead of outright deleting them which includes all their settings, passkeys and audit logs etc.

### Pitch

I think it would be a nice feature in general to (temporary) disable users, but keeping their data and settings intact.

In particular when using LDAP as a provider, where a misconfigured LDAP filter easily leads to the loss of all settings and stored passkeys for the affected users. I don't know how Pocket-ID handles (temporary) disconnects from the LDAP provider, but these might also lead to loss of user data?

I just tried this with a test user. I moved the user into an LDAP group that is excluded from syncing to Pocket-ID via an LDAP filter. After the next sync the user was removed from Pocket-ID as intended, but when I added the user back to the original group, Pocket-ID created a completly new user from scratch and all passkeys and logins that where generated on the clients were now invalid.
I'm using [LLDAP](https://github.com/lldap/lldap) as the provider and a uuid for the unique user identifier, which didn't change obviously.

An alternative approach would be some kind of recycle bin on Pocket-IDs side? Allthough personally I'd prefer a general option to disable users.

kmendell:
I am looking into this , hopefully will have something soon :)

PR #437, opened 2025-04-13
feat: disable/enable users

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/404


kmendell:
@stonith404 Im leaving this as a draft for now, but could you take a look and see if you agree with this approach?

backend/internal/controller/user_controller.go
```diff
@@ -49,6 +49,9 @@ func NewUserController(group *gin.RouterGroup, authMiddleware *middleware.AuthMi
 
 	group.DELETE("/users/:id/profile-picture", authMiddleware.Add(), uc.resetUserProfilePictureHandler)
 	group.DELETE("/users/me/profile-picture", authMiddleware.WithAdminNotRequired().Add(), uc.resetCurrentUserProfilePictureHandler)
+
+	group.PUT("/users/:id/disable", authMiddleware.Add(), uc.disableUserHandler)
```
ItalyPaleAle, line 53: Should there be a single route such as `/users/:id/set` which allows passing options in the body, for example `{“enabled”:false}`? In the future there may be other options passed to users

backend/internal/service/ldap_service.go
```diff
@@ -279,6 +279,22 @@ func (s *LdapService) SyncUsers(ctx context.Context, tx *gorm.DB, client *ldap.C
 			Where("ldap_id = ?", ldapId).
 			First(&databaseUser).
 			Error
+
+		// If a user is found (even if disabled), enable them since they're now back in LDAP
+		if databaseUser.ID != "" && databaseUser.Disabled {
```
ItalyPaleAle, line 284: This would make it impossible to disable users imported form LDAP… Shouldn’t we allow controlling disabled/enabled independently from LDAP?

kmendell:
This isn't final. Im waiting for Elias. to make sure he likes this direction.

backend/internal/service/ldap_service.go
kmendell: Everything for LDAP users is controlled by the LDAP server, pocket id shouldn't be able to modify them

backend/internal/controller/user_controller.go
```diff
@@ -105,14 +107,19 @@ func (uc *UserController) listUsersHandler(c *gin.Context) {
 		return
 	}
 
-	var usersDto []dto.UserDto
-	if err := dto.MapStructList(users, &usersDto); err != nil {
-		_ = c.Error(err)
-		return
+	// Map the users to DTOs
```
stonith404, line 110: Is there any reason for this change?

backend/internal/model/user.go
```diff
@@ -19,6 +19,7 @@ type User struct {
 	IsAdmin   bool   `sortable:"true"`
 	Locale    *string
 	LdapID    *string
+	Disabled  bool `sortable:"true" gorm:"default:false"`
```
stonith404, line 22:
```suggestion
	Disabled  bool `sortable:"true"`
```

This has no effect as we are not using Gorm for migrations

frontend/src/lib/services/user-service.ts
```diff
@@ -95,4 +95,11 @@ export default class UserService extends APIService {
 		const res = await this.api.put(`/users/${id}/user-groups`, { userGroupIds });
 		return res.data as User;
 	}
+
+	async setUserAttribute(userId: string, attribute: string, value: any) {
+		const payload: Record<string, any> = {};
+		payload[attribute] = value;
+		const res = await this.api.put(`/users/${userId}/set`, payload);
+		return res.data as User;
+	}
```
stonith404, line 104:
```suggestion

async setUserAttribute(userId: string, attribute: string, value: any): Promise<User> {
  const response = await this.api.put(`/users/${userId}/set`, { [attribute]: value });
  return response.data as User;
}
```

I think this solution is a bit cleaner.

backend/internal/service/user_service.go
```diff
@@ -606,3 +637,64 @@ func (s *UserService) ResetProfilePicture(userID string) error {
...
 }
+
+// DisableUser disables a user by setting the "disabled" field to true
+func (s *UserService) DisableUser(ctx context.Context, userID string) error {
+	user, err := s.GetUser(ctx, userID)
+	if err != nil {
+		return err
+	}
+
+	// Log the user being disabled
```
stonith404, line 648: I would remove these kind of comments. They don't really help to understand the code better.

backend/internal/service/user_service.go
```diff
@@ -606,3 +637,64 @@ func (s *UserService) ResetProfilePicture(userID string) error {
...
+	log.Printf("Disabling user: %s (%s)", user.Username, userID)
+
+	// Add retry logic for database locks (especially helpful for SQLite)
+	maxRetries := 5
```
stonith404, line 652: Is this retry logic really necessary? As far as I know Sqlite just waits until the database is unlocked.

backend/internal/middleware/auth_middleware.go
stonith404, line 1: Inside the auth middleware we have to check if the user is disabled because a disabled user shouldn't be able to do requests.

backend/internal/service/user_service.go
```diff
@@ -170,9 +175,35 @@ func (s *UserService) UpdateProfilePicture(userID string, file io.Reader) error
 }
 
 func (s *UserService) DeleteUser(ctx context.Context, userID string, allowLdapDelete bool) error {
-	return s.db.Transaction(func(tx *gorm.DB) error {
-		return s.deleteUserInternal(ctx, userID, allowLdapDelete, tx)
-	})
+	tx := s.db.Begin()
+	defer func() {
+		if r := recover(); r != nil {
```
stonith404, line 180: Why is a call of `recover` necessary?

stonith404 (changes requested):

backend/internal/middleware/auth_middleware.go
kmendell: I had this initially but i moved it to the Webauthn service as the auth middleware still logged in the user, unless maybe i did it wrong?

backend/internal/service/user_service.go
kmendell: Ill do a run over it , i put those in for my debugging, and sanity, i always forget to remove them...

backend/internal/service/user_service.go
kmendell: No i dont think so , i was having issues with the DB being locked, but i think i fixed those anotehr way, and just forgot these changes, i will change that back.

stonith404: Oh, okay. Let me know if you encounter this issue again. If the database is locked, there is probably a transaction that hasn't been committed or rolled back correctly.

backend/internal/service/user_service.go
stonith404: No worries 😅

backend/internal/service/user_service.go
```diff
@@ -170,9 +175,35 @@
...
+		return err
+	}
+
+	// Only soft-delete if user is LDAP and soft-delete is enabled and not allowing hard delete
+	if user.LdapID != nil && s.appConfigService.GetDbConfig().LdapSoftDeleteUsers.IsTrue() && !allowLdapDelete {
+		if !user.Disabled {
+			// Block hard delete if LDAP user is not disabled
+			tx.Rollback()
+			return fmt.Errorf("LDAP user must be disabled before deletion")
+		}
```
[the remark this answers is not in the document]
kmendell: @stonith404 Should we worry about this alert? I know there were others that were fine.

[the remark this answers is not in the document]
stonith404: It's okay because a malicious user ID would cause the function to fail early since the user ID can't be found in the database.

stonith404 (approved):
Looks good to me, thanks! Please take a quick look at my changes and let me know if you don't agree with any of it.

kmendell:
Looks good to me as well!

Issue #428, opened 2025-04-10
🐛 Bug Report: Profile picture is empty if user has no first or last name


ItalyPaleAle:
### Reproduction steps

1. Create a user without a first or last name (could be imported from LDAP)
2. The profile picture is an empty white square (and saved on a file named `data/.png`)

### Expected behavior

It should have have a default profile picture with a placeholder. For example, a dash, or a silhouette, or something else.

### Actual Behavior

There's a white square as profile picture, saved to file `data/.png`

### Version and Environment

main branch

PR #449, opened 2025-04-16
fix: profile picture empty for users without first or last name

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/428

If no first or last name is found it uses the first two letters of the username to create the profile picture.


stonith404 (approved):
Looks good to me, thanks.

PR #451, opened 2025-04-17
feat: send email to user when api key expires within 7 days

kmendell:


backend/resources/email-templates/api-key-expiring-soon_text.tmpl
```diff
@@ -0,0 +1,12 @@
...
+API Key Expiring Soon
+====================
+
+Hello {{ .Data.Name }},
+
+This is a reminder that your API key "{{ .Data.ApiKeyName }}" will expire on {{ .Data.ExpiresAt.Format "2006-01-02 15:04:05 MST" }}.
+
+Please generate a new API key if you need continued access.
+
+If you have any questions, contact your administrator.
```
stonith404, line 11:
```suggestion
```
I think this sentence is unnecessary, or do you really want to keep it?

backend/resources/email-templates/api-key-expiring-soon_html.tmpl
```diff
@@ -0,0 +1,18 @@
...
+        </div>
+        <div class="warning">Warning</div>
+    </div>
+    <div class="content">
+        <h2>API Key Expiring Soon</h2>
+        <p>
+            Hello {{ .Data.Name }},<br/><br/>
+            This is a reminder that your API key <strong>{{ .Data.ApiKeyName }}</strong> will expire on <strong>{{ .Data.ExpiresAt.Format "2006-01-02 15:04:05 MST" }}</strong>.<br/><br/>
+            Please generate a new API key if you need continued access.<br/><br/>
+            If you have any questions, contact your administrator.
```
stonith404, line 15:
```suggestion
            Please generate a new API key if you need continued access.
```
I think this sentence is unnecessary, or do you really want to keep it?

backend/internal/dto/api_key_dto.go
```diff
@@ -11,12 +11,13 @@ type ApiKeyCreateDto struct {
...
-	ExpiresAt   datatype.DateTime  `json:"expiresAt"`
-	LastUsedAt  *datatype.DateTime `json:"lastUsedAt"`
-	CreatedAt   datatype.DateTime  `json:"createdAt"`
+	ID                  string             `json:"id"`
+	Name                string             `json:"name"`
+	Description         string             `json:"description"`
+	ExpiresAt           datatype.DateTime  `json:"expiresAt"`
+	LastUsedAt          *datatype.DateTime `json:"lastUsedAt"`
+	CreatedAt           datatype.DateTime  `json:"createdAt"`
+	ExpirationEmailSent bool               `json:"expirationEmailSent"`
```
stonith404, line 20:
```suggestion
```

Is this really necessary inside the DTO, because the value ins't used in the frontend?

backend/internal/job/api_key_expiry_job.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Fatalf("Failed to create a new scheduler: %v", err)
+	}
+
+	registerJob(ctx, scheduler, "ExpiredApiKeyEmailJob", "0 0 * * *", jobs.checkAndNotifyExpiringApiKeys)
+
+	// Start the scheduler before running the first check
+	scheduler.Start()
+
+	// Run the first check asynchronously, so it doesn't block startup
+	go func() {
```
stonith404, line 33: I think it's sufficient for the job to run only at midnight. Or do you have any reason why it should also run at startup?

backend/internal/service/api_key_service.go
```diff
@@ -117,3 +119,53 @@ func (s *ApiKeyService) ValidateApiKey(ctx context.Context, apiKey string) (mode
...
+	cutoff := now.AddDate(0, 0, daysAhead)
+
+	// Convert to Unix timestamps
+	nowUnix := now.Unix()
+	cutoffUnix := cutoff.Unix()
+
+	err := s.db.
+		WithContext(ctx).
+		Preload("User").
+		Where("expires_at > ? AND expires_at <= ? AND expiration_email_sent = ?", nowUnix, cutoffUnix, false).
```
stonith404, line 135: We should use `datatype.DateTime` instead of unix timestamps for queries or storing date and time data. 

This is an example:
https://github.com/pocket-id/pocket-id/blob/e21ee8a871134863e0834c32a58df3578dbd8289/backend/internal/job/db_cleanup_job.go#L36

The reason for this datatype is to prevent issues with the date and time of Sqlite. In Sqlite this datatype returns an unix timestamp and for Postgres it returns a `time.Time`.

backend/internal/dto/api_key_dto.go
kmendell: I think i was more over just making sure that it was there for the API responses. Maybe that not needed?

backend/internal/job/api_key_expiry_job.go
kmendell: I was doing that just to mimic how the ldap job did it, and more for testing. But i think the midnight job should be enough yes. Ill change it :)

backend/internal/service/api_key_service.go
kmendell: Huh not sure how i missed that , i will get that fixed.

stonith404: No worries, I know it's a bit confusing.

backend/internal/dto/api_key_dto.go
stonith404: You're right 👍

backend/internal/service/api_key_service.go
kmendell: @stonith404 This should be fixed now i believe

stonith404 (approved):
Looks good to me :)

PR #452, opened 2025-04-17
feat: new login code card position

James18232:
feature to move the Login Code card higher on the page for improved UX.


kmendell (changes requested):
I would say this shouldn't go any higher than the account details card, it shouldn't be the first thing on the page.

I won't have time to dive deep into this right now, but i'll look tomorrow if elias hasn't yet.

James18232:
noted re location, but suggest the most used features should be the most accessible - eg within finger clicking distance on smaller screens (perhaps applied with responsive design triggers). two taps max, one to open the OIDC app, one to request a code and done is a good UX flow imho. 

Making the create button the same width as the card on smaller screens would improve this further and also be comfortable for both left/right handers on mobile devices. thanks.

James18232:
i have tweaked the PR as suggested - only change location and button width for smaller screens, otherwise no change for desktop users. trigger point is tailwinds sm screen, which seems appropriate.

James18232:
changes as suggested - only changes for smaller screens.

kmendell (approved):
Looks good to me! @stonith404 If all looks good to you LGTM

PR #457, opened 2025-04-18
feat: add ability to send login code via email

### LINKED ISSUE #435: 🚀 Feature: Send One-Time Link via Email
### Feature description

1. Navigate to the Users page in the Pocket ID admin dashboard.
2. Locate the user you want to set up a passkey for.
3. Click the three dots on the right side of the user row.
4. Click One-Time Link.
5. Select an Expiration Time for the link.
6. Click Generate Link
7.  **Have the option to email that link to the user you want to set up a passkey for.** 

### Pitch

In my opinion, implementing this feature would significantly reduce the administrative burden associated with user setup, even though it is currently quite minimal.

Furthermore, I do not believe it poses the same risks as the One-Time Access Email, as the administrator could simply email those links themselves.

stonith404:
Closes #435


kmendell:
@stonith404 I took a look at this as i had seom free time, 1. I fixed the tests for you :). 2. Just an idea but maybe we should hide the send email button if the One Time Access Email is not enabled?

PR #465, opened 2025-04-20
ci/cd: setup caching and improve ci job performance

kmendell:


.github/workflows/e2e-tests.yml
```diff
@@ -153,23 +213,25 @@ jobs:
...
+          pocket-id:test
 
           docker logs -f pocket-id-postgres &> /tmp/backend.log &
 
       - name: Run Playwright tests
         working-directory: ./frontend
         run: npx playwright test
 
-      - uses: actions/upload-artifact@v4
-        if: always() && github.event.pull_request.head.ref != 'i18n_crowdin'
```
stonith404, line 165: I would keep the Crowdin check because else this will also run for the translations PR.

.github/workflows/e2e-tests.yml
```diff
@@ -19,21 +19,41 @@ jobs:
...
-          tags: pocket-id/pocket-id:test
+          push: false
+          load: false
+          tags: pocket-id:test
           outputs: type=docker,dest=/tmp/docker-image.tar
           build-args: BUILD_TAGS=e2etest
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+
+      - name: Move cache
```
stonith404, line 46: Is this step really necessary? Couldn't we just store it directly to`/tmp/.buildx-cache`?

.github/workflows/e2e-tests.yml
kmendell: Whoops i just forgot to add it back in after i reverted back , ill get that back added in :)

kmendell: Should be added back now :)

stonith404 (approved):

PR #466, opened 2025-04-20
refactor: add ldap tests

kmendell:


kmendell:
@stonith404 Feel free to tweak / add more tests, but this should be the basic logic for it :)

PR #471, opened 2025-04-22
fix: prevent deadlock when trying to delete LDAP users

ItalyPaleAle:
Fixup from #437
Reverted LDAP to use `deleteUserInternal` which accepts an existing transaction. Without that, it would cause a deadlock in SQLite, as we can't start a transaction if there's another one active already

Also renamed `DisableUser` to `disableUserInternal` since it's another "internal" method


kmendell (approved):

stonith404:
Thanks!

PR #481, opened 2025-04-24
refactor: do not force redirects to happen on the server

ItalyPaleAle:
Small refactor on the frontend: do not force redirects to happen on the server with `+page.server.ts`; use `+page.ts` instead, as they can happen on the client too

### DIFF
--- a/frontend/src/hooks.server.ts
+++ b/frontend/src/hooks.server.ts
@@ -27,13 +27,11 @@ const authenticationHandle: Handle = async ({ event, resolve }) => {
 	const isPublicPath = ['/authorize', '/health'].includes(event.url.pathname);
 	const isAdminPath = event.url.pathname.startsWith('/settings/admin');
 
-	if (!isUnauthenticatedOnlyPath && !isPublicPath) {
-		if (!isSignedIn) {
-			return new Response(null, {
-				status: 302,
-				headers: { location: '/login' }
-			});
-		}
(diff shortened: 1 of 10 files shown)


kmendell:
Just curious What does this solve? What benefit does this provide? Maybe im missing something but i dont see the point.

frontend/src/hooks.server.ts
```diff
@@ -81,7 +79,7 @@ function verifyJwt(accessToken: string | undefined) {
 		const jwtPayload = decodeJwt<{ isAdmin: boolean }>(accessToken);
 		if (jwtPayload?.exp && jwtPayload.exp * 1000 > Date.now()) {
 			isSignedIn = true;
-			isAdmin = jwtPayload?.isAdmin || false;
```
user00265, line 84: What is the purpose of changing readable code with shorthand on this instance?

ItalyPaleAle: `jwtPayload?.isAdmin || false` may not always be boolean. For example if the initial value is `2`, it remains `2`, which is a truthy value but not boolean. `!!(jwtPayload?.isAdmin)` is guaranteed to be boolean.

stonith404: It's impossible that `isAdmin` isn't a bool, unless the user would manipulate the JWT manually which would break the page anyway. 

In terms of readability I think this is still readable so we can keep it.

stonith404 (approved):
Thanks!

PR #482, opened 2025-04-25
refactor: graceful shutdown for server

ItalyPaleAle:
Initial work on graceful shutdown, for now focused on the Gin server.

Implements the [graceful shutdown pattern](https://www.freecodecamp.org/news/graceful-shutdowns-k8s-go/), by capturing SIGINT or SIGTERM signals, and asking the Gin server to drain all active requests before shutting down. This makes it so requests aren't interrupted while in-flight, if possible.

Additionally, includes _rudimental_ graceful shutdown for the job scheduler, although this is not complete yet (the app doesn't wait for the scheduler to be shut down before exiting).

As part of the work on the scheduler, there's now a single gocron scheduler created, instead of 4. This is useful because each gocron scheduler spawns its own set of goroutines in background.


backend/internal/bootstrap/application_images_bootstrap.go
```diff
@@ -55,6 +54,11 @@ func imageAlreadyExists(fileName string, destinationFiles []os.DirEntry) bool {
 }
 
 func getImageNameWithoutExtension(fileName string) string {
-	splitted := strings.Split(fileName, ".")
-	return strings.Join(splitted[:len(splitted)-1], ".")
+	idx := strings.LastIndexByte(fileName, '.')
```
kmendell, line 57: What's the purpose of this change?

backend/internal/utils/signals/signal.go
```diff
@@ -0,0 +1,40 @@
...
+https://github.com/kubernetes-sigs/controller-runtime/blob/8499b67e316a03b260c73f92d0380de8cd2e97a1/pkg/manager/signals/signal.go
+Copyright 2017 The Kubernetes Authors.
+License: Apache2 (https://github.com/kubernetes-sigs/controller-runtime/blob/8499b67e316a03b260c73f92d0380de8cd2e97a1/LICENSE)
+*/
+
+var onlyOneSignalHandler = make(chan struct{})
+
+// SignalContext returns a context that is canceled when the application receives an interrupt signal.
+// A second signal forces an immediate shutdown.
+func SignalContext(parentCtx context.Context) context.Context {
```
kmendell, line 22: Isn't this something docker already handles?

kmendell (commented):
I would like more context on the purpose of these changes, as while i get the idea here, i feel like there may be a better approach. I just want to know you goal with this, is GIN not exiting properly causing issues?  (though im not sure how often this happens, you would have to have constant requests for this to happen in my mind)

backend/internal/bootstrap/application_images_bootstrap.go
ItalyPaleAle: I saw this code which was not efficient, and optimized it. `strings.Split` iterates through the entire string and allocates a slice on the heap (which could be re-allocated if there are many dots). It then re-composes a string which causes more allocations.

A simpler approach is to look for the last dot in the string and take what's before that. This doesn't cause any allocation on the heap.

ItalyPaleAle:
The primary goal is to (start implementing) graceful shutdown. Normally when an app receives the termination signal, it exits right away, and if there are requests in-flight (or in the queue), they are aborted half-way. The idea of graceful termination is that once the process receives the termination signal, it exits gracefully: first it stops accepting new requests, and then it waits for requests currently in-flight to terminate. It's a common pattern and if you want more ... context on it (ahah!) I suggest reading the [article I linked](https://www.freecodecamp.org/news/graceful-shutdowns-k8s-go/).

To add to the above, and to answer your question as to why Docker isn't doing it already... It's a bit more complicated.

Leave Docker aside for a moment and imagine the process is running in a shell without a container orchestrator and without a container. You normally terminate the process by hitting CTRL+C: that sends the process the "interrupt signal", aka SIGINT, which means the kernel is politely asking the process to shut down. (A similar signal is SIGTERM which is sent when you use the `kill` command or on Windows when you terminate the process from Process Manager).

What's important to note is that SIGINT/SIGTERM are asking the process to shut itself down. In Go apps, if you don't "trap" the signal yourself, by default Go terminates the process right away with a non-0 signal.

Now, what this PR does is "trapping" SIGINT/SIGTERM. When the process receives the signal, it starts a "graceful" shutdown sequence. You can see it happening [on this line](https://github.com/pocket-id/pocket-id/pull/482/files#diff-6a53e9025bc0ab38a81173528e643798f7d5d9a086a29646818997443f21c0a3R30), which has a channel that receives a message when the process receives one of those signals. When that happens, it cancels the context, which then triggers the graceful shutdown procedure (with the help of the Go's HTTP server and other code, which I haven't fully implemented _yet_ - so only the server is properly shutting down gracefully in this PR).

> In addition to that, note that:
>
> 1. A second SIGINT/SIGTERM received by the app causes an immediate shutdown (so if you press CTRL+C twice, the app always shuts down right away). [See this code](https://github.com/pocket-id/pocket-id/pull/482/files#diff-6a53e9025bc0ab38a81173528e643798f7d5d9a086a29646818997443f21c0a3R34-R36)
> 2. The HTTP server itself is given a timeout to shut down of 5s [(see code)](https://github.com/pocket-id/pocket-id/pull/482/files#diff-4c4f608c327a055cd5223e61ff6c84f00d38c10b5daf62f7e040e0245f6288f6R160-R162). Otherwise, there's a chance it could block for too long.

The other signal you need to be aware of is SIGKILL. This is sent when you run the command `kill -9` (with the `-9` flag). This is a forceful shutdown: it tells the kernel to terminate the process, no if's or but's. That signal can NOT be trapped by a process.

Now, let's talk about containers. Here, Docker, Podman, Kubernetes etc have a very similar behavior.

1. When you terminate the container, they send the process with pid 0 in the container a SIGTERM.
2. The orchestrator gives the container a grace period to shut itself down. In the case of Docker, it is 10 seconds [(docs here)](https://docs.docker.com/reference/cli/docker/container/stop/)
3. If the process isn't exited after the graceful termination time, then the orchestrator sends a SIGKILL (forcefully terminates the process)

> One thing to point out is that the container orchestrator sends the signal to the process with PID 0 in the container. In the case of Pocket ID, that's actually the shell script that runs all the processes (Go, Node, Caddy)

backend/internal/utils/signals/signal.go
ItalyPaleAle: See my [comment below](https://github.com/pocket-id/pocket-id/pull/482#issuecomment-2829249000) for a deeper explanation

kmendell (approved):
This looks good to me. I was able to confirm the functionality. 

I would say maybe we finish the scheduler shutdown part before we merge?

Issue #380, opened 2025-03-23
🐛 Bug Report: Audit log always display my Nginx Proxy Manager's IP


LucasJanin:
### Reproduction steps

My Pocket-id is installed on an LXC on my Proxmox using Proxmox VE Helper Scripts

```
sudo bash -c "$(wget -qLO - https://github.com/community-scripts/ProxmoxVE/raw/main/ct/pocketid.sh)"
```
I added to my `/opt/pocket-id/backend/.env`
```
TRUST_PROXY=true
CADDY_DISABLED=true
```

Set up Nginx Proxy Manager to pocket-id with this Custom Nginx Configuration

```
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-Proto $scheme;
proxy_set_header X-Forwarded-Host $host;
```

However, the IPs shown in the Audit Log are always the Nginx Proxy Manager's IP.

<img width="1224" alt="Image" src="https://github.com/user-attachments/assets/0619b8a6-26c6-4b50-b067-7e8627784a4b" />

To ensure my Nginx Proxy Manager is set up correctly, I use this Docker:

```
sudo docker run -d --name httpbin -p 3000:80 kennethreitz/httpbin
http://localhost:3000/headers
```

Setup a new host in Nginx Proxy Manager with that same Custom Nginx Configuration

```
curl -H "X-Forwarded-For: 192.168.1.99" https://ip-echo.xxx.xx
```

```
{
  "path": "/",
  "headers": {
    "host": "ip-echo.xxx.xx",
    "x-forwarded-scheme": "https",
    "x-forwarded-proto": "https",
    "x-forwarded-for": "192.168.1.99, 192.168.1.30",
    "x-real-ip": "192.168.1.30",
    "user-agent": "curl/8.7.1",
    "accept": "*/*"
  },
  "method": "GET",
  "body": "",
  "fresh": false,
  "hostname": "ip-echo.xxx.xx",
  "ip": "192.168.1.99",
  "ips": [
    "192.168.1.99",
    "192.168.1.30"
  ],
  "protocol": "https",
  "query": {},
  "subdomains": [
    "ip-echo"
  ],
  "xhr": false,
  "os": {
    "hostname": "16727cfaf0b3"
  },
  "connection": {}
}
```

From my understanding, the Nginx Proxy Manager behaves as expected.

### Expected behavior

[...]

PR #483, opened 2025-04-25
fix: ensure to use correct forwarded ip

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/380

GIN doesnt seem to repect the X-Client-IP or Forwarded IP if the built in caddy is not used. This logic does not work yet , but still looking into it.

### DIFF
--- a/backend/internal/bootstrap/router_bootstrap.go
+++ b/backend/internal/bootstrap/router_bootstrap.go
@@ -54,6 +54,7 @@ func initRouter(ctx context.Context, db *gorm.DB, appConfigService *service.AppC
 	rateLimitMiddleware := middleware.NewRateLimitMiddleware()
 
 	// Setup global middleware
+	r.Use(middleware.NewTrustProxyMiddleware().Add())
 	r.Use(middleware.NewCorsMiddleware().Add())
 	r.Use(middleware.NewErrorHandlerMiddleware().Add())
 	r.Use(rateLimitMiddleware.Add(rate.Every(time.Second), 60))
--- a/backend/internal/middleware/trust_proxy.go
+++ b/backend/internal/middleware/trust_proxy.go
@@ -1,0 +1,44 @@
+package middleware
+
+import (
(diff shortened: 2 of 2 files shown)


savely-krasovsky:
It's strange because in my case Pocket-ID is definitely behind Traefik proxy and still it works properly.

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -54,6 +54,7 @@ func initRouter(ctx context.Context, db *gorm.DB, appConfigService *service.AppC
 	rateLimitMiddleware := middleware.NewRateLimitMiddleware()
 
 	// Setup global middleware
+	r.Use(middleware.NewTrustProxyMiddleware().Add())
```
ItalyPaleAle, line 57: Why not just relying on https://pkg.go.dev/github.com/gin-gonic/gin#Engine.SetTrustedProxies ?

kmendell: Thats already set by default and we still see the issue. See the issue referenced.

ItalyPaleAle: Right, but this PR is essentially equivalent to [what Gin is already doing](https://github.com/gin-gonic/gin/blob/75ccf94d605a05fe24817fc2f166f6f2959d5cea/context.go#L802C1-L837C2) (if you look at the blame for those lines, you'll see that I partially contributed that code :) )

I do experience the issue in #380 too, but it is a problem with Podman in my case too. Traefik is correctly injecting X-Forwarded-For, and backend services are receiving it, but the injected IP is wrong. I have validated that replacing Pocket ID with another service, like [whoami](https://github.com/traefik/whoami), shows the header is present but incorrect

kmendell: Yes it is that why its in draft, Im more over exploring options to see if i can fix this.

PR #490, opened 2025-04-27
Fix incorrectly swapped refreshToken and accessToken

### LINKED ISSUE #489: 🐛 Bug Report: Logins fail after upgrading pocket id
### Reproduction steps

Reproduction steps: integrate latest main branch pocket-id and latest immich. Observe logins fail.

Disclaimer: I updated pocket-id _and_ the app I was using that uses pocket-id. However I'm pretty sure that the bug is in pocket-id.

I use an app called [Immich](https://immich.app). This integrates with pocket-id for login. I recently found I can't log in. What I'm observing is a failed login, and in the pocket-id logs, I observe:

```
Apr 27 18:46:30 homelab pocket-id-backend-start[3369060]: Error #01: failed to parse token: jwt.ParseString: failed to parse string: unknown payload type (payload is not JWT?)
```

When I MITM the connection between Immich and pocket-id, what I observe is that immich successfully hits `/api/oidc/token` and is given an `access_token`, an `id_token` and a `refresh_token`. It then tries to hit the `oidc/userinfo` endpoint with the `access_token`. This then fails because it's not a JWT, and the login fails.

### Expected behavior

Login succeeds. The userinfo endpoint should work with the accesstoken that was generated, most likely?

### Actual Behavior

HTTP 500.

### Version and Environment

4c33793678709eb4981be2c1fd5803bace5f5939

### Log Output

_No response_

j-baker:
Fixes #489.

Oops. I'm not entirely sure how this would be tested because it's so high up in the stack - needs HTTP afaict. Either way, it wasn't tested so far!

Good news is that this was only broken in the last day or two I guess and probably hasn't gotten to a release!

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -155,7 +155,7 @@ func (oc *OidcController) createTokensHandler(c *gin.Context) {
 		input.ClientID, input.ClientSecret, _ = c.Request.BasicAuth()
 	}
 
-	idToken, refreshToken, accessToken, expiresIn, err :=
+	idToken, accessToken, refreshToken, expiresIn, err :=
 		oc.oidcService.CreateTokens(c.Request.Context(), input)
 
 	switch {


kmendell (approved):
It was accidentally changed in the device code PR, https://github.com/pocket-id/pocket-id/commit/22f7d64bf08a5a1ecbe5eee0052453b730f5c360#diff-21ae94b87213fba831ae50c5ff9a5c85c28396c549269cfd1e33a0561c158298R158

kmendell:
Props to you on finding this :), And thank you!

PR #493, opened 2025-04-29
refactor: complete graceful shutdown implementation and add service runner

ItalyPaleAle:
Completes the work started in #482 to support graceful shutdown.

1. Introduces a ServiceRunner which manages long-running background services, running them until the context is canceled (and collecting errors)
2. Refactored `initBootstrap` and the GeoLite service's Run method to implement the ServiceRunner's Service interface
3. Moved the initialization of all "services" (not long-running) and the scheduler into their own function

### DIFF
--- a/backend/internal/bootstrap/bootstrap.go
+++ b/backend/internal/bootstrap/bootstrap.go
@@ -3,6 +3,8 @@ package bootstrap
 import (
 	"context"
 	"fmt"
+	"log"
+	"time"
 
 	_ "github.com/golang-migrate/migrate/v4/source/file"
 
@@ -47,6 +49,18 @@ func Bootstrap() error {
 		Run(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to run services: %w", err)
+	}
+
+	// Invoke all shutdown functions
+	// We give these a timeout of 5s
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer shutdownCancel()
+	err = utils.
+		// TODO: Add shutdown services here
+		NewServiceRunner().
+		Run(shutdownCtx)
+	if err != nil {
+		log.Printf("Error shutting down services: %v", err)
 	}
 
 	return nil
--- a/backend/internal/bootstrap/bootstrap.go
+++ b/backend/internal/bootstrap/bootstrap.go
@@ -2,25 +2,52 @@ package bootstrap
 
 import (
 	"context"
+	"fmt"
 
 	_ "github.com/golang-migrate/migrate/v4/source/file"
 
-	"github.com/pocket-id/pocket-id/backend/internal/service"
+	"github.com/pocket-id/pocket-id/backend/internal/utils"
 	"github.com/pocket-id/pocket-id/backend/internal/utils/signals"
 )
 
-func Bootstrap() {
+func Bootstrap() error {
 	// Get a context that is canceled when the application is stopping
 	ctx := signals.SignalContext(context.Background())
 
 	initApplicationImages()
 
+	// Perform migrations for changes
 	migrateConfigDBConnstring()
-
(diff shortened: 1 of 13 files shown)


kmendell (approved):
Look good to me! I tested it and all functionality remains the same.

stonith404:
Looks good to me at first glance. I'll merge this tomorrow.

backend/internal/service/geolite_service.go
```diff
@@ -42,25 +42,56 @@ var tailscaleIPNets = []*net.IPNet{
...
+
+func (s *GeoLiteService) Run(ctx context.Context) error {
+	if s.disableUpdater {
+		// If the updater is disabled, this runner is a no-op, which is blocked until the context is canceled
+		<-ctx.Done()
+		return ctx.Err()
+	}
+
+	// Update the GeoIP database immediately (after a 1s delay) and every 2 days
+	ticker := time.NewTicker(48 * time.Hour)
```
stonith404, line 69: Wouldn't it be cleaner to schedule the update with the existing scheduler like we do with the other jobs?

ItalyPaleAle: You make a valid point. I did it this way as it's what it is currently, but it should use the scheduler.

stonith404: Besides that everything looks good to me. I think that's a nice addition, thanks :)

ItalyPaleAle: I'll have it done tonight or tomorrow at the latest

ItalyPaleAle: Done

backend/internal/bootstrap/bootstrap.go
```diff
@@ -2,25 +2,66 @@ package bootstrap
...
+		return fmt.Errorf("failed to run services: %w", err)
+	}
+
+	// Invoke all shutdown functions
+	// We give these a timeout of 5s
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer shutdownCancel()
+	err = utils.
+		// TODO: Add shutdown services here
+		NewServiceRunner().
```
ItalyPaleAle, line 60: @daenney you can add the shutdown functions for OTel here

daenney: Perfect. I'll wait for this to land first and update then. Thank you for the pointer!

backend/internal/service/geolite_service.go
```diff
@@ -42,25 +42,22 @@ var tailscaleIPNets = []*net.IPNet{
...
-		err := service.updateDatabase(ctx)
-		if err != nil {
-			log.Printf("Failed to update GeoLite2 City database: %v", err)
-		}
-	}()
-
 	return service
 }
 
+func (s *GeoLiteService) HasBackgroundService() bool {
```
stonith404, line 57: Wouldn't it make more sense to make `disableUpdater` public instead of having this getter? If you want to keep this function I would recommend to rename it because I find `HasBackgroundService` a bit misleading.

stonith404:
Golangci warns that the error of `registerScheduledJobs` isn't handled. After this is fixed it should be ready to merge :)

backend/internal/service/geolite_service.go
ItalyPaleAle: I would prefer not exposing a property, just so no one can set it. But I will rename the function

ItalyPaleAle:
> Golangci warns that the error of registerScheduledJobs isn't handled. After this is fixed it should be ready to merge :)

Whoops, that was my miss, fixed

backend/internal/service/geolite_service.go
stonith404: You're right, that makes sense.

stonith404 (approved):
Thank you for these changes!

PR #494, opened 2025-04-29
feat: add healthz endpoint

ItalyPaleAle:
Adds the /healthz endpoint, which responds with a 204 status code (No Content) to all requests.

This can be used to check if Pocket ID is alive and working. It can be used with container orchestrators too (see [Docker's HEALTHCHECK](https://docs.docker.com/reference/dockerfile/#healthcheck))

### DIFF
--- a/frontend/src/hooks.server.ts
+++ b/frontend/src/hooks.server.ts
@@ -24,7 +24,7 @@ const authenticationHandle: Handle = async ({ event, resolve }) => {
 	const { isSignedIn, isAdmin } = verifyJwt(event.cookies.get(ACCESS_TOKEN_COOKIE_NAME));
 
 	const path = event.url.pathname;
-	const isUnauthenticatedOnlyPath = path == '/login' || path.startsWith('/login/') || path == '/lc' || path.startsWith('/lc')
+	const isUnauthenticatedOnlyPath = path == '/login' || path.startsWith('/login/') || path == '/lc' || path.startsWith('/lc/')
 	const isPublicPath = ['/authorize', '/device', '/health', '/healthz'].includes(path);
 	const isAdminPath = path == '/settings/admin' || path.startsWith('/settings/admin/');
 


backend/internal/controller/healthz_controller.go
kmendell, line 1: @ItalyPaleAle This for a backend health check? We already have a /health (https://github.com/pocket-id/pocket-id/tree/main/frontend/src/routes/health) endpoint, maybe it may be worth trying to combind these two?

ItalyPaleAle: Oh yeah good point. I should change that FE heallth-check to use /healthz on the backend instead of requesting the app config, which uses less resources.

Also, it should probably be /healthz (as an alias, to avoid breaking existing apps), as that's the standard for cloud-native apps

stonith404: I decided to list the app configuration because it also checks if the database connection is working. However, since the app configuration is now stored in memory, this doesn't make much sense anymore. But maybe there is another more performant way to also check that the database connection is working?

ItalyPaleAle: If we want to check for the database's availability too, a simple way could be to run a query like `SELECT 1`. We don't need to retrieve any real data, it just needs to prove the DB is responding. I would recommend NOT running this query on every invocation of `/healthz` however, because healthchecks can be very frequent (for example, on K8s healthchecks can be performed by more than one master node at the same time, so you get 3 requests every 30s instead of 1), running the query on every time can greatly increase the load on the DB (this is especially bad on SQLite since each query blocks the entire DB). If we do that, we should do the DB pings every 30s and cache the response.

**However**, I do not really recommend querying the health of the DB during the scheduled healthchecks. For two reasons:

1. The simpler reason is that it increases the load of the DB. If every app (not just Pocket ID) did a healthcheck on the DB every 30s, it would cause a lot of repeated extra work.
2. The more important reason is that it should not be Pocket ID's responsibility (or any other app's) to check if the DB is healthy, because there's nothing the apps can be do about that anyways. The orchestrator (Docker, K8s, etc) should be responsible for recovering/restarting the DB.

Let me elaborate on point 2. A principle of cloud-native architecture is that every dependency can fail at any time.

Downstream dependencies (such as Pocket ID calling Postgres) should always be prepared to handle transient failures. For example, if the DB goes down, when it comes back up, they should resume working automatically (Pocket ID does this already today, thanks to the DB libraries we use). Today we handle transient failures in a simplistic way: if the query fails, the user's request fails with a 500 response. This is fine for our use case, but there are other things that can be done to make apps more resilient including implementing retries (automatically retry the query 3 times if it fails), backoffs (wait an incremental amount of time between retries), and circuit breakers (if the DB fails 3 times in 10s, respond with a 500 right away and don't ping the DB again, to not increase its load; after a delay we can try again) ([more about these 3 things in this article](https://dilankam.medium.com/understanding-retries-exponential-backoffs-and-circuit-breakers-in-distributed-systems-4355db103505)).

In all of those scenarios, it nowhere says that apps should mark themselves as offline if an upstream dependency has failed. When an app marks itself as failed (health-checks fail), it tells the orchestrator (Docker, K8s) to restart the container. But if the failure is upstream, that is pointless since upon restarts the app would still not be working. That makes the orchestrator restart the app again, but because it uses an incremental backoff, every time it waits for longer. This means that if the DB comes back online, we are actually risking _extending_ the downtime for users, because they will need to wait for the DB _and_ the app to be restarted. In some cases the orchestrator will also give up if the app doesn't start after N attempts, so you risk having cascading failures.

The above holds true for healthchecks that are used for the orchestrators, which is what the /healthz endpoint is normally used for (in K8s terms, these are called liveness/readiness probes).

There could be a secondary use case where we want to show to users if the entire solution is working E2E. For example, it could be used to show in the UI a message like "Hey something seems not to be working". For that, yes, it could make sense to also check if the DB is responding.

If you want that, it can be done by running the `SELECT 1` query I suggested above, **but** I would recommend it to be a separate endpoint. For example, `/healthz/full` or `/healthz?full=1`

ItalyPaleAle: PS two interesting reads as to why an app's own healtcheck (liveness/readiness probes) shouldn't be concerned with upstream dependencies: https://www.reddit.com/r/kubernetes/comments/wk42l9/dont_do_this_with_your_k8s_health_checks/ https://www.reddit.com/r/kubernetes/comments/11q6et5/should_we_check_database_reachability_in_health/

kmendell: The interval is defined here: https://github.com/pocket-id/pocket-id/blob/cf3fe0be84f6365f5d4eb08c1b47905962a48a0d/docker-compose.yml#L13.

I guess my only thought based on what you just said, this seems like a "advanced" feature, (for lack of a better term). As some containers out there dont even have health checks. We provide the default health endpoint in the docker compose example file, and i would not assume its casuing issues as we have not seen issues. 

Moreover, As i get what your saying, as previoulsy mentioned pocket id isnt designed to be scaled (in the current state), so in my personal opinion i dont think it makes sense to bring up the more than one DB connection as the helthcheck 1. only lists the app config, 2. there should only really be one database connection. Im open to discussion on my point of view on this but this seems more catered towards scalability vs core functionality. As 90% of users that use pocket id may not even monitor the health status.

ItalyPaleAle: I'm not sure I understand, @kmendell, sorry.

My argument is that Pocket ID's health-checks on the /healthz endpoint, which is defined on the file you linked:

https://github.com/pocket-id/pocket-id/blob/cf3fe0be84f6365f5d4eb08c1b47905962a48a0d/docker-compose.yml#L12-L13

This is instrumenting Docker to invoke `/health` every 90s.

The problem I want to highlight is if `/health` is _also_ checking the health of upstream dependencies like DBs.

Let's say the DB fails. Ideally the DB runs in Docker so Docker would detect the failure and restart it. But it can take a few minutes.

If in the meanwhile we also make Pocket ID's `/health` endpoint report a failure, Docker will kill the Pocket ID container too. It will try to restart it, but if the DB is still down, Pocket ID will be restarted again. This will prolong the duration of a downtime for Pocket ID, because first the DB needs to start up, and then we need to wait for Docker to restart Pocket ID.

The more cloud-native way to handle upstream failures is to keep the apps that aren't failed themselves up, and have then ready to resume working upon the upstream dependency is available again. This prevents cascading failures.

Not so much about scaling, but more about uptime/availability.

> there should only really be one database connection

Just a note on this, that the DB drivers we use do manage a connection pool, at least for Postgres. It actually makes more than 1 connection behind the scene. This is necessary because each query/transaction takes over a connection, so if you want more than 1 query/transaction in parallel, the driver needs to maintain multiple connections. Gorm uses pgx to connect to Postgres, which [by default](https://github.com/jackc/pgx/blob/c92d0a90458bd137ac40e45316919d9d61576472/pgxpool/pool.go#L18-L19) maintains between 0 and 4 connections to the DB

stonith404: Okay, that makes sense, I didn't think of it that way. In this case I wouldn't test the database connection.

frontend/src/hooks.server.ts
```diff
@@ -23,33 +23,33 @@ const paraglideHandle: Handle = ({ event, resolve }) => {
 const authenticationHandle: Handle = async ({ event, resolve }) => {
 	const { isSignedIn, isAdmin } = verifyJwt(event.cookies.get(ACCESS_TOKEN_COOKIE_NAME));
 
-	const isUnauthenticatedOnlyPath = event.url.pathname.startsWith('/login') || event.url.pathname.startsWith('/lc')
-	const isPublicPath = ['/authorize', '/device', '/health'].includes(event.url.pathname);
-	const isAdminPath = event.url.pathname.startsWith('/settings/admin');
+	const path = event.url.pathname;
+	const isUnauthenticatedOnlyPath = path == '/login' || path.startsWith('/login/') || path == '/lc' || path.startsWith('/lc/')
```
ItalyPaleAle, line 27: Stumbled upon this while adding the `/healthz` route and made sure to add `/login` separately.

The reason is that `path.startsWith('/lc')` will also match `path.startsWith('/lcfoo')`, which doesn't exist today but it may one day. And that could cause problems (this has caused security issues in other projects!).

frontend/src/hooks.server.ts
```diff
@@ -23,33 +23,33 @@ const paraglideHandle: Handle = ({ event, resolve }) => {
...
+	const isPublicPath = ['/authorize', '/device', '/health', '/healthz'].includes(path);
+	const isAdminPath = path == '/settings/admin' || path.startsWith('/settings/admin/');
 
 	if (!isUnauthenticatedOnlyPath && !isPublicPath && !isSignedIn) {
 		return new Response(null, {
-			status: 302,
+			status: 303,
```
ItalyPaleAle, line 33: Small change. 302 redirects tell browsers to maintain the verb (so a POST request should, per spec, be redirected as POST). 303 tells clients to switch to GET ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/303))

docker-compose.yml
```diff
@@ -7,9 +7,9 @@ services:
       - 3000:80
     volumes:
       - "./data:/app/backend/data"
-    # Optional healthcheck  
+    # Optional healthcheck
     healthcheck:
-      test: "curl -f http://localhost/health"
+      test: "curl -f http://localhost/healthz"
```
ItalyPaleAle, line 12: While this changed to /healthz, which is more standardized, /health continues to work with no plan on being deprecated

kmendell:
Cool, First glance looks good , ill wanna test to make sure. Elias may beat me too it though :D

frontend/src/routes/healthz/+server.ts
```diff
@@ -0,0 +1,18 @@
+import type { RequestHandler } from '@sveltejs/kit';
+import axios from 'axios';
+
+export const GET: RequestHandler = async () => {
+	const backendOK = await axios
+		.get(process!.env!.INTERNAL_BACKEND_URL + '/healthz')
+		.then(() => true, () => false);
+
+	return new Response(
+		backendOK ? `{"status":"HEALTHY"}` : `{"status":"UNHEALTHY"}`,
```
stonith404, line 10:
```suggestion
		JSON.stringify({ status: backendOK ? 'HEALTHY' : 'UNHEALTHY' }),
```

Wouldn't this be a bit cleaner?

stonith404 (approved):

stonith404:
Thank you :)

frontend/src/routes/healthz/+server.ts
ItalyPaleAle: Sorry I just saw this. Since these are constants, essentially, it's probably faster to skip the overhead of re-encoding as JSON on every call. A bit of a micro-optimization.

PR #495, opened 2025-04-30
feat: OpenTelemetry tracing and metrics (#262)

daenney:
~~I haven't tested this yet, and the tracing exporter needs some additional work to be configurable so that folks can use the gRPC based exporter as well. But this should be the gist of it.~~

https://github.com/pocket-id/pocket-id/pull/495#issuecomment-2843412297

Fixes: https://github.com/pocket-id/pocket-id/issues/262


backend/internal/bootstrap/otel_boostrap.go
```diff
@@ -0,0 +1,85 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+)
+
+func defaultResource(ctx context.Context) (*resource.Resource, error) {
+	return resource.New(ctx,
+		resource.WithAttributes(
+			semconv.ServiceName("pocket-id-backend"),
+			semconv.ServiceVersion("0.0.1"),
```
daenney, line 22: Need to pull the version from somewhere. This is just a placeholder.

daenney:
You can call `otel.Meter()` in any other packages, or `otel.Tracer()` to define metrics or do something with the tracer. Within the http Handler lifecycle you can also use `trace.SpanFromContext(ctx)` with gives you a Span that you can call `TraceProvider().Tracer()` on to get the tracer (like if you want to start a child span).

kmendell:
Hey @daenney , Thank you for this! 

Just to clarify, are you planning on adding the rest of that logic? (mentioned in your comments), or did we need to look at that? :)

kmendell:
> @kmendell Yes, I am! I just pushed this out so folks could take a look and see if there's anything that you're not happy with.

So far it looks good, Will have to wait and see the finished product and testing as well. I dont use Otel, but once this is complete ill try to spin something up to test.

daenney:
For tracing, you can use https://github.com/CtrlSpice/otel-desktop-viewer. If you want the full stack, Grafana provides a very useful all-in-one image: https://github.com/grafana/docker-otel-lgtm.

I'll have the PR updated probably later today with some testing done.

daenney:
Here we go, the whole kit and caboodle :smile:.

This wires up:
* Tracing exporters, for OTLP GRPC, HTTP and stdout.
* Metrics exporters, for OTLP GRPC, HTTP, Prometheus exposition and stdout.

The stdout exporters should only ever be used for development.

I've added two `consts` for the tracer and meter name which should be used whenever any custom instrumentation is done. So when we're creating our own child spans on an existing span, or when registering custom metrics.

daenney:
For this to end up being somewhat more useful, in follow up PRs we probably should:
* Create some spans for the major operations on the tracer.
* Add a couple of custom metrics. The default histogram is very limited and doesn't even distinguish per path.

Another question that's worth thinking on is if we want `/metrics` on the default Gin router, or a custom one on a different port so folks can chose if they want to expose it or not. Alternatively, we could implement an option for basic auth on the `/metrics` endpoint instead, which is trivial to configure in Prometheus for scraping.

kmendell:
> Another question that's worth thinking on is if we want `/metrics` on the default Gin router, or a custom one on a different port so folks can chose if they want to expose it or not. Alternatively, we could implement an option for basic auth on the `/metrics` endpoint instead, which is trivial to configure in Prometheus for scraping.

If its on the default GIN Router it should be protected by the API so an API Key would need to be used to access it, does that work? Honestly ive never used either Otel or Prometheus. 


> I'll do a PR against the website/docs once this is merged for how to configure it.
You can make a PR on the website/docs preview branch if you would like, as thats the "staging" one.

daenney:
> > Another question that's worth thinking on is if we want `/metrics` on the default Gin router, or a custom one on a different port so folks can chose if they want to expose it or not. Alternatively, we could implement an option for basic auth on the `/metrics` endpoint instead, which is trivial to configure in Prometheus for scraping.
> 
> If its on the default GIN Router it should be protected by the API so an API Key would need to be used to access it, does that work? Honestly ive never used either Otel or Prometheus.

Sure. I assume that's just as simple as setting the `Bearer` header? If so that works and I can hook it after the auth middleware.

> > I'll do a PR against the website/docs once this is merged for how to configure it.
> 
> You can make a PR on the website/docs preview branch if you would like, as thats the "staging" one.

Ah cool!

kmendell:
@daenney i think it uses the x-api-key header if i remember right, i cant look at the very moment.

Dockerfile
```diff
@@ -25,6 +26,7 @@ RUN CGO_ENABLED=1 \
   GOOS=linux \
   go build \
   -tags "${BUILD_TAGS}" \
+  -ldflags="-X github.com/pocket-id/pocket-id/backend/internal/common.Version=${VERSION}" \
```
daenney, line 29: I've added this so the version is stamped in the binary, which lets code retrieve the value.

It can be set at build time using `docker build --build-arg VERSION=$(cat .version)` for example.

daenney:
Alright, I think I'm done :smile:.

backend/internal/bootstrap/otel_boostrap.go
```diff
@@ -0,0 +1,150 @@
...
+	tp := sdktrace.NewTracerProvider(options...)
+
+	otel.SetTracerProvider(tp)
+	otel.SetTextMapPropagator(
+		propagation.NewCompositeTextMapPropagator(
+			propagation.TraceContext{},
+			propagation.Baggage{},
+		))
+
+	f := func() { //nolint:contextcheck
```
ItalyPaleAle, line 84: Once #493 is merged it could be used to run OTel as a background service

backend/internal/common/env_config.go
```diff
@@ -76,4 +90,35 @@ func init() {
 	if parsedAppUrl.Path != "" {
 		log.Fatal("PUBLIC_APP_URL must not contain a path")
 	}
+
+	validTraceExporters := []string{"http", "grpc", "stdout"}
```
ItalyPaleAle, line 94: For HTTP and GRPC, shouldn't we rely on the env var [`OTEL_EXPORTER_OTLP_PROTOCOL`](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol) (or the more specific `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` and `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL`)?

So `METRICS_EXPORTERS` and `TRACES_EXPORTERS` would only used to enable stdout/prometheus, and all OTEL stuff would be configured via the `OTEL_*` env vars?

backend/internal/common/env_config.go
```diff
@@ -76,4 +90,35 @@ func init() {
...
+	validMetricsExporters := []string{"http", "grpc", "stdout", "prometheus"}
+
+	validateExporters(EnvConfig.TracesExporters, validTraceExporters, "TRACES_EXPORTERS")
+	validateExporters(EnvConfig.MetricsExporters, validMetricsExporters, "METRICS_EXPORTERS")
+}
+
+func validateExporters(exps []string, valid []string, env string) {
+	if len(exps) != 0 {
+		if EnvConfig.AppEnv == "production" && slices.Contains(exps, "stdout") {
+			log.Fatal("stdout exporters must not be used in production")
```
ItalyPaleAle, line 104: Is this really necessary? While I agree it is not a good idea to do this, why crashing the app?

kmendell:
I'll try to review this tonight, as least some of it.

backend/internal/common/env_config.go
daenney: It's not necessary, but it avoids people running with bad configurations.

The output formats of the stdout exporters are not stable, and as such can never be relied on. They really should not be enabled unless you know what you're doing, and leaving them accidentally enabled is far from ideal.

backend/internal/common/env_config.go
daenney: Mmm, good point. I could use the `autoexport` package to make that simpler. I'll take a look at that.

daenney:
Alright, this changes things a bit based on a comment from @ItalyPaleAle.

There's now 2 config options, MetricsEnabled and TracingEnabled. For each of those, when set, a trace or meter provider is installed with the Span Exporter and Meter Reader automatically configured using the autoexport package instead. That makes things much easier.

I've explicitly opted for the boolean toggles, because the NoopSpanExporter and NoopMetricReader aren't currently exported by the OTEL packages so I can't easily install them as the fallbacks using `autoexport.WithFallback...` options.

backend/internal/common/env_config.go
daenney: This is gone now since everything is configured by the OTEL variables. Since this package doesn't own them I'm no longer doing any validation.

kmendell:
I tried testing this last night i could not get it to work with the local otel viewer (probably user error honestly)....

stonith404:
@daenney Thank you, it's working for me. I think this would be ready to merge as soon as #493 is merged and OTel is migrated to a background service.

As soon as this PR is merged we would have to add the relevant environment variables like `METRICS_ENABLED`, `TRACING_ENABLED` and `OTEL_EXPORTER_OTLP_ENDPOINT` to the docs. Are there more environment variables from the OTel SDK that should be mentioned?

kmendell:
@stonith404 I did have him make a PR on the `preview` branch with the docs last night, but those will probably need to be updated

daenney:
I'll wait for #493 to land before making any further changes. I'll update the docs site again once this is in, otherwise we'll be playing a game of whack-a-mole between PRs.

daenney:
I just realised that autocert, when used with `OTEL_METRICS_EXPORTER=prometheus` automatically starts up a second http server bound to localhost:9464. That stuff is configurable, but not not starting the separate http server. Idk if we want that. If not I'd have to go back to the previous option where I set up all the metric readers by hand.

ItalyPaleAle:
> I just realised that autocert, when used with `OTEL_METRICS_EXPORTER=prometheus` automatically starts up a second http server bound to localhost:9464. That stuff is configurable, but not not starting the separate http server. Idk if we want that. If not I'd have to go back to the previous option where I set up all the metric readers by hand.

TIL too, but looks like it's documented here: https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport#NewMetricReader

Two options:

1. Accept that prometheus will be on its own server. I actually think this is the most common approach, to have a separate listener for the metrics server (and it's normally un-authenticated). Port can be configured with `OTEL_EXPORTER_PROMETHEUS_PORT`
2. You can create an `init()` method (which will be executed when the package is loaded in memory, so before OTel is initialized) to check if the env var is set to `prometheus` and remove it from there

daenney:
Honestly I think the second server, which is how I normally also do things, is what I'd prefer anyway. That avoids any accidental exposure to the outside world, and if you do do it then you're on the hook for ensuring authentication is properly configured. Since you can also configure OTLP push metrics instead, which is what it'll do by default, that seems kinda fine.

stonith404:
#493 is now merged.

kmendell:
@daenney i did end up getting it working, it was a combination of me missing a build arg and being overly tired :D

ItalyPaleAle:
@daenney fixed merge conflicts here and use the new shutdown: https://github.com/daenney/pocket-id/pull/1

daenney:
I think Github's lost the plot with the merge conflicts now :smile:

daenney:
There, much better.

daenney:
There seems to be something up with the Geolite service (or I botched my rebases and rework)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xb9145c]

goroutine 1 [running]:
github.com/pocket-id/pocket-id/backend/internal/service.(*GeoLiteService).DisableUpdater(...)
	/home/daenney/Development/github.com/pocket-id/backend/internal/service/geolite_service.go:61
github.com/pocket-id/pocket-id/backend/internal/job.(*Scheduler).RegisterGeoLiteUpdateJobs(0xc000443300?, {0x1b4bae0?, 0xc00012e500?}, 0xc00045c7e0?)
	/home/daenney/Development/github.com/pocket-id/backend/internal/job/geoloite_update_job.go:17 +0x1c
github.com/pocket-id/pocket-id/backend/internal/bootstrap.registerScheduledJobs({0x1b4bae0, 0xc00012e500}, 0xc00035ac90, 0xc0002178c0, 0xc000443300)
	/home/daenney/Development/github.com/pocket-id/backend/internal/bootstrap/scheduler_bootstrap.go:17 +0xbd
github.com/pocket-id/pocket-id/backend/internal/bootstrap.Bootstrap()
	/home/daenney/Development/github.com/pocket-id/backend/internal/bootstrap/bootstrap.go:47 +0x205
main.main()
	/home/daenney/Development/github.com/pocket-id/backend/cmd/main.go:14 +0x13
exit status 2
```

ItalyPaleAle:
It could have been me. I'll take a look later today

daenney:
@ItalyPaleAle No need, fixed. It was my bug. I had accidentally axed the line that initialised the geolite service.

kmendell (approved):

kmendell:
Look good to me! @stonith404 if all looks good to you as well this should be good to merge.

daenney:
All good from my side. I'll follow up with some PRs to add a few custom metrics and trace points.

jfroy:
:rocket:

PR #513, opened 2025-05-07
fix: handle CORS correctly for endpoints that SPAs need

### LINKED ISSUE #329: 🚀 Feature: Allow setting additional origins
### Feature description

a possibility to extend the `Access-Control-Allow-Origin` header to allow other origins

- via env
- or even better: use the `OIDC Clients -> Callback URLs` (remove subpath if necessary) and allow them by default.
  -  the urls defined there already have the permission to login, they should also be allowed to use the other endpoints (wellknown/userinfo)

### Pitch

I have Pocket Id running via Docker and is accessible at `https://auth.dev.localhost` which is fine and a frontend (React / SPA) running at `https://apps.dev.localhost`.
I also have an oidc client configured with the “public” flag enabled and wanted to add the client to the frontend for authentication. for this I am using oidc-react.
now the lib tries to reach the `.well-known/openid-configuration` endpoint to load the metadata but this does not work because the `Access-Control-Allow-Origin` header only allows the `PUBLIC_APP_URL` which points to  `https://auth.dev.localhost` - the same for the other endpoints like the userinfo endpoint. 

so it is actually hardly possible to use the provider in the frontend in a decent way

stonith404:
fixes #329

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -129,9 +129,6 @@ func (oc *OidcController) authorizationConfirmationRequiredHandler(c *gin.Contex
 // @Success 200 {object} dto.OidcTokenResponseDto "Token response with access_token and optional id_token and refresh_token"
 // @Router /api/oidc/token [post]
 func (oc *OidcController) createTokensHandler(c *gin.Context) {
-	// Disable cors for this endpoint
-	c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
-
 	var input dto.OidcCreateTokensDto
 	if err := c.ShouldBind(&input); err != nil {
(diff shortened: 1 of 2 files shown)


kmendell (commented):
@stonith404 Do we want to make a ENV Variable to make this configurable?

kmendell (approved):

kmendell:
Looks good to me :) Someone from the issue tested, and seems like it fixed it for them.

ovizii:
Excuse my ignorance, but this seems fixed and merged, but I cannot find anything on how this works in the documentation or in any changelogs. Can someone clarify if we should still use ghcr.io/pocket-id/pocket-id:development-cors or what to do?

PR #520, opened 2025-05-09
refactor!: serve the static frontend trough the backend

stonith404:
Fixes: https://github.com/pocket-id/pocket-id/issues/536

## TODO
- [x] [Cache static assets](https://github.com/pocket-id/pocket-id/pull/520#discussion_r2081971318)
- [x] Setup GitHub action to build binaries and attach them to new releases

## Breaking changes
The following breaking changes should be documented in version `1.0.0`:
- Port in the "ports" section and in the health check in the `docker-compose.yml` has to be changed to `1411`.
- `PUBLIC_UI_CONFIG_DISABLED` has been renamed to `UI_CONFIG_DISABLED`
- `CADDY_DISABLED`  has been removed. If you've disabled Caddy, you probably have path mappings to `/.well-known` and `/api` in your reverse proxy configuration. These mappings aren't necessary anymore, your reverse proxy should now solely point to Pocket ID on port `1411` (or the port set with the `PORT` variable).
- `CADDY_PORT` has been removed and replaced by `PORT`
- `BACKEND_PORT` has been removed and replaced by `PORT`
- The behavior of `PORT` has been changed. This variable now doesn't change the port of the frontend anymore, but for the whole application.


Dockerfile
```diff
@@ -31,25 +34,20 @@ RUN CGO_ENABLED=1 \
   .
 
 # Stage 3: Production Image
-FROM node:22-alpine
-# Delete default node user
-RUN deluser --remove-home node
+FROM alpine
```
ItalyPaleAle, line 37: In a future PR, let's consider moving this to a [distroless](https://github.com/GoogleContainerTools/distroless) container image?

(For now we'd need a "base" image such as `gcr.io/distroless/base-debian12` - we can't use `static` because we depend on CGo due to the use of SQLite... that can be worked around too, but it's yet another separate task)

Pros:

- Much smaller image
- Base image contains a lot less software so there's less risk of a security vulnerability, and needs to be updated less frequently

Cons:

- By default it doesn't include a shell, so you can't get into the container. There are variants of the distroless image that do include a shell, but they'd need to be published separately.

ItalyPaleAle (commented):
Aside from some minor comments, this LGTM and I like it. THanks for putting all the effort to refactor the frontend so it doesn't depend on a server.

THere are some other things that can be done to make this even better, and we can do this in a follow-up PR. A quick win for example is to add a service worker on the static front-end so clients can cache it better locally. Additionally, the SW could cache static things such as the JWKS document, which otherwise would be requested very frequently. All these things can be done in other PRs.

Dockerfile
kmendell: I think alpine should work for now. Maybe in the future when stuff isnt in a constant chnaging state (ie 1.0) we could look at that but since stuff is changing still with features being added removed etc, having that alpine base is nice as it has utilities for debugging if needed. Plus as we disscussed before there is a acceptable risk factor in containers and in any application.

As far as size the distroless image only seems to be 5MB smaller at its base, which isnt a big enough reduction in my opinion, and this refactor already decresed the size to a very small image. So i dont think i see the need currently.

backend/internal/common/env_config.go
```diff
@@ -52,7 +52,7 @@ var EnvConfig = &EnvConfigSchema{
 	UploadPath:               "data/uploads",
 	KeysPath:                 "data/keys",
 	AppURL:                   "http://localhost",
-	Port:                     "8080",
+	Port:                     "80",
```
ItalyPaleAle, line 55: I liked port 8080 because it's non-privileged. Makes it possible to run pocket ID as non-root user...

stonith404: Yeah port 80 is not optimal, but it would break every installation, if we change the port. I'm not sure if this is worth it.

backend/frontend/frontend_included.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+
+		// Try to serve the requested file
+		path := strings.TrimPrefix(c.Request.URL.Path, "/")
+		if _, err := fs.Stat(distFS, path); os.IsNotExist(err) {
+			// File doesn't exist, serve index.html instead
+			c.Request.URL.Path = "/"
+		}
+
+		fileServer.ServeHTTP(c.Writer, c.Request)
```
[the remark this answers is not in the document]
stonith404: Thanks, I'll look into it.

backend/internal/common/env_config.go
ItalyPaleAle: Yes, but it wouldn't be the only breaking change in this release. May as well just rip the band-aid?

stonith404: The release currently would only break installations that have Caddy disabled. But yeah maybe it makes sense to change the default port because I have to create a breaking change release anyway.

Dockerfile
stonith404: I like the idea of a distroless image but this probably needs some more work. For example the `create-one-time-access-token.sh` script would have to migrated to Go. 

I won't do this in this PR, but in the future this would be a good idea.

backend/internal/common/env_config.go
kmendell: i have been testing this throughly and nothing has been a major breaking change yet, in my prod server. beside me changing it to port 8080 but that's it

backend/internal/bootstrap/bootstrap.go
```diff
@@ -21,7 +21,6 @@ func Bootstrap() error {
 	initApplicationImages()
 
 	// Perform migrations for changes
-	migrateConfigDBConnstring()
```
ItalyPaleAle, line 24: If you're removing this (which I agree with), should we remove `migrateKey()` too? That's been around for even longer

kmendell: Yeah good point probably, Ill take a look at it. :)

backend/frontend/frontend_included.go
```diff
@@ -37,3 +38,36 @@ func RegisterFrontend(router *gin.Engine) error {
...
+	if ifModifiedSince := r.Header.Get("If-Modified-Since"); ifModifiedSince != "" {
+		ifModifiedSinceTime, err := time.Parse(http.TimeFormat, ifModifiedSince)
+		if err == nil && f.lastModified.Before(ifModifiedSinceTime.Add(1*time.Second)) {
+			// Client's cached version is up to date
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
+	}
+
+	w.Header().Set("Last-Modified", f.lastModified.UTC().Format(http.TimeFormat))
```
ItalyPaleAle, line 69: Nit: `f.lastModified.UTC().Format(http.TimeFormat)` is a constant and can be cached in the object (in `NewFileServerWithCaching`)

stonith404: Makes sense, thanks.

kmendell (approved):
@stonith404 Everything looks good to me, and ive been using this for the last week or so.  The only changes i think we still need to add is to update the example docker-compose.yml and the .env.example to make sure those are correct. I can do that later if you dont beat me to it

ItalyPaleAle (approved):

ItalyPaleAle:
A few more tweaks: https://github.com/pocket-id/pocket-id/pull/533

gnur:
Right now the focus seems to be on the docker image, but this does allow it to be deployed as a single binary. Maybe the instructions on how to build it can be included and also add the binaries for all arches as a build step to publish with the releases? 

(Or, what would be the easiest way to build this on my laptop right now so I can start using it? :D )

scripts/development/build-binaries.sh
```diff
@@ -0,0 +1,57 @@
...
+build_platform "linux-386" "linux" "386" ""
+build_platform "linux-arm64" "linux" "arm64" ""
+build_platform "linux-armv7" "linux" "arm" "7"
+
+# macOS builds
+build_platform "macos-x64" "darwin" "amd64" ""
+build_platform "macos-arm64" "darwin" "arm64" ""
+
+# Windows builds
+build_platform "windows-x64" "windows" "amd64" ""
```
ItalyPaleAle, line 50: Suggest keeping windows off for now. I don't believe we support Windows and its `\` as path separator. It would likely require some more testing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+jobs:
+  build-docker-image:
+    runs-on: ubuntu-22.04 # Using an older version because of https://github.com/actions/runner-images/issues/11471
```
ItalyPaleAle, line 10: Did you experience this issue or just in theory?

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
ItalyPaleAle, line 6:
```suggestion
      - "v*"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+      packages: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ghcr.io/${{ github.repository }}
```
ItalyPaleAle, line 22: Before this step, add one like:

```yaml
      # Lowercase REPO_OWNER which is required for containers
      - name: Set lowercase REPO_OWNER
        run: |
          REPO_OWNER=${{ github.repository_owner }}
          echo "REPO_OWNER=${REPO_OWNER,,}" >>${GITHUB_ENV}
```

Then change this to `ghcr.io/${{ env.REPO_OWNER }}/pocket-id` (also note you need to add the image name `/pocket-id`)

The reason is that Docker images' names must all be lowercased, and if your username has an uppercase letter (say, ItalyPaleAle :) ) that breaks the flow in forks

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
```
ItalyPaleAle, line 57: We should pin a specific version, not just LTS?

.github/workflows/release.yml
kmendell: Whats the purpose of this? based on my understanding it wont cpature tags pushed to x.x.1 or x.2.1 etc with your logic.

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+        working-directory: frontend
+        run: npm run build
+      - name: Build binaries
+        run: sh scripts/development/build-binaries.sh
+      - name: Upload binaries to release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release upload ${{ github.ref_name }} backend/.bin/*
+
+  publish-release:
```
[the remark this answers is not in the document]
kmendell: Release notes are created by our release script. And manually uploading binaries takes time, and manualy effort in a process that we woudl want automated. I dont see why its an issue?

.github/workflows/release.yml
kmendell: LTS is technially pinned. But it will pull the latest LTS with security fixes which is important. 

I know Node js 24 is ging into LTS later this year but at the point it still should be fine.

.github/workflows/release.yml
ItalyPaleAle: Yes it will capture everything that start with `v`

.github/workflows/release.yml
ItalyPaleAle: People can fork the project and make their own builds if they wish to. I ran into this problem when I was working on Dapr and needed my own builds to test things

.github/workflows/release.yml
stonith404: Yes, I had this issue in the past. Maybe it's fixed though, because this issue doesn't seem really active anymore. 

I'll remove it and add it back if the issue still occurs.

.github/workflows/release.yml
stonith404: Does this really make sense? All of our release tags follow the pattern `vx.x.x`. With your suggested change it would run for every tag that starts with "v" which could result that the action runs for tags that don't mark a new version.

.github/workflows/release.yml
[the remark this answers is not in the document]
stonith404: I would keep this automatic. The release flow is like this:
1. I run `scripts/development/create-release.sh` locally, which does the following:
     - Generates release notes based on the conventional commits messages
     - Tags the latest commit with the version
     - Pushes the tag
     - Creates a draft release on GitHub
2. GitHub runs the `release.yml` action because of the new tag, which does the following:
     - Builds and pushes Docker image
     - Builds and attaches the binaries to the draft release
     - If the build of the Docker image and binaries were successful the release gets published

.github/workflows/release.yml
stonith404: That's interesting, but it seems that `docker/metadata-action` [does this automatically](https://github.com/docker/metadata-action/tree/v5/?tab=readme-ov-file#image-name-and-tag-sanitization).

.github/workflows/release.yml
stonith404: I think it makes sense to specify a particular version because we are also using a specific version in the Dockerfile. 

This could result in the tests running successfully while the build fails in this action.

.github/workflows/release.yml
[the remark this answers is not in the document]
ItalyPaleAle: Sounds good

.github/workflows/release.yml
ItalyPaleAle: Oh great. Didn't use to be the case

.github/workflows/release.yml
stonith404: Currently, we only create tags for releases, but I still believe this pattern is useful because it can prevent accidental release creations if we ever use tags for something else.

stonith404:
The PR should be ready now. 

We have to make sure that we create the `v1.0.0` release right after the PR gets merged because the `docker-compose.yml` and `.env` have changed and users use the these files directly from the main branch.

`v1.0.0` should be ready after the migration guide is finished and the installation instructions of the standalone have been adapted. As I have my final exams next week, I probably don't have the time to create the new release in the next few days.

stonith404:
I'm merging this into the `v1.0.0` branch.

PR #530, opened 2025-05-14
refactor: switch SQLite driver to pure-Go implementation

ItalyPaleAle:
This PR changes the driver used to connect to SQLite to `github.com/glebarez/sqlite`.

The main difference with the previous driver (`gorm.io/driver/sqlite`) is that the new one uses a pure-Go implementation of SQLite based on `modernc.org/sqlite`.

There are three main benefits with making this change:

1. It makes it possible to build Pocket ID with CGo disabled, which means that cross-compilation (building for different archs or OS's) becomes a lot simpler, requiring nothing more than the standard Go compiler
2. Binaries are fully statically-compiled, with no dependency on libc. This makes it simpler to run across all Linux distributions, even those that use different libc implementations (like Alpine)
3. It helps with a possible future goal of running on distroless containers

The pure-Go driver is slightly slower, more noticeable with very large databases, larger than Pocket ID users are expected to have.

The `modernc.org/sqlite` package is built by "transpiling" the C library into Go code, and they have extensive test suites to ensure the behavior is consistent. The same package is used by many apps successfully in production, so stability shouldn't be a concern.

The "biggest" user-facing change is that `github.com/mattn/go-sqlite3` supports some additional args in the query string which need to be converted to PRAGMA statements with the `modernc.org/sqlite` package. To avoid confusing users (and potentially introducing yet another breaking change), I've added a function that converts the common query string args into the format supported by the modernc package. Most users would not need to worry about it, as wery few should change the DB connection string.


ItalyPaleAle:
Moving back to draft since there are some additional changes required with parsing timestamps

ItalyPaleAle:
Should be fixed now

stonith404 (approved):
Looks good to me, thank you!

PR #532, opened 2025-05-14
fix: flaky unit test in db_bootstrap_test

ItalyPaleAle:
Go's url.Values can be serialized in random order, which the test case didn't account for

Fixup from #530


kmendell:
Thanks!

Issue #531, opened 2025-05-14
🐛 Bug Report: Animations should be a lot quicker


ItalyPaleAle:
### Reproduction steps

Animations were introduced in a recent release, but they tend to be relatively slow, up to 500ms.

While they are nice, slow animations artificially slow down the UX and make the app feel slower. This was also covered in a [recent Syntax episode](https://youtu.be/nxzVZ7FYdwE?t=360&si=hj-GyZR8yaJY6uge).

Ideally all animations should be quick, 100 or 200ms at most.

### Expected behavior

Animations should be 200ms or faster

### Actual Behavior

Animations are slow, up to 500ms

### Version and Environment

main branch

mitchplze:
Not sure if you saw but [disabling animations](https://pocket-id.org/docs/configuration/environment-variables#overriding-the-ui-configuration) altogether is possible with `DISABLE_ANIMATIONS=true` as a workaround.

ItalyPaleAle:
Yes, I disabled animations in my setup. This still remains as a bug IMHO.

kmendell:
@ItalyPaleAle The timing is set to 150ms with a delay of 50ms, is there a certain spot you are seeing this take up to 500ms?

kmendell:
I set it to a max of 300ms here this should speed up the animations d534fad1cf4fe16bfd94e6ec2ccdc83864c70dd8

kmendell:
I did 300 over 200 as 200 was almost instant, 300 is still fast and still looks decent.

ItalyPaleAle:
@kmendell thanks but I think i see a lot of slower animations, for example:

https://github.com/pocket-id/pocket-id/blob/d6104bbb351672e2b3d6aa2c09083359d0428d94/frontend/src/routes/authorize/components/client-provider-images.svelte#L25
https://github.com/pocket-id/pocket-id/blob/d6104bbb351672e2b3d6aa2c09083359d0428d94/frontend/src/routes/authorize/components/client-provider-images.svelte#L34
https://github.com/pocket-id/pocket-id/blob/d6104bbb351672e2b3d6aa2c09083359d0428d94/frontend/src/routes/authorize/components/client-provider-images.svelte#L42
https://github.com/pocket-id/pocket-id/blob/d6104bbb351672e2b3d6aa2c09083359d0428d94/frontend/src/routes/authorize/components/client-provider-images.svelte#L47
[...]

PR #537, opened 2025-05-16
v1.0.0

stonith404:
Fixes: https://github.com/pocket-id/pocket-id/issues/531


kmendell:
@stonith404 I cant add it right now but this selector should work for the Toasts for the e2e tests 		
`await expect(page.locator(`li[data-sonner-toast][data-type="success"] div[data-title]:has-text("add text here")`)).toBeVisible();
`

kmendell (approved):
Looks good! :)

PR #540, opened 2025-05-17
refactor: Replace create-one-time-access-token script with in-app functionality (v1.0)

ItalyPaleAle:
As discussed here: https://github.com/pocket-id/pocket-id/pull/533#discussion_r2093286087

Removes the create-one-time-access-token.sh script and embeds the functionality within the Go binary.

It can now be invokes as `pocket-id one-time-access-token <user name or email>`.

This helps because:

1. Reduces complexity and improves maintenability because there's no duplicate functionality in the shell script
2. Ensures the shell script works well with SQLite even after changing the connection string format


backend/internal/model/user.go
```diff
@@ -84,3 +85,24 @@ type OneTimeAccessToken struct {
 	UserID string
 	User   User
 }
+
+func NewOneTimeAccessToken(userID string, expiresAt time.Time) (*OneTimeAccessToken, error) {
```
stonith404, line 89: I would prefer to add this to the user service instead. In my opinion the models shouldn't have any logic.

ItalyPaleAle: The problem with putting it in the user service is that it forces the service to be initialized, which requires initializing a bunch of other services.

Unless you mean just moving the function to the user service **file**, but not place it within the service itself?

stonith404: Ok yeah good point. 

Do you think it's a good idea to add it into the user service file without adding it as a method on the user service struct? Alternatively we could also add it as an util function. What would you suggest?

ItalyPaleAle: I wouldn't recommend adding it as util function as that could create circular dependencies.

I will move it to the user service file, that should be ok

PR #541, opened 2025-05-17
feat: Update release pipelines (1.0)

ItalyPaleAle:
Various updates to the release pipeline to make builds fully reproducible, publish attestations, and speed-up container builds.

1. Create an additional `Dockerfile-prebuilt` which includes binaries pre-built in the pipeline, rather than re-building the binary for each arch (which can be very slow when running under emulation with QEMU)
2. Ensure the release pipeline builds binaries with CGo disabled
3. Pass the ldflag `-buildid` with the version. In addition to `-trimpath` (already present) makes builds fully reproducible
4. Publish provenance attestation for the binaries (attestation is published as a pipeline artifact) and container images (pushed to the container registry) using SLSA (see: https://github.com/actions/attest-build-provenance/)
   - Provenance attestation allows users to check that the binary they are running is exactly the same that was built by the pipeline, and had not been tampered with. A very nice thing for a security-related app like this!
5. Run `chmod +x` on the `.sh` files so it doesn't need to be done when building containers (and they can run locally without issues)


.github/workflows/release.yml
```diff
@@ -6,73 +6,94 @@ on:
...
-          labels: ${{ steps.meta.outputs.labels }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          images: |
+            ghcr.io/${{ github.repository }}/pocket-id
+          tags: |
+            type=semver,pattern={{version}},prefix=v
+            type=semver,pattern={{major}}.{{minor}},prefix=v
+
+      # Lowercase REPO_OWNER which is required for containers
```
kmendell, line 50: As discussed here: https://github.com/pocket-id/pocket-id/pull/520#discussion_r2093547147. This is not needed

ItalyPaleAle: Correct for the docker actions. But it’s required here: https://github.com/pocket-id/pocket-id/pull/541/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R85

.github/workflows/e2e-tests.yml
```diff
@@ -59,6 +59,9 @@ jobs:
           node-version: 22
           cache: "npm"
           cache-dependency-path: frontend/package-lock.json
+      - uses: actions/setup-go@v5
```
stonith404, line 62: Is this necessary, because we don't use Go in this action?

ItalyPaleAle: Oh right, that was a mistake. I'll revert

stonith404:
I don't think that building binaries separately before copying them into the Docker image speeds things up. In my test repository, binary compilation alone took approximately 13 minutes. The current implementation forces the Docker image build to wait until all binaries are completely built.

By integrating binary compilation directly into the Docker build process, we can run binary compilation and image building at the same time.

stonith404:
Actually you're right, building the binary in Docker for ARM takes really a long time:
```
#40 [linux/arm64 backend-builder 8/8] RUN CGO_ENABLED=1   GOOS=linux   go build   -tags ""   -o /app/backend/pocket-id-backend   .
#40 DONE 1140.3s
```

ItalyPaleAle:
Yes, the reason is that with the multi-stage approach, the entire build runs using emulation (QEMU), so even building the front-end is repeated twice but when using emulation it's a lot slow.

Speed was not the main goal of this change, however, just a nice-to-have. Primarily wanted to have a simplified build where all builds, for all archs, are built at the same time (without emulation), and signed. This gives us the guarantee now that the binary that is inside the container is the same one that's outside, and which is signed (SLSA).

.github/workflows/release.yml
```diff
@@ -6,73 +6,94 @@ on:
...
         with:
-          context: .
-          platforms: linux/amd64,linux/arm64
-          push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          images: |
+            ghcr.io/${{ github.repository }}/pocket-id
```
stonith404, line 45: `ghcr.io/${{ github.repository }}/pocket-id` results in `ghcr.io/pocket-id/pocket-id/pocket-id`

Instead of storing the `REPO_OWNER` as an env variable I would just create a `DOCKER_IMAGE_NAME` variable that we can use here and in the attestion.

scripts/development/build-binaries.sh
```diff
@@ -8,7 +8,7 @@ build_platform() {
     os=$2
     arch=$3
     arm_version=${4:-""}
-    pocket_id_version=$(cat ../.version)
+    pocket_id_version=$(cat ../.version | sed 's/^\s*\|\s*$//g')
```
ItalyPaleAle, line 11:
```suggestion
    # "sed" is used to remove leading or trailing whitespace characters
    pocket_id_version=$(cat ../.version | sed 's/^\s*\|\s*$//g')
```

.github/workflows/release.yml
```diff
@@ -31,6 +31,13 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
+      - name: Set DOCKER_IMAGE_NAME
+        run: |
+          # Lowercase REPO_OWNER which is required for containers
+          REPO_OWNER=${{ github.repository_owner }}
+          echo "REPO_OWNER=${REPO_OWNER,,}" >>${GITHUB_ENV}
+          DOCKER_IMAGE_NAME="ghcr.io/${REPO_OWNER}/pocket-id"
```
stonith404, line 39:
```suggestion
          DOCKER_IMAGE_NAME="ghcr.io/${REPO_OWNER}/pocket-id"
          echo "DOCKER_IMAGE_NAME=${DOCKER_IMAGE_NAME}" >>${GITHUB_ENV}
```

ItalyPaleAle: Woops, not enough caffeine today clearly

ItalyPaleAle: However the `${REPO_OWNER,,}` needs to stay, because that's how shell scripts lowercase the string (one more example of how the syntax of bash is incredibly clear :)

Issue #361, opened 2025-03-19
🚀 Feature: Support private_key_jwt client authentication method per RFC7523 (Security)


ItalyPaleAle:
### Feature description

RFC7523 describes authenticating clients using a JWT, signed by a third-party, instead of a client secret.

In the case most applicable to Pocket ID, this would be used as part of the Authorization Code flow when invoking the `/token` endpoint.

When making a call to the `/token` endpoint with `grant_type=authorization_code`, instead of including a `client_secret`, callers include two more parameters:

- `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer` (this is a constant)
- `client_assertion=...` value is a JWT issued by a third-party with some requirements as specified in [section 3 or the RFC](https://datatracker.ietf.org/doc/html/rfc7523#section-3)

The identity server uses the assertions passed by the client to validate their identity instead of a pre-shared client secret.

From a technical standpoint, this requires making Pocket ID accept for each client (app) a list of identity providers that are authorized to sign tokens to prove the clients' identities.

Eventually, the goal is to make it possible to use things such as managed identities when running in cloud environments, workload identity on K8s, or SPIFFE.

Links:

- RFC 7523: https://datatracker.ietf.org/doc/html/rfc7523
- Sample from the Authlete docs: https://www.authlete.com/kb/oauth-and-openid-connect/client-authentication/client-auth-private-key-jwt/
- Discussion in Keycloak: https://github.com/keycloak/keycloak/discussions/14823
- Docs from Okta: https://developer.okta.com/docs/api/openapi/okta-oauth/guides/client-auth/#jwt-with-private-key

### Pitch

**Secrets are bad for security**. It takes a lot of education to teach users how to handle them properly, and even then people misplaces secrets all the time. Just one data point: in 2022, [researchers found](https://www.gitguardian.com/files/the-state-of-secrets-sprawl-report-2023) 10 million secrets in GitHub repositories.

Even when users do all things right, secrets are often stored _somewhere_ in plain text, which makes them susceptible to leaking in other ways.

The industry is moving away from secrets in general. For example, at Microsoft where I used to work, in 2024 there was a massive effort to get rid of all shared secrets from applications, using other solutions including managed identities instead. External users were encouraged to do the same.

Popular examples include support for OIDC in GitHub Actions, which is widely used already: https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments

kmendell:
[...]

PR #550, opened 2025-05-19
refactor: some clean-up in OIDC service and controller (1.0)

ItalyPaleAle:
As I started working on #361, I wanted to split out a small initial refactor/code cleanup, so the upcoming main PR wouldn't contain irrelevant refactorings.

- In `OidcService`, `CreateToken` and all the functions it calls now accept a struct and return a struct. This avoids passing up to 5 arguments and having up to 5 return values (some functions did not use all input args and did not return all 5 outputs).
- Define constants for the grant types, including `urn:ietf:params:oauth:grant-type:device_code`
- When verifying client credentials, if the user passed a client secret we always validate it, even for public clients (before, we entirely skipped the validation if the client was public, which was not the ideal behavior (if the client does pass a credential, it should be validated or rejected)


stonith404 (approved):
Thank you :)

PR #371, opened 2025-03-21
feat: russian localization added and enabled

savely-krasovsky:
I translated it manually, checking the source codes. Tested and verified translation using locally deployed Docker container. Translation file is the same as in #368.

PR #554, opened 2025-05-21
feat: polish translations

mikolaj92:
I've added a polish translations based on the current english file from main and https://github.com/pocket-id/pocket-id/pull/371 pull request.


stonith404:
Thank you!

PR #566, opened 2025-05-25
Support for JWT bearer assertions for client authentication (backend changes)

ItalyPaleAle:
This PR contains the **backend** changes for #361 

## TLDR

- Implements support for authenticating OAuth2 clients using JWT bearer client assertions in lieu of a client secret
- Complies with [RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523)

## Goals

Full discussion is in #361, but to recap...

We want to enable confidential clients (OAuth2 applications) to perform a token exchange without having to use a client secret, because client secrets are long-lived shared secrets and are not the best solution for security.

- When clients (apps) exchange an auth code for an access token, by invoking the `/api/oidc/token` endpoint, normally they include a client ID and client secret. With this change, instead of a client secret they can include `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer` (this is a constant defined in the RFC) and a `client_assertion=...` containing a JWT signed by an **external** identity server, which is issued to the application to identify itself
- For example, apps that run on cloud services could use things like Azure managed identities or AWS IAM service roles. Apps that run on Kubernetes can use a (bound) service account token. Or it could be a token issued with SPIRE. (Note: doesn't mean Pocket ID must run on the cloud or on K8s, it's independent from where Pocket ID is hosted)

## References

- RFC 7523, specifically [section 2.2](https://datatracker.ietf.org/doc/html/rfc7523#section-2.2) ("Using JWTs for client authentication"), and additional requirements in section 3
- Additional context from [IETF draft for Workload Identity Practices](https://www.ietf.org/archive/id/draft-ietf-wimse-workload-identity-practices-00.html)
- Used the Microsoft Entra ID implementation as reference implementation ([docs](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-access-token-with-a-certificate-credential))

## How this works

Users should be able to add one or more federated identities for each app in Pocket ID. For each federated identity they'll need to provide:

- The issuer (Required), which is the `iss` for the external identity service
- The subject, which defaults to the client ID for the OAuth client (but can be set to a different value, e.g. could be a SPIFFE ID when using SPIFFE/SPIRE)
- The audience for the token, which by default is the URL of the Pocket ID server
- The URL of the JWKS, which by default is `<iss>/.well-known/jwks.json`

Then, OAuth2 apps will be able to present as `client_assertion` JWTs that are signed by the issuer above, and which have the subject and audience configured as per above.

## How you can test this

This PR only includes the backend changes, and not what needs to be added to the UI. However, it can still be tested with some work.

Credentials:

- Issuer: `https://alessandrosegala.blob.core.windows.net/`
- JWKS URI: `https://alessandrosegala.blob.core.windows.net/$root/jwks.json`

To test it:

1. Create a new OAuth application in Pocket ID. In the database, edit the row and set `credentials` to this JSON string: `{"federatedIdentities":[{"issuer":"https://alessandrosegala.blob.core.windows.net/","subject":"123456abcdef","audience":"api://PocketIDTokenExchange","jwks":"https://alessandrosegala.blob.core.windows.net/$root/jwks.json"}]}`
2. Start the OAuth authorization flow (I've used [oauth.tools](https://oauth.tools/) but any other way works) and obtain an access code
3. When you get the access code, you can use one of the tokens below and curl to exchange it for an auth token:

```sh
CODE=your-access-code
CLIENT_ID=your-client-id
REDIRECT_URI=your-redirect-uri
curl -Ss -X POST \
http://localhost:1411/api/oidc/token \
-H 'Content-Type: application/x-www-form-urlencoded' \
-d 'grant_type=authorization_code&redirect_uri='${REDIRECT_URI}'code&code='${CODE}'&client_id='${CLIENT_ID}'&client_assertion_type=urn%3Aietf%3Aparams%3Aoauth%3Aclient-assertion-type%3Ajwt-bearer&client_assertion='${CLIENT_ASSERTION}
```

Use one of these values for `CLIENT_ASSERTION`:

```sh
# Valid token
CLIENT_ASSERTION=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkZaU1Z3WkU3aUxTek1jcnhRb2lOOWRjcEphS2VNUzdTQnE5eFN6SVhLLTQifQ.eyJzdWIiOiIxMjM0NTZhYmNkZWYiLCJpc3MiOiJodHRwczovL2FsZXNzYW5kcm9zZWdhbGEuYmxvYi5jb3JlLndpbmRvd3MubmV0LyIsImlhdCI6MTc0ODE0NzI1NiwiYXVkIjoiYXBpOi8vUG9ja2V0SURUb2tlbkV4Y2hhbmdlIiwiZXhwIjoxNzQ5MjQ3MjU2fQ.ZBTPYFxe4X2qHic4t-P09kfgfCf0x9FfMSlH2zc_2dadAQ9rZDKGTekAuSZ9sL7FofrxzElh_tUKAYV0ScpKcnQJY-t7gB1bT6jdevndmTUmBOa8NcFfSSAxMyBxzuD-OAzhQG78JMk_Gf4LJVLwyXL7Ngz_feM7o3zxI0TRPBlurfE500PPwpfhZB90ggEiMtxBLHFws0rbUat98FBOcsKdjqkp9p0NsLuzDtxM2W1nOJ_LXk9rOKPb5naTKHzgajWOV9yg1PSx_vxqkFVH9BV9byimgLc1BKXRxKfm79YrRo3CWmJ-FvcIkpvAkgOdqi6KiZ2KpZSPJjH8sCYKXw

# Fails - invalid subject
CLIENT_ASSERTION=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkZaU1Z3WkU3aUxTek1jcnhRb2lOOWRjcEphS2VNUzdTQnE5eFN6SVhLLTQifQ.eyJzdWIiOiJCQUQiLCJpc3MiOiJodHRwczovL2FsZXNzYW5kcm9zZWdhbGEuYmxvYi5jb3JlLndpbmRvd3MubmV0LyIsImlhdCI6MTc0ODE0NzI1NiwiYXVkIjoiYXBpOi8vUG9ja2V0SURUb2tlbkV4Y2hhbmdlIiwiZXhwIjoxNzQ5MjQ3MjU2fQ.RtpSGnJ0w45Epr8p6nRrA0dLe-V69HU3RwRfEvjffApYkM3ln-rkyeVWjzJIZ-qpe4yQG5UtkkOk4mMrb-L0_WRDP25p_bdCu42UTa_B3OpAATkTfWlGI2HWBr4w3ExUkLFK7UXfs0l5A11huEbbSQwqsBuk3rTkeWwlE8pq8RAjZbVTkpezMHBt-L9L-eeDmiX71ZxYFuUnQ557geyhh0guO17OcVaTbDiSKNkkQZQHOZhs2Yy8aVjeJgBoUDAJwlIdRI4O_JhqdlZgk88HHFl47akNom4sLfyFmo8ez3BD89-IINuQnY27LWT9sHiTNi0o3-6bA_bilNgpwHsFXw

# Fails - invalid audience
CLIENT_ASSERTION=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkZaU1Z3WkU3aUxTek1jcnhRb2lOOWRjcEphS2VNUzdTQnE5eFN6SVhLLTQifQ.eyJzdWIiOiIxMjM0NTZhYmNkZWYiLCJpc3MiOiJodHRwczovL2FsZXNzYW5kcm9zZWdhbGEuYmxvYi5jb3JlLndpbmRvd3MubmV0LyIsImlhdCI6MTc0ODE0NzI1NiwiYXVkIjoiYXBpOi8vQkFEIiwiZXhwIjoxNzQ5MjQ3MjU2fQ.S3gAzBKH9JPF0bbr2qgI6WI29YCXtipsVTNhIKMlgwzPHIILR3maDM2e8_jQyuayZhpHeGXa7eTrNyVhKBksmfhkM6Js_Vmv-L1Q6Oc0FkTjMJvTWC_H7G6a1O5eo3jlWICVfSk6sHOG8mosCBNA0B3xZVrs5H5MPm5Of9AEIrtAo6VbtRo6fkwCmDxm6fzeTtws3g7WlWrHjQZcGrKbXo44Hzsp6WPwmCF8Xp9GgkRbympshizJSewRQsBnoWE0xQ7KFmTLhvmBTrwiQ-ifWrre1VoqCOjuoGOoMEnxPlVYcHBurdeK5VHsxgm9BwgmxThjrt2KVc2SQ-DvWLkEpw

# Fails - signed by an invalid key
CLIENT_ASSERTION=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkZaU1Z3WkU3aUxTek1jcnhRb2lOOWRjcEphS2VNUzdTQnE5eFN6SVhLLTQifQ.eyJzdWIiOiIxMjM0NTZhYmNkZWYiLCJpc3MiOiJodHRwczovL2FsZXNzYW5kcm9zZWdhbGEuYmxvYi5jb3JlLndpbmRvd3MubmV0LyIsImlhdCI6MTc0ODE0NzI1NiwiYXVkIjoiYXBpOi8vUG9ja2V0SURUb2tlbkV4Y2hhbmdlIiwiZXhwIjoxNzQ5MjQ3MjU2fQ.EnrllAAKlOSv_3odLItIQHx28HwbV84RhHY5Hfh95eb8gzvUWYHR_vAuIj91hY5oyEtOn6qM3WQRRdWAWPoOcuxbVSW09TG6MXA6DZo7Pd43_IkWAovH5qen7uIb0wiv_3Vf3dls_YRzuHd5Mo2wdIKVmYJSrg22Qd5x6clcDXhwrfO_5B1H8WLlV-KaiEMBjN-7YFh59Eb4wbHrNaXrFyYWjuD-PZeG1NH9aSFBNeOq0KArjYIvn96mjhlKiE_DT15CK1zPjndikbK_SwAovOgGCV4Jnazm0_XF9AOCuPNEJlKwb0MhhcNcsr3Aelo9_3zxLKbe8DbEgI0I4ZjWYw
```

## What needs to be done for the UI

I am thinking the UI should have a table with the ability to add multiple federated identities. For each one, users will have to provide the issuer. Other values (audience, subject, and JWKS URL) can be empty and can be set automatically by Pocket ID.


kmendell:
@ItalyPaleAle Is this what you wanted help with the frontend on? I think we should contain it all in one PR. Ill see if i can make some sort of UI for this, i need to make sure i understand what is needed though.

ItalyPaleAle:
> @ItalyPaleAle Is this what you wanted help with the frontend on? I think we should contain it all in one PR. Ill see if i can make some sort of UI for this, i need to make sure i understand what is needed though.

Yes it is, if you’re willing to help!

Feel feee to ping me on discord with any question

ItalyPaleAle:
> @ItalyPaleAle Is this what you wanted help with the frontend on? I think we should contain it all in one PR. Ill see if i can make some sort of UI for this, i need to make sure i understand what is needed though.

Yes it is, if you’re willing to help!

Feel feee to ping me on discord with any question

As a maintainer of the repo you should have write access to my branch so you can push to that if you want

kmendell:
There seems to be some issues with when create a client with credentials. I will look into that

kmendell:
@ItalyPaleAle How many of these would a user setup would it be more than like 5? if so i wanna redo the UI, if its not then i think the UI should be fine.

ItalyPaleAle:
> @ItalyPaleAle How many of these would a user setup would it be more than like 5? if so i wanna redo the UI, if its not then i think the UI should be fine.

Hard to know for sure but i don’t think most people will have more than 1 or 2

stonith404:
@ItalyPaleAle Do you think we could make `Credentials` on the `OidcClient` not nullable? I think it would make more sense that the list is just empty instead of null/nil when there are no credentials?

frontend/messages/en-US.json
```diff
@@ -348,6 +348,12 @@
 	"the_device_has_been_authorized": "The device has been authorized.",
 	"enter_code_displayed_in_previous_step": "Enter the code that was displayed in the previous step.",
 	"authorize": "Authorize",
+	"federated_identities": "Federated Identities",
+	"federated_identities_description": "Add external identity providers (Azure, AWS, Kubernetes, etc.) that can authenticate this client using JWT tokens instead of client secrets for enhanced security.",
```
ItalyPaleAle: Suggestion:

```suggestion
	"federated_identities_description": "Using federated identities, you can authenticate OAuth2 clients (apps) using JWT tokens issued by third-party authorities, including cloud platforms, Kubernetes, SPIRE, etc. When available, this is a potentially safer alternative to using client secrets.",
```

I think I can also add a docs page and link to that?

ItalyPaleAle:
> @ItalyPaleAle Do you think we could make `Credentials` on the `OidcClient` not nullable? I think it would make more sense that the list is just empty instead of null/nil when there are no credentials?

Are you talking about the data in the database, or the DTO object, or both?

frontend/messages/en-US.json
stonith404: Looks good, maybe we can omit the examples to make the description a bit shorter? Adding a docs section would probably make sense too.

stonith404:
> Are you talking about the data in the database, or the DTO object, or both?

I would do it everywhere null/nil just feels a bit redundant as it indicates the same state as an empty list. Additionally it would type safety a bit easier on the forms in the frontend.

frontend/messages/en-US.json
ItalyPaleAle:
```suggestion
	"federated_identities_description": "Using federated identities, you can authenticate OAuth2 clients (apps) using JWT tokens issued by third-party authorities. When available, this is a potentially safer alternative to using client secrets.",
```

kmendell: I think we should shorten it to something like this: 

`Using federated identities, you can authenticate OAuth2 clients using JWT tokens issued by third-party authorities. Learn More.` 

(learn more would be a link) That cleans up and doesn't add un-needed fluff/filler words.

ItalyPaleAle: I like that, I’d just maybe add “OAuth2 clients (applications)” because the terminology is confusing (unless you’re involved with the RFCs!)

stonith404: Or we just call it 'OIDC Clients' like we do everywhere else in the admin UI? I mean yeah, 'OAuth2 clients' would be more correct, but for the sake of simplicity I think this would make more sense.

stonith404:
@ItalyPaleAle Is there a way we can test the JWT bearer assertion in our end-to-end tests? Is it possible to use a dummy token which is valid for a very long time along side with a dummy federated identity or would we have to host our own  federated identity for e2e tests?

frontend/messages/en-US.json
kmendell: I agree with Elias, We should keep it consistant throughout the UI.

kmendell: Added here [`878a42f` (#566)](https://github.com/pocket-id/pocket-id/pull/566/commits/878a42f26b432240d28b2dd527f26b59caa3a2b5)

ItalyPaleAle:
> @ItalyPaleAle Is there a way we can test the JWT bearer assertion in our end-to-end tests? Is it possible to use a dummy token which is valid for a very long time along side with a dummy federated identity or would we have to host our own federated identity for e2e tests?

Done!

frontend/messages/en-US.json
ItalyPaleAle: Thanks @kmendell 

Once the docs are out we can upate to add a link?

Issue #538, opened 2025-05-17
🚀 Feature: Auto detect and populate callback URL


mitchplze:
### Feature description

Request is to provide the option to automatically detect and populate the `Callback URL` when the field is left blank during initial OIDC client setup (or via checkmark/other mechanism).

This concept is borrowed from [authentik](https://docs.goauthentik.io/docs/add-secure-apps/providers/oauth2/#additional-configuration-options-with-redirect-uris):

> When you create a new OAuth 2.0 provider and app in authentik and you leave the Redirect URI field empty, then the first time a user opens that app, authentik uses that URL as the saved redirect URL.

For this field: 

<img width="467" alt="Image" src="https://github.com/user-attachments/assets/c6792f1b-5476-4c79-a740-9391dea752e9" />

### Pitch

- Depending on the client application and their documentation, it is often difficult to identify the exact callback URL for optimal security, while still allowing SSO with Pocket ID to work.

- Sometimes callback URLs are randomly generated by the application, and not immediately apparent until the first authorization attempt.

- With this feature, the first app to 'try' the OIDC connection after initially setting it up in Pocket - will win - and populate the client with the proper value.

- Auto detecting this value is inherently more secure than using wildcards (where possible).

- Potential challenge: the callback URL does not currently appear to be logged / visible in the UI.

ItalyPaleAle:
I like the suggestion of "TOFU" (Trust On First Use), but if the main problem is that the callback URL is sometimes hard to detect, how about we just show an error message saying `callback url https://example.com is invalid`?

kmendell:
> I like the suggestion of "TOFU" (Trust On First Use), but if the main problem is that the callback URL is sometimes hard to detect, how about we just show an error message saying `callback url https://example.com is invalid`?

An error is already thrown and showed if the callback url is not right, im confused on what you're suggesting by this.

mitchplze:
I'm not sure what causes the error y'all are talking about to be thrown in any case. 

The box seems to accept anything: 

<img width="968" alt="Image" src="https://github.com/user-attachments/assets/6ceaed0a-9fb3-442f-9046-46e258f62fb3" />

kmanwar89:
Threw my upvote into the ring here -- what I noticed is you can't create the app until the callback is provided, but some apps don't provide you the callback until the IdP is configured, so it becomes chicken/egg.

[...]

PR #583, opened 2025-05-28
feat: auto detect callback url

kmendell:
fixes: https://github.com/pocket-id/pocket-id/issues/538

### DIFF
--- a/backend/internal/service/oidc_service.go
+++ b/backend/internal/service/oidc_service.go
@@ -73,7 +73,6 @@ func (s *OidcService) Authorize(ctx context.Context, input dto.AuthorizeOidcClie
 	}
 
 	// Get the callback URL of the client. Return an error if the provided callback URL is not allowed
-	// This now handles TOFU logic automatically when no URLs are configured
 	callbackURL, err := s.getCallbackURL(client.CallbackURLs, input.CallbackURL, input.ClientID, tx, ctx)
 	if err != nil {
 		return "", "", err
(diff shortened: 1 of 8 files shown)


frontend/src/routes/settings/admin/oidc-clients/oidc-client-form.svelte
```diff
@@ -30,15 +30,15 @@
...
+		callbackURLs: existingClient?.callbackURLs || [], // Changed from [''] to []
 		logoutCallbackURLs: existingClient?.logoutCallbackURLs || [],
 		isPublic: existingClient?.isPublic || false,
 		pkceEnabled: existingClient?.pkceEnabled || false
 	};
 
 	const formSchema = z.object({
 		name: z.string().min(2).max(50),
-		callbackURLs: z.array(z.string().nonempty()).nonempty(),
+		callbackURLs: z.array(z.string()).optional().default([]), // Made optional and allow empty strings
```
stonith404, line 41:
```suggestion
		callbackURLs: z.array(z.string().nonempty()).default([])
```
We still have to make sure that the callback URL isn't empty, if the user add one.

backend/internal/service/oidc_service.go
```diff
@@ -947,7 +947,7 @@ func (s *OidcService) ValidateEndSession(ctx context.Context, input dto.OidcLogo
 		return "", &common.OidcNoCallbackURLError{}
 	}
 
-	callbackURL, err := s.getCallbackURL(userAuthorizedOIDCClient.Client.LogoutCallbackURLs, input.PostLogoutRedirectUri)
+	callbackURL, err := s.getCallbackURL(userAuthorizedOIDCClient.Client.LogoutCallbackURLs, input.PostLogoutRedirectUri, userAuthorizedOIDCClient.Client.ID, s.db, ctx)
```
ItalyPaleAle, line 950: This doesn't use a transaction however, just queries in sequence. If multiple attempts are made at the same time, you can end up with more than 1 authorized URL

PR #585, opened 2025-05-29
fix: run jobs at interval instead of specific time

ItalyPaleAle:
While this is helpful for all jobs, it's particularly useful for the analytics job, to avoid all instances to send the ping at the exact same time, and to make sure it's always about 24 hours away.

Additionally, runs all database cleanup jobs in a single transaction, to avoid contention when all cleanup jobs are started at the same time (since they'd be competing to lock the database, especially with SQLite).

### DIFF
--- a/backend/internal/job/db_cleanup_job.go
+++ b/backend/internal/job/db_cleanup_job.go
@@ -2,6 +2,7 @@ package job
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"time"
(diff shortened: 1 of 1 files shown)


backend/internal/job/api_key_expiry_job.go
```diff
@@ -29,16 +33,16 @@ func (j *ApiKeyEmailJobs) checkAndNotifyExpiringApiKeys(ctx context.Context) err
...
 
 	for _, key := range apiKeys {
 		if key.User.Email == "" {
 			continue
 		}
-		if err := j.apiKeyService.SendApiKeyExpiringSoonEmail(ctx, key); err != nil {
-			log.Printf("Failed to send email for key %s: %v", key.ID, err)
+		err = j.apiKeyService.SendApiKeyExpiringSoonEmail(ctx, key)
+		if err != nil {
+			slog.ErrorContext(ctx, "Failed to send expiring API key notification email", slog.String("key", key.ID), slog.Any("error", err))
```
stonith404, line 45: Should we use `slog` everywhere in the application in the future?

backend/internal/job/db_cleanup_job.go
```diff
@@ -14,55 +16,66 @@ import (
...
 	jobs := &DbCleanupJobs{db: db}
 
-	return errors.Join(
-		s.registerJob(ctx, "ClearWebauthnSessions", "0 3 * * *", jobs.clearWebauthnSessions, false),
-		s.registerJob(ctx, "ClearOneTimeAccessTokens", "0 3 * * *", jobs.clearOneTimeAccessTokens, false),
-		s.registerJob(ctx, "ClearOidcAuthorizationCodes", "0 3 * * *", jobs.clearOidcAuthorizationCodes, false),
-		s.registerJob(ctx, "ClearOidcRefreshTokens", "0 3 * * *", jobs.clearOidcRefreshTokens, false),
-		s.registerJob(ctx, "ClearAuditLogs", "0 3 * * *", jobs.clearAuditLogs, false),
-	)
+	// Run every 24 hours and now
```
stonith404, line 19: I think it would make more sense to keep the old jobs but use [DurationRandomJob](https://pkg.go.dev/github.com/go-co-op/gocron/v2#DurationRandomJob) because if clearing fails for one record, all the clearing for the other records will be rolled back, which doesn't make sense to me.

ItalyPaleAle: Fair. That's what I tried earlier: https://github.com/pocket-id/pocket-id/pull/585/commits/e38fd916944d033836adead0bfa02a5915b1376b I can revert.

...although, if one job fails, then we do have some big problems

ItalyPaleAle: Done

stonith404 (approved):
Great, thanks!

PR #615, opened 2025-06-05
feat: add unix socket support

CnTeng:
Allow the server to listen UNIX socket.

### DIFF
--- a/backend/internal/bootstrap/router_bootstrap.go
+++ b/backend/internal/bootstrap/router_bootstrap.go
@@ -101,21 +101,27 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
 
 	// Set up the server
 	srv := &http.Server{
-		Addr:              net.JoinHostPort(common.EnvConfig.Host, common.EnvConfig.Port),
 		MaxHeaderBytes:    1 << 20,
 		ReadHeaderTimeout: 10 * time.Second,
 		Handler:           r,
 	}
 
 	// Set up the listener
-	listener, err := net.Listen("tcp", srv.Addr)
+	network := "tcp"
+	addr := net.JoinHostPort(common.EnvConfig.Host, common.EnvConfig.Port)
+	if common.EnvConfig.UnixSocket != "" {
+		network = "unix"
+		addr = common.EnvConfig.UnixSocket
+	}
+
+	listener, err := net.Listen(network, addr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


stonith404:
Thank you for your contribution. Could you please elaborate on the advantages this brings and the potential use cases?

CnTeng:
> Thank you for your contribution. Could you please elaborate on the advantages this brings and the potential use cases?

- Both Caddy and Nginx support reverse proxying using Unix sockets directly:

```caddyfile
your.pocket-id.xxx {
  reverse_proxy unix//run/pocket-id.sock
}
```

- You can control the socket’s owner and group to restrict access to Caddy only, which enhances security.
- Several applications also support this feature:
  - [`miniflux`](https://miniflux.app/docs/configuration.html#listen-addr) supports listening on a Unix socket via `LISTEN_ADDR` and also supports systemd socket activation.
  - [`ntfy-sh`](https://docs.ntfy.sh/config/?h=unix#config-options) supports the `listen-unix` option to bind to a Unix socket.
  - [`wabdav`](https://github.com/hacdias/webdav/blob/04cee682fb42c7f684eba106c14aff3ba2fa20c0/cmd/root.go#L113) supports Unix sockets and systemd socket activation.
 
Maybe we can add `systemd socket` support in next pr.

kmendell:
What's the benefit of this over normal HTTP though is my question? How popular is this? I've never heard of it till now. 

I guess maybe im not following the advantages of it honestly.

CnTeng:
> What's the benefit of this over normal HTTP though is my question? How popular is this? I've never heard of it till now.
> 
> I guess maybe im not following the advantages of it honestly.

I have given some examples. And, It's just an other way for http service. You can control the owner and no need to determine which port to use.

CnTeng:
And [`authelia`](https://www.authelia.com/configuration/miscellaneous/server/#address) support it too.

kmendell:
Oh okay gotcha, I can wrap my head around that :D.

stonith404:
@CnTeng Thanks for the clarification. Wouldn't `UNIX_SOCKET` make more sense as the name of the environment variable?

ItalyPaleAle:
It's not super common but other apps do support HTTP over UDS (Unix Domain Sockets). More here: https://serverfault.com/questions/124517/what-is-the-difference-between-unix-sockets-and-tcp-ip-sockets

While it's up to the maintainers to decide if they want this or not, the code change itself LGTM.

stonith404 (approved):

PR #623, opened 2025-06-06
refactor: upgrade to Zod v4

stonith404:
Zod 4 brings the advantage of translated validation messages. The translated validation messages are included in the package and will be lazily loaded based on the user's locale.


stonith404:
@kmendell Great, thanks for reviewing.

PR #636, opened 2025-06-08
chore: Add docs link and rename to Federated Client Credentials

ItalyPaleAle:
Follow-up from #566

Adds link to docs (see https://github.com/pocket-id/website/pull/104) and renames the feature in the UI to "federated client credentials", to make it specific this is about client credentials and not user identity


stonith404 (approved):
Thanks!

PR #640, opened 2025-06-09
feat: allow introspection endpoint to use Federated Client Credentials

ItalyPaleAle:
Follow-up from #566 to complete the work started there

- Add support for using federated client credentials for the introspection endpoint, to validate auth and refresh tokens. Calls to the endpoint use `Authorization: Bearer <jwt>` for authorization.
- As part of that change, and for other reasons discussed on Discord, the refresh token's format has changed, and it's now a JWT containing the actual refresh code (stored in the DB, hashed), plus the ID of the client that uses it and the ID of the user it belongs to
   - This is required because otherwise there's no way to know the client ID when introspecting a refresh token using federated client credentials
   - It also allows more careful database lookups
   - The RFC doesn't mandate anything about the format of refresh tokens, which are opaque strings for clients, so this remains fully-compliant

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -198,7 +198,7 @@ func (oc *OidcController) userInfoHandler(c *gin.Context) {
 		return
 	}
 
-	token, err := oc.jwtService.VerifyOauthAccessToken(authToken)
+	token, err := oc.jwtService.VerifyOAuthAccessToken(authToken)
 	if err != nil {
 		_ = c.Error(err)
 		return
@@ -306,9 +306,10 @@ func (oc *OidcController) introspectTokenHandler(c *gin.Context) {
 	// find valid tokens) while still allowing it to be used by an application that is
 	// supposed to interact with our IdP (since that needs to have a client_id
 	// and client_secret anyway).
-	clientID, clientSecret, _ := c.Request.BasicAuth()
+	var creds service.ClientAuthCredentials
+	creds.ClientID, creds.ClientSecret, _ = c.Request.BasicAuth()
 
-	response, err := oc.oidcService.IntrospectToken(c.Request.Context(), clientID, clientSecret, input.Token)
+	response, err := oc.oidcService.IntrospectToken(c.Request.Context(), creds, input.Token)
 	if err != nil {
 		_ = c.Error(err)
 		return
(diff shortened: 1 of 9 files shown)


backend/internal/controller/oidc_controller.go
```diff
@@ -306,9 +306,22 @@ func (oc *OidcController) introspectTokenHandler(c *gin.Context) {
...
 	// and client_secret anyway).
-	clientID, clientSecret, _ := c.Request.BasicAuth()
+	var (
+		creds service.ClientAuthCredentials
+		ok    bool
+	)
+	creds.ClientID, creds.ClientSecret, ok = c.Request.BasicAuth()
+	if !ok {
+		// If there's no basic auth, check if we have a bearer token
+		const prefix = "bearer "
```
stonith404: I think it would make more sense if we extract this extraction logic into a util function.

backend/internal/service/oidc_service.go
```diff
@@ -406,22 +409,39 @@ func (s *OidcService) createTokenFromRefreshToken(ctx context.Context, input dto
...
 	}()
 
-	_, err := s.verifyClientCredentialsInternal(ctx, tx, input)
+	client, err := s.verifyClientCredentialsInternal(ctx, tx, clientAuthCredentialsFromCreateTokensDto(&input))
 	if err != nil {
 		return CreatedTokens{}, err
 	}
 
+	// The ID of the client that made the call must match the client ID in the token
+	if client.ID != clientID {
```
stonith404, line 429: Couldn't we remove the client ID and user ID from the `oidc_refresh_token` if the client ID is already stored in the token? Storing the expiration probably still makes sense to speed up the clean up.

ItalyPaleAle: Not sure I understand the question, but:

- The user ID can be removed, but I thought it wouldn't hurt to have it in there so we can use it in the WHERE clause when querying the DB for binding the refresh token even more.
- For the client ID:
   - When using basic auth, which is the option for using client secrets, the Authorization header contains `Authorization: Basic <base64(client-id + ':' + ' client-secret>` so the client ID is included in the header
   - When using federated credentials, the Authorization header contains the JWT as a bearer token (`Authorization: Bearer <jwt>`) and there's no room fo the client ID. Because the JWT is issued by a third-party IdP, there's no guarantee that it will contain the client ID either. So we need to fetch the client ID from somewhere else, and using the access token or refresh token themselves seemed the most straightforward option

stonith404: Fair enough, that makes sense, thanks.

stonith404 (approved):

PR #643, opened 2025-06-09
feat: ui accent colors

kmendell:


kmendell:
@stonith404 We can probably remove the CheckboxWithLabel component now, i left it for you to decide but everything that used it is now using the Switch version.

PR #647, opened 2025-06-10
feat: auto-focus on the login buttons

ItalyPaleAle:
On the /login and /authorize views, use auto-focus to make the "Login" and "Sign In" buttons (respectively) automatically focused

This helps when doing keyboard navigation. For example, imagine refreshing the page with CMD/CTRL+R, being sent to Pocket ID to re-authorize an app after a session expired. Your fingers were on the keyboard, but now you need to use your mouse to select "Sign In" or "Login" (or press tab multiple times and pay attention to where the focus is), and then go back to the fingerprint sensor to make the passkey work.

By setting auto-focus, you can just hit enter and go straight to the passkey.

### DIFF
--- a/frontend/src/lib/components/ui/button/button.svelte
+++ b/frontend/src/lib/components/ui/button/button.svelte
@@ -37,11 +37,13 @@
 			variant?: ButtonVariant;
 			size?: ButtonSize;
 			isLoading?: boolean;
+			autofocus?: boolean;
 		};
 </script>
 
 <script lang="ts">
 	import LoaderCircle from '@lucide/svelte/icons/loader-circle';
+	import { onMount } from 'svelte';
 
 	let {
 		class: className,
@@ -52,9 +54,18 @@
 		type = 'button',
 		disabled,
 		isLoading = false,
+		autofocus = false,
 		children,
 		...restProps
 	}: ButtonProps = $props();
+
+	onMount(async () => {
+		// Using autofocus can be bad for a11y, but in the case of Pocket ID is only used responsibly in places where there are not many choices, and on buttons only where there's descriptive text
+		if (autofocus) {
+			// Use setTimeout to make sure the element is showing
+			setTimeout(() => ref?.focus(), 100)
+		}
+	});
 </script>
 
 {#if href}
--- a/frontend/src/routes/authorize/+page.svelte
+++ b/frontend/src/routes/authorize/+page.svelte
@@ -138,14 +138,20 @@
 				</Card.Root>
 			</div>
 		{/if}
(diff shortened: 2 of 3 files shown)


stonith404:
I like this improvement. Couldn't we just use the native [`autofocus`](https://www.w3schools.com/tags/att_button_autofocus.asp) attribute on the button?

ItalyPaleAle:
> I like this improvement. Couldn't we just use the native [`autofocus`](https://www.w3schools.com/tags/att_button_autofocus.asp) attribute on the button?

I tried that but didn't work, at least not for the sign in page. It seems that the attribute only works if the element exists at page load: https://stackoverflow.com/a/27314017

kmendell:
Yeah i dont see the autofocus element passed to the https://bits-ui.com/docs/components/button component either, though the merge props should work with it technially.

kmendell:
Though tabindex exists, im not sure if it does the same thing or not though https://github.com/huntabyte/bits-ui/blob/2dd8e34fdc24b8d39ae50fd6081220f0fa81f4f1/packages/bits-ui/src/lib/bits/button/components/button.svelte#L22

ItalyPaleAle:
> Yeah i dont see the autofocus element passed to the https://bits-ui.com/docs/components/button component either, though the merge props should work with it technially.

I did manage to get the prop passed. The problem was that it was just not working, at least on the signin page: it was not auto-focused. Seems to be because the element doesn't exist when the page's DOM loads initially and it's added later.

As for tabindex, that controls the order when elements are focused when you press tab. But per MDN it should not be used with values others than 0 and -1: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex

kmendell:
I swear i saw a change from huntabyte about autofocus somewhere but maybe that was not related to the button component, i was trying to find that but i cant seem too. Either way i think the proposed changes may be the simpilest way to do this, just by first glance.

ItalyPaleAle:
> I swear i saw a change from huntabyte about autofocus somewhere but maybe that was not related to the button component, i was trying to find that but i cant seem too. Either way i think the proposed changes may be the simpilest way to do this, just by first glance.

I don't _super love_ this solution (especially the need for the setTimeout for synchronization - I considered that acceptable just because this is used for focusing and not something major) so if the author of the component has found a better one that's great!

stonith404:
Okay that's interesting thanks for clarifying. 

I searched for issues, PRs and commits in the Shadcn Svelte repo but didn't find a change related to auto focus. If you find it @kmendell we could adapt it in the future.

stonith404 (approved):

Issue #634, opened 2025-06-07
🚀 Feature: Approximate Location Recognizes Local IPv6 as LAN, Internal Network


LucasJanin:
### Feature description

Currently, Pocket-ID recognizes only local IPv4 addresses as part of a "LAN, internal network" in the Approximate Location. Local IPv6 addresses are always treated as external traffic.

Proposal:
Introduce an environment variable ⁠LOCAL_IPV6_RANGES that contains a list of IPv6 ranges used in the local network. Pocket-ID will utilize this list to determine if an IP address is part of the "LAN, internal network." If an IP address is not included in this list, Pocket-ID will use MAXMIND to find the approximate location.

Example Variable:
⁠LOCAL_IPV6_RANGES=2001:0db8:abcd:000::/56,2001:0db8:abcd:001::/56

### Pitch

This option will quickly determine whether the traffic is local. Distinguishing between local and external visitors is crucial, as external traffic can pose a greater threat than trusted local network traffic. By accurately identifying local traffic, we can enhance security.

dougmeredith:
I don't have a strong opinion on this (or need for it), but I do think there is another way it could be viewed:

Location is meant to be just that; a location. Since there is no way to geo-locate IPv4 private addresses, LAN/Internal is really just how things get displayed when we don't have the location. IPv6 global addresses don't suffer from this shortcoming, and since the location can (theoretically) be determined, the correct location should be recorded.

Perhaps a better concept would be that of trusted IP ranges, which is distinct from location.

kmendell:
@LucasJanin Can you try this image `ghcr.io/pocket-id/pocket-id:ipv6-ranges` the env variable `LOCAL_IPV6_RANGES` can be set for local IPv6 ranges and those should be makred as LAN, Internal Network in the audit log.

LucasJanin:
Hi @kmendell 

It's working perfectly! 

<img width="1455" alt="Image" src="https://github.com/user-attachments/assets/cf5992c1-d6b7-4542-9cd8-e924a2011880" />

However, I can't fully test it because I performed a fresh Docker installation (I'm using the bare metal version). I'm looking forward to testing it on my main installation with the compiled version.

Thanks for adding this request.

PR #657, opened 2025-06-15
feat: configurable local ipv6 ranges for audit log

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/634


backend/internal/service/geolite_service.go
```diff
@@ -22,9 +23,11 @@ import (
...
 
 type GeoLiteService struct {
-	httpClient     *http.Client
-	disableUpdater bool
-	mutex          sync.RWMutex
+	httpClient      *http.Client
+	disableUpdater  bool
+	mutex           sync.RWMutex
+	localIPv6Ranges []*net.IPNet
+	ipv6RangeMutex  sync.RWMutex
```
stonith404: Is the mutex really necessary since write operations are only done during initialization?

kmendell: [`e57db63` (#657)](https://github.com/pocket-id/pocket-id/pull/657/commits/e57db63f37a33dc43c9c998f5085811226770ecc)

@stonith404 Removed them here :)

stonith404 (approved):

backend/internal/service/geolite_service.go
stonith404: Awesome, thanks :)

PR #661, opened 2025-06-18
feat: allow setting unix socket mode

CnTeng:
This is a supplement to #615.

It is difficult to set socket permissions externally, so add new option to set socket permissions internally.

### DIFF
--- a/backend/internal/common/env_config.go
+++ b/backend/internal/common/env_config.go
@@ -33,6 +33,7 @@ type EnvConfigSchema struct {
 	Port               string     `env:"PORT"`
 	Host               string     `env:"HOST"`
 	UnixSocket         string     `env:"UNIX_SOCKET"`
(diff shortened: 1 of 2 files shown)


CnTeng:
All right, I will add it later.

backend/internal/common/env_config.go
```diff
@@ -53,6 +54,7 @@ var EnvConfig = &EnvConfigSchema{
 	Port:               "1411",
 	Host:               "0.0.0.0",
 	UnixSocket:         "",
+	UnixSocketMode:     "0666",
```
ItalyPaleAle: Should default value be empty? So we avoid running the chmod unless the user explicitly wants it?

CnTeng: Well, use system default mode is better.

Issue #635, opened 2025-06-07
🚀 Feature: Global Audit Log adds a local traffic filter


LucasJanin:
### Feature description

Add a "Location" filter on the Global Audit Log page, positioned between Events and Clients. This filter will offer two options: Local and External. The Local option will display only visitors from the "LAN, Internal Network," while the External option will show all other visitors.

### Pitch

This option will allow for the quick identification of whether the traffic is local. Clearly distinguishing between local and external visitors is crucial, as external traffic can pose a greater potential threat than trusted local network traffic. By accurately identifying local traffic, we can enhance security and improve the overall integrity of the audit log.

PR #662, opened 2025-06-18
feat: location filter for global audit log

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/635

### DIFF
--- a/backend/internal/service/audit_log_service.go
+++ b/backend/internal/service/audit_log_service.go
@@ -153,9 +153,9 @@ func (s *AuditLogService) ListAllAuditLogs(ctx context.Context, sortedPagination
 	if filters.Location != "" {
 		switch filters.Location {
 		case "external":
-			query = query.Where("country != 'Internal Network' AND country IS NOT NULL AND country != ''")
+			query = query.Where("country != 'Internal Network'")
 		case "internal":
-			query = query.Where("country = 'Internal Network' OR country IS NULL OR country = ''")
+			query = query.Where("country = 'Internal Network'")
 		}
 	}
 
--- a/backend/internal/service/audit_log_service.go
+++ b/backend/internal/service/audit_log_service.go
@@ -151,9 +151,10 @@ func (s *AuditLogService) ListAllAuditLogs(ctx context.Context, sortedPagination
 		}
 	}
 	if filters.Location != "" {
-		if filters.Location == "external" {
+		switch filters.Location {
+		case "external":
 			query = query.Where("country != 'Internal Network' AND country IS NOT NULL AND country != ''")
-		} else if filters.Location == "internal" {
+		case "internal":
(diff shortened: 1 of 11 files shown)


backend/internal/service/audit_log_service.go
```diff
@@ -150,6 +150,14 @@ func (s *AuditLogService) ListAllAuditLogs(ctx context.Context, sortedPagination
 			return nil, utils.PaginationResponse{}, fmt.Errorf("unsupported database dialect: %s", dialect)
 		}
 	}
+	if filters.Location != "" {
+		switch filters.Location {
+		case "external":
+			query = query.Where("country != 'Internal Network' AND country IS NOT NULL AND country != ''")
```
stonith404: Do we really have to check if the country isn't null? Can't we just check if the country isn't "Internal Network"?

frontend/src/routes/settings/audit-log/global/+page.svelte
```diff
@@ -82,6 +88,19 @@
 					bind:value={filters.event}
 				/>
 			</div>
+			<div>
+				<SearchableSelect
```
stonith404, line 92: I think a "normal" select without search would fit better here because we don't have much options.

backend/internal/service/audit_log_service.go
ItalyPaleAle: Can you please create an index on `country`? Right now, it's not indexed, so this query would always perform a full table scan

kmendell: Yeah your right i was more over covering my bases with it, ill get it changed.

kmendell: Good catch, Ill get it added.

frontend/src/routes/settings/audit-log/global/+page.svelte
kmendell: [`841de5b` (#662)](https://github.com/pocket-id/pocket-id/pull/662/commits/841de5b81aa7aaa7694cf9cfc80d67dd185e332e)

kmendell: @stonith404  See if you like that approach this way the styling stays consistent.

backend/internal/service/audit_log_service.go
kmendell: https://github.com/pocket-id/pocket-id/pull/662/commits/5ec2ba06c269097266ad1805642388f67bdc592c

kmendell: [`1afd9e0` (#662)](https://github.com/pocket-id/pocket-id/pull/662/commits/1afd9e051ad72c4be44ff069099711df779e3063)

ItalyPaleAle (approved):

PR #669, opened 2025-06-22
fix : typo in french language

manuelrais:
There is a spelling mistake in the French translation. It concerns the message that appears when only one key is configured.


kmendell (approved):
Thanks!

Issue #358, opened 2025-03-19
🚀 Feature:


nanhoes:
### Feature description

Allow users to create their own login instead of admin creation.

### Pitch

I am using pocket as a login for jellyfin. I would like the user to be able to create their own pocket login so I dont have to do it myself for each user. The user will create there pocket login and passkey, then be able to sign-up/sign-in to jellyfin. Thanks for reading!

kmendell:
This can be done from the API, technically. or are you referring to a GUI Flow for this?

nanhoes:
GUI flow preferably as I would like it to be all done on the user side

kmendell:
@stonith404 Thoughts? basically adding a user based signup page, (probably with the ability to turn it off).

kmendell:
Putting the needs more upvotes label on till stonith gives his input.

stonith404:
Does it really make sense for users to sign up on an IDP? Allowing sign-ups would mean that anyone could register for all your services. As far as I know, Authentik and Authelia do not support user sign-ups.

I'm curious about your use-case. Do you have many users that you would have to register?

nanhoes:
Right, it’d probably be best if there are default user permissions on creation for the services. 

I have almost 20 jellyfin/jellyseerr users. 
I definitely could add each user to pocket manually then send each person their person invite. But as services start to transition to sso it’d be nice to have a one stop shop user sign up page, then ideally I can remove password auth for all exposed services.

kmendell:
> Right, it’d probably be best if there are default user permissions on creation for the services.
> 

Can you explain this part a bit more? This may seem more doable then a sign up flow

nanhoes:
Admin can set default permissions that will automatically be applied to new users (i.e. choosing which services the user will have access to). Admin should also be able to modify permissions for each user after creation. 

Example flow as admin:
- Set default permissions to only allow new users to have access to jellyfin, jellyseerr
- Create user, send invite link (or send time-sensitive invite link to a signup page for user to create their own account)
- User adds passkey, then login to permitted services

kmendell:
@nanhoes This can be achieved by restricting access to application with groups, then after you create the users just assign them to those groups. Is that not a acceptable solution? or am i misunderstanding?

nanhoes:
The main request for a sign-up page would minimize admin work as doing this for more than a handful of users would become cumbersome. 

[...]

PR #672, opened 2025-06-22
feat: self-service user signup

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/358

This works similarly to the login codes.

- An admin can create a Signup Token Link with a customizable Expiration and Usage Limit.
image

- Users can use this link to create their account and initial passkey.

# TODOS:

- [ ] Make a configurable "Opt-in" switch to allow signups
- [ ] Create Audit log events when a signup takes place with the generated links.


kmendell:
@stonith404 Can you do a quick pre-review to see if you like this? Before i start adding tests for it. Also that one test keeps failing and no idea why...

backend/internal/controller/user_controller.go
```diff
@@ -495,6 +502,131 @@ func (uc *UserController) updateUserGroups(c *gin.Context) {
...
+	userAgent := c.GetHeader("User-Agent")
+
+	user, accessToken, err := uc.userService.SignupWithToken(c.Request.Context(), token, input, ipAddress, userAgent)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	maxAge := int(uc.appConfigService.GetDbConfig().SessionDuration.AsDurationMinutes().Seconds())
+	cookie.AddAccessTokenCookie(c, maxAge, accessToken)
```
ItalyPaleAle, line 615: Do you think it's ok to let users be automatically signed in after they create an account? Or should we ask them to explicitly sign in after creating an account?

The first is more convenient; the latter may be a bit safer as we have a way to immediately confirm that the user saved the passkey correctly, so admins can catch problems right away.

I am really not sure which option is better.

backend/internal/dto/signup_token_dto.go
```diff
@@ -0,0 +1,26 @@
...
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+)
+
+type SignupTokenCreateDto struct {
+	ExpiresAt  time.Time `json:"expiresAt" binding:"required"`
+	UsageLimit int       `json:"usageLimit" binding:"required,min=1,max=100"`
+}
+
+type SignupTokenDto struct {
+	ID         string            `json:"id"`
```
ItalyPaleAle, line 15: Do we need an ID if we already have a token that's supposedly unique? Could that be the "primary key"?

backend/internal/service/user_service.go
```diff
@@ -650,3 +775,20 @@ func NewOneTimeAccessToken(userID string, expiresAt time.Time) (*model.OneTimeAc
 
 	return o, nil
 }
+
+func NewSignupToken(expiresAt time.Time, usageLimit int) (*model.SignupToken, error) {
+	// Generate a random token
+	randomString, err := utils.GenerateRandomAlphanumericString(16)
```
ItalyPaleAle, line 760: Should this be shorter than 16, if we expect users to need to type it and it's limited in use anyways?

backend/internal/model/signup_token.go
```diff
@@ -0,0 +1,28 @@
...
+
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+)
+
+type SignupToken struct {
+	Base
+
+	Token      string            `gorm:"uniqueIndex;not null" json:"token" sortable:"true"`
+	ExpiresAt  datatype.DateTime `gorm:"not null" json:"expiresAt" sortable:"true"`
+	UsageLimit int               `gorm:"not null;default:1" json:"usageLimit" sortable:"true"`
```
ItalyPaleAle: Should we allow unlimited usages (`UsageLimit = 0`) and without time limit (`ExpiresAt = 0 or NULL`) too?

backend/internal/service/user_service.go
```diff
@@ -650,3 +775,20 @@ func NewOneTimeAccessToken(userID string, expiresAt time.Time) (*model.OneTimeAc
...
+	if err != nil {
+		return nil, err
+	}
+
+	token := &model.SignupToken{
+		Token:      randomString,
+		ExpiresAt:  datatype.DateTime(expiresAt),
+		UsageLimit: usageLimit,
```
ItalyPaleAle, line 768: Here too... why enforcing limits? Can't these be unlimited in time and count?

backend/internal/service/user_service.go
```diff
@@ -630,6 +630,131 @@ func (s *UserService) disableUserInternal(ctx context.Context, userID string, tx
...
+		return model.User{}, "", err
+	}
+
+	// If usage limit reached, delete the token
+	if signupToken.IsUsageLimitReached() {
+		err = tx.WithContext(ctx).Delete(&signupToken).Error
+		if err != nil {
+			return model.User{}, "", err
+		}
+	}
```
ItalyPaleAle: We can avoid an extra query if the token has reached its max capacity
```suggestion
	// If usage limit reached, delete the token
	if signupToken.IsUsageLimitReached() {
		err = tx.WithContext(ctx).Delete(&signupToken).Error
		if err != nil {
			return model.User{}, "", err
		}
	} else {
        err = tx.WithContext(ctx).Save(&signupToken).Error
        if err != nil {
            return model.User{}, "", err
        }
    }
```

backend/internal/service/user_service.go
```diff
@@ -630,6 +630,131 @@ func (s *UserService) disableUserInternal(ctx context.Context, userID string, tx
...
+		First(&signupToken).
+		Error
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return model.User{}, "", &common.TokenInvalidOrExpiredError{}
+		}
+		return model.User{}, "", err
+	}
+
+	if !signupToken.IsValid() {
```
ItalyPaleAle: Should we delete the expired signup tokens here? Down below you delete the ones that have reached capacity, but if the token has just expired, it's not deleted from here.

ItalyPaleAle (commented):
I have focused my review on the backend changes, and I am good with them. Just leaving some small feedback, but nothing blocking

backend/internal/controller/user_controller.go
kmendell: Yes, because we need this for the passkey flow, at the end of the day if the admin generated the token or allows users to sign up its considered trusted.

backend/internal/dto/signup_token_dto.go
kmendell: It follows the same logic as the Onetime access codes, it uses the 'Base' model which includes and ID.

backend/internal/model/signup_token.go
kmendell: No logic is in place for that, but it follows the same way access tokens work, and api keys if a user wants to set a very unlimited time period thats fine but to stay consistant with how other things work we did it this way.

backend/internal/service/user_service.go
kmendell: This is here to enforce the usage limit. See comment above

backend/internal/service/user_service.go
kmendell: Its deleted by the scheduled job.

backend/internal/controller/user_controller.go
ItalyPaleAle: I was more thinking about making sure the user has saved the passkey. This is like, a way for them to confirm they have saved it.

If something were wrong, they'd find out the passkey wasn't saved only the 2nd login attempt, which can be hours later, when they've been using the account for a while. So the "damage" could be worse.

Anyways, not something I have strong opinions on.

kmendell: How would the user sign in then if we didn't do it here? Its the same situation either way to be honest.

ItalyPaleAle: Oh, I see what you're saying. Makes sense. Disregard my comment!

backend/internal/service/user_service.go
kmendell: I have it after saving as if the usage count reaches its max after this most recent use it will delete it. Your logic wouldnt delete it till the next time or scheduled job. Which could lead to issues. Unless im missing something.

backend/internal/service/user_service.go
kmendell: I had it at 32, but to follow suite with how the onetime acces codes work, elias said to make it 16 with the rate limiting as well. So its semi that same logic.

backend/internal/service/user_service.go
ItalyPaleAle: No...

The current logic:

1. Always updates the count in the DB
2. After updating the DB, checks if the count has reached max (using the in-memory value) and then deletes it

So the update is performed always.

I'm suggesting skipping the update if we already know (because of the in-memory count) that the count has reached max

kmendell: Ahh yeah i see it now, i was confused. I can change it later.

kmendell: [2fb7efe](https://github.com/pocket-id/pocket-id/pull/672/commits/2fb7efe37f019fb92e67a6e89e97d5a27c8a6990)[2fb7efe](https://github.com/pocket-id/pocket-id/pull/672/commits/2fb7efe37f019fb92e67a6e89e97d5a27c8a6990)

PR #679, opened 2025-06-25
chore: gin logging less noisy and no logging for localhost (internal …

11notes:
…access like health cheGin in its default configuration is way too noisy, to quiet it down a bit this simple PR will replace the default with new instance and ad a skip and skippaths directive to the logger.ck)


stonith404:
Thanks for your contribution :)

I think it would also be good if we exclude all GET requests to 
- `/_app/*`
- `/fonts/*`
- `/api/application-configuration/background-image`
- `/api/application-configuration/logo`

because this are all static assets too.

kmendell:
@11notes This looks good to me so far, but i was trying to fix the linter issue, but for some reason git wont let me push to your branch, just keep throwing permission denied. Can you fix that linter issue? (its just a bad formatted file)

11notes:
@kmendell I will delete this PR and make a new one with the correct linter settings and a better filter.

PR #681, opened 2025-06-25
gin: less noisy logging by excluding certain paths via regex and excluding localhost all together

11notes:
gin: less noisy logging by excluding certain paths via regex and excluding localhost all together


backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -47,7 +49,24 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
...
+	r := gin.New()
+	r.Use(gin.LoggerWithConfig(gin.LoggerConfig{Skip:func(c *gin.Context) bool {	
+		for _, path := range loggerSkipPathsRegExp {
+			if path.MatchString(c.Request.URL.String()) {
+				return true
+			}
+		}
+	
+		// do not log if client is localhost (health check on /healthz)
+		return strings.Split(c.Request.RemoteAddr, ":")[0] == "127.0.0.1"
```
ItalyPaleAle: A few things on this:

1. Not all requests coming from "localhost" are healthchecks. For example, people running Pocket ID as a standalone binary and putting a reverse proxy in front would see all requests from "localhost" too. So this would break a lot of things...
   - If the goal is to exclude `/healthz`, consider excluding the path explicitly
2. "localhost" can be `127.0.0.1` but can also be `::1` when the host uses IPv6
3. Aside from the above, `strings. Split` is quite inefficient as it scans the entire string and causes allocations on the heap; if we do go down this path (but note my reservations above), consider using `strings. IndexByte(c.Request.RemoteAddr, ':')` to get the index, and if it's >= 0 then return `c.Request.RemoteAddr[:idx]`

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -47,7 +49,24 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
 		gin.SetMode(gin.TestMode)
 	}
 
-	r := gin.Default()
+	// do not log these URL's
+	loggerSkipPathsRegExp := []*regexp.Regexp{
+		regexp.MustCompile(`\/api\/application-configuration.*`),
```
ItalyPaleAle: Isn't this a very valid path we'd want logged?

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -47,7 +49,24 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
...
+		regexp.MustCompile(`\/_app\/.*`),
+		regexp.MustCompile(`\/fonts/.*`),
+	}
+
+	r := gin.New()
+	r.Use(gin.LoggerWithConfig(gin.LoggerConfig{Skip:func(c *gin.Context) bool {	
+		for _, path := range loggerSkipPathsRegExp {
+			if path.MatchString(c.Request.URL.String()) {
```
ItalyPaleAle: Can we avoid using multiple regex matchings here, since this is in the hot path of each request? Since you're only looking for prefixes, can we just use `strings.HasPrefix`?

backend/internal/bootstrap/router_bootstrap.go
11notes: 1.	True, I only thought about containers, not bare metal installs, healtz endpoint does not appear in the c.Request.URL object, only / for some reason, I don't know if the Healtzh middleware changes the routing?
3.	Correct again
4.	I would do it with RegExp too

backend/internal/bootstrap/router_bootstrap.go
11notes: So far I’ve not seen any end point under that path that needs to be logged, but you are the creator so you know better. What would be the correct string? Only exclude favicon?

backend/internal/bootstrap/router_bootstrap.go
11notes: Sure, whatever you prefer. I’m not familiar enough with Go to know the exact performance difference this would make in terms of μs.

backend/internal/bootstrap/router_bootstrap.go
ItalyPaleAle: I am not the creator, just a contributor :) 

I just think that all `/api*` routes should be included in logs since they are API calls. No strong preference on the other routes you've excluded

backend/internal/bootstrap/router_bootstrap.go
ItalyPaleAle: I think for excluding healthz, you can exclude `/healthz`, with exact match. I don't think it matters if requests are coming from localhost, right?

backend/internal/bootstrap/router_bootstrap.go
kmendell: Elias said that we can exclude any static asset api paths in the other PR. This is fine. 

It includes things like the app logo etc.

backend/internal/bootstrap/router_bootstrap.go
ItalyPaleAle: It can be substantial (regexs in general don't have great perf, in Go or not). This is on my machine:

```go
func BenchmarkPrefixMatchRegex(b *testing.B) {
	str := "/_app/some/path"
	re := regexp.MustCompile(`/_app/.*`)

	b.Run("HasPrefix", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			_ = strings.HasPrefix(str, "/_app/")
		}
	})

	b.Run("Regexp", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			_ = re.MatchString(str)
		}
	})
}
```

Result:

```
goos: Darwin
goarch: arm64
cpu: Apple M2 Pro
BenchmarkPrefixMatchRegex
BenchmarkPrefixMatchRegex/HasPrefix
BenchmarkPrefixMatchRegex/HasPrefix-12         	475927251	         2.338 ns/op	       0 B/op	       0 allocs/op
BenchmarkPrefixMatchRegex/Regexp
BenchmarkPrefixMatchRegex/Regexp-12            	 9412059	       130.8 ns/op	       0 B/op	       0 allocs/op
```

So the prefix match here is around 98% faster.

Note: I'm not _necessarily_ against regexs, I use them a lot too. I'm being a bit of a sticker here, in this specific case, since you're just doing prefix matching _and_ this is a hot path (it's part of each request)

backend/internal/bootstrap/router_bootstrap.go
ItalyPaleAle: So right now these are the routes that are under `/api/application-configuration`:

https://github.com/pocket-id/pocket-id/blob/1a77bd9914ea01e445ff3d6e116c9ed3bcfbf153/backend/internal/controller/app_config_controller.go#L33-L45

No concerns with lines 37-42, but I think the other ones are API calls we may want logged, since they retrieve data?

kmendell: Oh yes if data is being retrieved than we will want that. 

I miss read the regex at first.

ItalyPaleAle: Actually, I'd argue we should exclude L37-39 only (GET calls). L4-42 are PUT calls, which update the data, and we may want those logged too

kmendell: Yes only GET requests is what were saying.

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -47,7 +48,23 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
 		gin.SetMode(gin.TestMode)
 	}
 
-	r := gin.Default()
+	// do not log these URLs
+	loggerSkipPathsPrefix := []string{
+		"GET /api/application-configuration",
```
ItalyPaleAle: So as per the other thread (which I know had a lot of messages), we must not exclude all routes that are `GET /api/application-configuration`, but **only** these **exact** routes:

```
GET /application-configuration/logo
GET /application-configuration/background-image
GET /application-configuration/favicon
```

(With exact match I mean matching the exact route, or exact route + `/`)

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -47,7 +48,23 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
...
+		"GET /_app",
+		"GET /fonts",
+		"HEAD /healthz",
```
ItalyPaleAle, line 59:
```suggestion
		"GET /healthz",
```

ItalyPaleAle: Also this one needs to be an exact match too, not a prefix match

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -47,7 +48,23 @@ func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
...
+	}
+
+	r := gin.New()
+	r.Use(gin.LoggerWithConfig(gin.LoggerConfig{Skip:func(c *gin.Context) bool {	
+		for _, prefix := range loggerSkipPathsPrefix {
+			if strings.HasPrefix(fmt.Sprintf("%s %s", c.Request.Method, c.Request.URL.String()), prefix){
```
ItalyPaleAle:
```suggestion
			if strings.HasPrefix(c.Request.Method + " " + c.Request.URL.String(), prefix){
```

This is another optimization that matters in a hot path. `fmt.Sprintf` causes allocations on the heap, which means that it's extra work for the GC. For simple string concatenation, _especially_ in a hot path, please use the `+`

backend/internal/bootstrap/router_bootstrap.go
kmendell: The 3 that should NOT be logged are  

```go
group.GET("/application-configuration/logo", acc.getLogoHandler) 
group.GET("/application-configuration/background-image", acc.getBackgroundImageHandler) 
group.GET("/application-configuration/favicon", acc.getFaviconHandler) 
```

ItalyPaleAle:
@11notes thanks for the PR. I'm ok with it but can you please fix the CI failures (linter)?

stonith404 (approved):

11notes:
Perfect, thank you very much for merging this PR.

PR #682, opened 2025-06-25
feat: Encrypt private keys saved on disk and in DB

### LINKED ISSUE #580: 🚀 Feature: Encrypt private key saved to DB
### Feature description

The private key is currently stored on disk un-encrypted. Anyone with access to the file system can get the "keys to the kingdom".

The private key could optionally be encrypted when stored to disk. The decryption key is provided to Pocket ID via env vars (e.g. `ENCRYPTION_KEY`) or with the name of another file containing the key (`ENCRYPTION_KEY_FILE`).

This allows also using a command to retrieve the key encryption key, e.g. to fetch it from a vault of some sort.

### Pitch

Data stored on disk is not always safe and can be leaked or stolen by malicious code. Encrypting the private key is a simple and effective way to keep the "keys to the kingdom" safe.

ItalyPaleAle:
Fixes #580

Implements the design described here: https://github.com/pocket-id/pocket-id/issues/580#issuecomment-2918342289

- Adds support for storing private keys in the database
- Adds support for encrypting the private keys stored on disk (optional) or in the database (required)
- Adds support for running Pocket ID with in-memory only (ephemeral) keys

Adds these new config options:

- `KEYS_STORAGE`, which can be `file`, `database`, or `memory`. The default is `file` if no key encryption key is passed, or `database` if there's a key encryption key
- `KEYS_ENCRYPTION_KEY` allows passing a key to use to encrypt keys stored on disk or DB
- `KEYS_ENCRYPTION_KEY_FILE` is akin to `KEYS_ENCRYPTION_KEY` but contains the path to a file with the key to load

When keys are encrypted, they are tied to a specific instance of Pocket ID (identified by the InstanceID internal config setting).

Also note that keys are encrypted with AES-CBC-HMAC ("Encrypt-then-MAC", using a 256-bit key for encryption and SHA-512 for HMAC; additionally, uses PKCS#7 for padding). This is an authenticated cipher (AEAD) but that doesn't have the issues with nonce re-use like AES-GCM. Support for this isn't available out-of-the-box in the Go's standard library, so this uses code copied from the Dapr project (licensed under Apache2)


backend/internal/common/env_config.go
```diff
@@ -18,79 +20,119 @@ const (
...
-	GeoLiteDBPath:      "data/GeoLite2-City.mmdb",
-	GeoLiteDBUrl:       MaxMindGeoLiteCityUrl,
-	LocalIPv6Ranges:    "",
-	UiConfigDisabled:   false,
-	MetricsEnabled:     false,
-	TracingEnabled:     false,
-	TrustProxy:         false,
-	AnalyticsDisabled:  false,
-}
+var EnvConfig = defaultConfig()
```
kmendell, line 54: If we do this should we update all the locations that use common? or revert this to the old behavior to keep it consistent?

backend/internal/common/env_config.go
```diff
@@ -18,79 +20,119 @@ const (
...
 	switch EnvConfig.DbProvider {
 	case DbProviderSqlite:
 		if EnvConfig.DbConnectionString == "" {
-			log.Fatal("Missing required env var 'DB_CONNECTION_STRING' for SQLite database")
+			EnvConfig.DbConnectionString = defaultSqliteConnString
 		}
 	case DbProviderPostgres:
 		if EnvConfig.DbConnectionString == "" {
-			log.Fatal("Missing required env var 'DB_CONNECTION_STRING' for Postgres database")
+			return errors.New("missing required env var 'DB_CONNECTION_STRING' for Postgres database")
```
kmendell, line 103: Shouldn't this be a fatal error if its missing the postgres db connection string?

backend/internal/service/jwt_service.go
```diff
@@ -229,8 +265,8 @@ func (s *JwtService) VerifyAccessToken(tokenString string) (jwt.Token, error) {
 		jwt.WithValidate(true),
 		jwt.WithKey(alg, s.privateKey),
 		jwt.WithAcceptableSkew(clockSkew),
-		jwt.WithAudience(common.EnvConfig.AppURL),
-		jwt.WithIssuer(common.EnvConfig.AppURL),
+		jwt.WithAudience(s.appUrl),
```
kmendell: These are the changes i was referring to in my comment above.

backend/internal/utils/crypto/crypto.go
kmendell, line 1: May be a nit pick, but should this be crypto_service not just a util? Thinking outloud.

kmendell (commented):
Just some overview comments, i dont have time to fuly review this at the moment.

backend/internal/common/env_config.go
ItalyPaleAle: I changed the signature of the function to return an error: https://github.com/pocket-id/pocket-id/pull/682/files/742673d4f0def6332164086bfd84e28e8bace803#diff-e13de9f007e726aab0c4ee2bc26b7731372ac116c57ceb7d71d06071ba6d3b82R56-R61

Then the `init` method calls this, and if it returns an error, it shows a fatal log.

The reason for this change is so it's possible to add a unit test

backend/internal/common/env_config.go
ItalyPaleAle: No, it doesn't need to be changed.

This is just the initial value. Next, the `init` method (called when the app starts) will call `parseConfig`, which sets values to this:

https://github.com/pocket-id/pocket-id/pull/682/files/742673d4f0def6332164086bfd84e28e8bace803#diff-e13de9f007e726aab0c4ee2bc26b7731372ac116c57ceb7d71d06071ba6d3b82R90

Creating a `defaultConfig()` function was too done so I could write unit tests

backend/internal/service/jwt_service.go
ItalyPaleAle: This change is unrelated from your comment above (the value/behavior of `common.EnvConfig` hasn't changed), but I may be able to revert it anyways. I'll take a look

backend/internal/utils/crypto/crypto.go
ItalyPaleAle: Normally a "service" is called that way because it's long-running and has a lifecycle. This package does just have util functions.

backend/internal/service/jwt_service.go
ItalyPaleAle: Ok I partly reverted this change. I am now storing `envConfig` in the object instead of `appUrl` directly.

This is more convenient than using `common.EnvConfig` because it can be used with unit tests.

Both `common.EnvConfig` and the `envConfig` property of this object are pointers to the same underlying object. But storing the pointer makes it possible to create unit tests without having to deal with global variables.

backend/internal/common/env_config.go
kmendell: sure my point is other parts in pocket id dont follow this same pattern, and can cause confusion for devs i collorating is what im getting at. In my eyes should try to follow the same patterns in all parts of the codebase.

ItalyPaleAle: That's fine. Maybe we could do that in another PR? This is already pretty large

kmendell: Sure, i was just saying we should keep it aligned through the codebase im not against the change, hopefully you get what im saying.

stonith404:
Thanks @ItalyPaleAle, this is the next PR I'll review.

backend/internal/model/kv.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+type KV struct {
+	Key   string `gorm:"primaryKey;not null"`
+	Value *string
+}
+
+// TableName overrides the table name used by KV to `kv`
```
kmendell, line 8: Do we need this table name definition? I don't see it being used anywhere.

backend/internal/common/env_config.go
```diff
@@ -18,79 +20,119 @@ const (
...
+}
+
+func defaultConfig() EnvConfigSchema {
+	return EnvConfigSchema{
+		AppEnv:             "production",
+		DbProvider:         "sqlite",
+		DbConnectionString: "",
+		UploadPath:         "data/uploads",
+		KeysPath:           "data/keys",
+		KeysStorage:        "", // "database" or "file" or "memory"
```
kmendell: I would mark the File based method as deprecated and will be removed in the next major release maybe? based on this comment: https://github.com/pocket-id/pocket-id/issues/580#issuecomment-2917688008, unless im misunderstanding it.

backend/internal/common/env_config.go
```diff
@@ -18,79 +20,119 @@ const (
...
 	if err != nil {
-		log.Fatal("APP_URL is not a valid URL")
+		return errors.New("APP_URL is not a valid URL")
 	}
 	if parsedAppUrl.Path != "" {
-		log.Fatal("APP_URL must not contain a path")
+		return errors.New("APP_URL must not contain a path")
 	}
+
+	switch EnvConfig.KeysStorage {
```
kmendell, line 117: This key storage logic seems convoluted. 

Could this be simplified (e.g., require explicit KEYS_STORAGE and error if keys are missing when needed)?

Just set the default key storage to `file` and simplify this logic?

backend/internal/common/env_config.go
kmendell, line 1: I know you said this was for a unit test, but i cant wrap my head around why add a new multi step approach vs the simple env struct parsing we had before just for a unit test?

backend/internal/utils/crypto/pkcs7_padding.go
kmendell, line 1: Is it needed to "homegrow" all this padding code vs using a library out there? like `andreburgaud/crypt2go` (just an example i found). I feel like this may be unmaintainable if a issue would occur in this code.

backend/internal/model/kv.go
ItalyPaleAle: Yes otherwise GORM's default is to name the table `kvs`

backend/internal/common/env_config.go
ItalyPaleAle: This is up to the maintainers to decide.

The current design (https://github.com/pocket-id/pocket-id/issues/580#issuecomment-2918342289) doesn't say anything about deprecation, but it's an option.

One thing to note is that if we do deprecate this:

1. Users will always have to provide a key encryption key
2. It may break some scenarios such as storing the key in a file that is mounted as Docker Secret or K8s secret volume.

backend/internal/common/env_config.go
ItalyPaleAle: Ok so here's the problem.

Unit tests that deal with global state are a bit of a PITA. For example in `jwt_service_test.go` I need to initialize the service with different values.

Using global variables/state is possible but requires some convoluted things like this: https://github.com/pocket-id/pocket-id/pull/682/files/5b28eb50d6e31908e56572257bda08811ae68099#diff-1a83660cb312ce7c35ec2e99a6bf94d422d33c1350363d256b6687473310475aR11-R15

Additionally, I wanted a way to have a method like `defaultConfig()` that would print out the default config, so `env_config_test.go` can use that to test parsing with the default config as base

backend/internal/utils/crypto/pkcs7_padding.go
ItalyPaleAle: That library uses ECB mode which is not a good idea.

I had originally implemented this for Dapr a few years ago, and copied it here with almost no change. Importing the Dapr dependency directly wouldn't work because it has too much stuff that's tied with other components.

stonith404:
Thanks, the PR looks generally good to me. A few points:
- Wouldn't it make sense to store the encrypted key in the `app_config_variables` table with the `internal` tag? This would have the same behavior but we don't need a new `kvs` table. We de the same with the [`instanceID`](https://github.com/pocket-id/pocket-id/blob/5b28eb50d6e31908e56572257bda08811ae68099/backend/internal/model/app_config.go#L48).
- Does the memory provider really make sense? Storing the keys in the memory would invalidate all tokens on every restart, I don't really see a use-case here.
- Maybe we can rename the `KEYS_ENCRYPTION_KEY` environment variable to just `ENCRYPTION_KEY`? This would allow us to use this variable to encrypt other things in the future.

backend/internal/utils/crypto/pkcs7_padding.go
stonith404: I agree with @kmendell, I think these crypto files are a bit hard to maintain, if you aren't experienced with crypto. But @ItalyPaleAle if you really think there isn't a good library, I guess we have to do it like this.

backend/internal/common/env_config.go
kmendell: Maybe i missunderstood the comment, I was looking at the serverless part of it. But maybe im making stuff up

backend/internal/utils/crypto/pkcs7_padding.go
ItalyPaleAle: I will switch to AES-GCM which is available inside the Go standard library. As long as we don't encrypt more than ~4 billion values, it should be fine.

ItalyPaleAle:
@stonith404 thanks for the review. To answer your questions:

> Wouldn't it make sense to store the encrypted key in the app_config_variables table with the internal tag? This would have the same behavior but we don't need a new kvs table. We de the same with the instanceID.

It's an option, but I would prefer we keep it in a separate table. The reason is that everything in app_config_variables gets loaded into DbConfig and is available throughout the application for the entire lifetime. There's no reason to keep the key in-memory for the entire lifetime of the app, nor to make it so widely accessible.

(I could move the instance ID to the KV table too?)

> Does the memory provider really make sense? Storing the keys in the memory would invalidate all tokens on every restart, I don't really see a use-case here.

I'd say the use cases would be far and few indeed, perhaps just some test environments. It's a very simple implementation however (just 2 no-op functions), so was an easy add. I am not particularly attached to it, and happy to remove it. As an alternative it could be kept as an un-documented option.

> Maybe we can rename the KEYS_ENCRYPTION_KEY environment variable to just ENCRYPTION_KEY?

You got it

kmendell:
> @stonith404 thanks for the review. To answer your questions:
> 
> > Wouldn't it make sense to store the encrypted key in the app_config_variables table with the internal tag? This would have the same behavior but we don't need a new kvs table. We de the same with the instanceID.
> 
> It's an option, but I would prefer we keep it in a separate table. The reason is that everything in app_config_variables gets loaded into DbConfig and is available throughout the application for the entire lifetime. There's no reason to keep the key in-memory for the entire lifetime of the app, nor to make it so widely accessible.
> 

I thought that was the purpose of the `internal` flag i may be mistaken though..

stonith404:
> It's an option, but I would prefer we keep it in a separate table. The reason is that everything in app_config_variables gets loaded into DbConfig and is available throughout the application for the entire lifetime. There's no reason to keep the key in-memory for the entire lifetime of the app, nor to make it so widely accessible.

Yeah, okay, that makes sense. It would load the encrypted key into the memory unnecessarily. In my opinion I just find it a bit unnecessary to have a single table only for one value, I would just take unnecessary memory allocation as a drawback to keep the database structure simpler. The instance ID should still be in the `app_config_variables` table because it needs to be accessed multiple times and therefore it makes sense to keep it in the memory. 

But your solution is also alright for me, I would just prefer the other one.

> I'd say the use cases would be far and few indeed, perhaps just some test environments. It's a very simple implementation however (just 2 no-op functions), so was an easy add. I am not particularly attached to it, and happy to remove it. As an alternative it could be kept as an un-documented option.

Yeah, it's really simple, but because of that, I think that it makes sense to remove it since it would be very easy to add back in the future.

stonith404:
@kmendell The `internal` tag just hides the config variable from the endpoints, but everything else behaves like any other config variable.

ItalyPaleAle:
> Yeah, okay, that makes sense. It would load the encrypted key into the memory unnecessarily. In my opinion I just find it a bit unnecessary to have a single table only for one value, I would just take unnecessary memory allocation as a drawback to keep the database structure simpler. The instance ID should still be in the `app_config_variables` table because it needs to be accessed multiple times and therefore it makes sense to keep it in the memory. 
> 
> 
> 
> But your solution is also alright for me, I would just prefer the other one.
> 

Oh, it's less about the extra allocation (which isn't that huge of a deal) and more about making sure keys aren't in memory for longer than needed. It's not a huge risk as it was back in the days (where reading another process' memory was easy), but still feels "wrong" to keep a key in memory if it's not needed.

> 
> Yeah, it's really simple, but because of that, I think that it makes sense to remove it since it would be very easy to add back in the future.
> 

Ok!

backend/internal/common/env_config.go
stonith404: I would not deprecate it at the moment because it might be interesting to look into Docker Secrets (#685). I haven't looked into Docker Secrets but if they don't require file system access we can easily deprecate it.

ItalyPaleAle: Docker Secrets will read from a file. They are, however, read-only, so the key must be created outside of Pocket ID

ItalyPaleAle:
Also, I realized that with encrypted keys, it becomes essentially impossible to set keys manually (like using [custom private keys](https://pocket-id.org/docs/setup/advanced-topics#use-custom-private-keys)). Perhaps in a follow-up PR I can add another command to the binary (like `pocketed one-time-access-token`) that generates/replaces the private key, with the specific format, and stores it in the configured method (incl. in the database, encrypted).

kmendell:
> Also, I realized that with encrypted keys, it becomes essentially impossible to set keys manually (like using [custom private keys](https://pocket-id.org/docs/setup/advanced-topics#use-custom-private-keys)). Perhaps in a follow-up PR I can add another command to the binary (like `pocketed one-time-access-token`) that generates/replaces the private key, with the specific format, and stores it in the configured method (incl. in the database, encrypted).

Should this be something we include in this PR? If im understanding right if someone chooses to use this there key wont work? Or am i not understanding right.

ItalyPaleAle:
> > Also, I realized that with encrypted keys, it becomes essentially impossible to set keys manually (like using [custom private keys](https://pocket-id.org/docs/setup/advanced-topics#use-custom-private-keys)). Perhaps in a follow-up PR I can add another command to the binary (like `pocketed one-time-access-token`) that generates/replaces the private key, with the specific format, and stores it in the configured method (incl. in the database, encrypted).
> 
> Should this be something we include in this PR? If im understanding right if someone chooses to use this there key wont work? Or am i not understanding right.

I can do it in a subsequent PR, as a fast-follow.

They can continue to do what's in the docs, e.g. using the `step` CLI, but only IF the keys are stored on file AND not encrypted.

backend/internal/common/env_config.go
stonith404: @ItalyPaleAle I agree with Kyle, wouldn't it make sense to set it to "file" per default?

ItalyPaleAle: Up to you. I thought from the issue that the idea was that you wanted keys in the database preferably.

Without an encryption key the default (and only option) has to be file.

With an encryption key we can either default to file or DB. Let me know if you'd like me to change this

ItalyPaleAle:
The follow-up PR is #709 - this is currently in draft since the code won't work until this PR is merged.

backend/internal/common/env_config.go
kmendell: I talked with Elias, and we do think it should be defaulted to file since its being sett hat way anyways, and that could clean this up as bit as well. After that i think this should be good to go

ItalyPaleAle: Done

stonith404: Yes, but I think it would make sense to require the encryption key and store it in the database per default in the next breaking release.

kmendell (approved):

kmendell:
Once the Unit test is fixed this should be good to merge, Im not that familar with unit tests, but i can look later tonight, if you dont beat me too it.

ItalyPaleAle:
> Once the Unit test is fixed this should be good to merge, Im not that familar with unit tests, but i can look later tonight, if you dont beat me too it.

Oh my bad. I changed the default value and forgot to update the tests. Should be fixed now.

kmendell:
@ItalyPaleAle Can you also add some updated documentation for this as well?

ItalyPaleAle:
Thanks! I'll have docs together with the ones for #709

PR #695, opened 2025-06-28
fix: Auth fails when client IP is empty on Postgres

### LINKED ISSUE #686: 🐛 Bug Report: Cannot login when using Unix Socket (Cannot get client ip)
### Reproduction steps

I’m trying to use `pocket-id` behind nginx with a UNIX socket.  

NGINX config:
```
server {
	listen 0.0.0.0:443 ssl ;
	listen [::0]:443 ssl ;
	server_name example.org;
	http2 on;
	ssl_certificate /etc/nginx/ssl/lexample.org.pem;
	ssl_certificate_key /etc/nginx/ssl/example.org.key;
	location / {
		proxy_pass "http://unix:/run/pocket-id/pocket-id";
		proxy_set_header X-Real-IP $remote_addr;
		proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
		proxy_set_header X-Forwarded-Proto $scheme;
	}
}
```

And then, try to login with a client

Additional context:
https://github.com/gin-gonic/gin/issues/2718

### Expected behavior

Login correctly

### Actual Behavior

Something went wrong

### Version and Environment

v1.4.1
```
APP_URL=https://example.org
DB_CONNECTION_STRING=user=pocket-id dbname=pocket-id host=/run/postgresql
DB_PROVIDER=postgres
TRUST_PROXY=true
UNIX_SOCKET=/run/pocket-id/pocket-id
UNIX_SOCKET_MODE=0666
```

### Log Output

```
[GIN] 2025/06/27 - 11:35:36 | 200 |    3.448381ms |                 | GET      "/api/webauthn/login/start"
2025/06/27 11:35:46 Failed to get IP location: failed to parse IP address: ParseAddr(""): unable to parse IP

2025/06/27 11:35:46 github.com/pocket-id/pocket-id/backend/internal/service/audit_log_service.go:47 ERROR: invalid input syntax for type inet: "" (SQLSTATE 22P02)
[2.120ms] [rows:0] INSERT INTO "audit_logs" ("id","created_at","event","ip_address","country","city","user_agent","data","user_id") VALUES ($1$,$2$,$3$,$4$,$5$,$6$,$7$,$8$,$9$)
2025/06/27 11:35:46 Failed to create audit log: ERROR: invalid input syntax for type inet: "" (SQLSTATE 22P02)

2025/06/27 11:35:46 github.com/pocket-id/pocket-id/backend/internal/service/audit_log_service.go:67 ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)
[0.175ms] [rows:0] SELECT count(*) FROM "audit_logs" WHERE user_id = $1$ AND ip_address = $2$ AND user_agent = $3$
2025/06/27 11:35:46 Failed to count audit logs: ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)
[GIN] 2025/06/27 - 11:35:46 | 500 |   28.119935ms |                 | POST     "/api/webauthn/login/finish"
Error #01: commit unexpectedly resulted in rollback
```

ItalyPaleAle:
Fixes #686

Auth fails with Postgres when the client IP is empty, for example when proxies connect to Pocket ID via a Unix Domain Socket (UDS).

This was due to the fact that in Postgres we store IPs with the INET column type, which doesn't accept empty values.

The fix involves making the field nullabe.

Additionally, improved error handling throughout the process.


backend/internal/dto/audit_log_dto.go
```diff
@@ -10,7 +10,7 @@ type AuditLogDto struct {
 	CreatedAt datatype.DateTime `json:"createdAt"`
 
 	Event     model.AuditLogEvent `json:"event"`
-	IpAddress string              `json:"ipAddress"`
+	IpAddress *model.AuditLogIP   `json:"ipAddress,omitempty"`
```
kmendell: Does this need to be a pointer? Looking at how the AuthLogEvent type works, we should be able to follow that logic i think? basically just have the pointer in the function instead. Besides that this looks good to me

ItalyPaleAle: Yeah it didn't work without a pointer. This is because the model uses a pointer here (it's a nullable field), and `MapStructList` doesn't allow copying from pointer to non-pointer 🤷

(We may want to consider switching from a custom `MapStructList` to something like https://github.com/jinzhu/copier ?)

ItalyPaleAle: #698 then I can change this

stonith404:
Thanks! I think it would make more sense if we make the Sqlite column nullable too. Like you said it's not possible to alter a column in Sqlite but we can just create a new table and copy the data from the old table to the new one:
```sql
CREATE TABLE audit_logs_new
(
    id               TEXT NOT NULL PRIMARY KEY,
    created_at       DATETIME,
    event            TEXT NOT NULL,
    ip_address       TEXT, 
    user_agent       TEXT NOT NULL,
    data             BLOB NOT NULL,
    user_id          TEXT REFERENCES users,
    country          TEXT,
    city             TEXT
);

INSERT INTO audit_logs_new
SELECT id, created_at, event, ip_address, user_agent, data, user_id, country, city
FROM audit_logs;

DROP TABLE audit_logs;

ALTER TABLE audit_logs_new RENAME TO audit_logs;

```

Even though the migration isn't very sexy, I think it makes the code more maintainable if we don't have to differentiate between Sqlite and Postgres.

ItalyPaleAle:
@stonith404 Your suggestion makes sense, I'll rework that. (Just a note that the table can be large so the migration may take a while)

While I'm at it... How would you feel about removing the country and city columns and make them part of the JSON "data"? Since they aren't always populated and aren't searchable anyways.

kmendell:
> @stonith404 Your suggestion makes sense, I'll rework that. (Just a note that the table can be large so the migration may take a while)
> 
> While I'm at it... How would you feel about removing the country and city columns and make them part of the JSON "data"? Since they aren't always populated and aren't searchable anyways.

They are used by the global audit log, so they are searched through, but it can easily be redone to work, but is there really a point to changing this? What is the current approach affecting? I dont see the point behind it.

ItalyPaleAle:
I guess there are now queries for those columns:

https://github.com/ItalyPaleAle/pocket-id/blob/796ad3c186d7dbe8afa3d6ec022e1f00b8fe3cb5/backend/internal/service/audit_log_service.go#L162-L169

I was thinking it could be for consistency with the other fields that are in the JSON payload.

ItalyPaleAle:
Changed to use nullable fields everywhere

backend/internal/dto/audit_log_dto.go
```diff
@@ -1,22 +1,21 @@
...
 
-	Event     model.AuditLogEvent `json:"event"`
-	IpAddress string              `json:"ipAddress"`
-	Country   string              `json:"country"`
-	City      string              `json:"city"`
-	Device    string              `json:"device"`
-	UserID    string              `json:"userID"`
-	Username  string              `json:"username"`
-	Data      model.AuditLogData  `json:"data"`
+	Event     string            `json:"event"`
```
ItalyPaleAle, line 11: Changed to string since this is now possible thanks to using the copier library

backend/resources/migrations/postgres/20250628000000_audit_logs_ip_null.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE audit_logs ALTER COLUMN ip_address DROP NOT NULL;
+
+-- Add missing indexes
+CREATE INDEX idx_audit_logs_created_at ON audit_logs(created_at);
```
ItalyPaleAle, line 4: These columns were used in WHERE or ORDER BY clauses but did not have an index

stonith404 (approved):
Thanks for the changes!

PR #698, opened 2025-06-29
refactor: use github.com/jinzhu/copier for MapStruct

ItalyPaleAle:
The current implementation of MapStruct has some limitations, including the fact that it requires the same type in source and destination, and doesn't allow copying from pointer to non-pointer.

Switch the implementation to github.com/jinzhu/copier which is very widely used and comprehensive

Note: this doesn't change the DTO structs, but it can be done in a subsequent PR


stonith404 (approved):
Nice, thank you!

PR #704, opened 2025-06-30
fix: token introspection authentication not handled correctly

### LINKED ISSUE #700: 🐛 Bug Report: Introspection Endpoint - Client id or secret not provided
### Reproduction steps

- Create OIDC Client with secret
- obtain an access token
- try to do a request against the introspection endpoint
  - with basic auth `clientid:clientsecret` as base64
  - body with `token=<accessToken>`

### Expected behavior

I would at least expect an error message that you can do something with, because clientId and secret were specified. 

### Actual Behavior

No matter what i do, i always get an error message.

- i have an older oidc client defined there ( public client ) without secret, if i try to use this one i am getting: `"Invalid client secret"`
- if i try to create a new client which looks similar to the client above, i get `"Client id or secret not provided"`

### Version and Environment

Version: 1.5.0 (updated from 1.1.0 - only noticed it because an app started to log auth errors)

### Example:
```sh
curl --location --request POST '<instance>/api/oidc/introspect' \
--header 'Authorization: Basic NzQzYjJjZTctNjcwZC00MTA0LTk3ODgtODhkN2VlMWM4ZDlhOm1KOHlZOW5CQm9oSHNOYmFHbXBBNmZnNmt6dzNSSUZO' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'token=<token>'
```

### Log
```
[GIN] 2025/06/29 - 13:32:27 | 400 |     320.247µs |     192.168.0.1 | POST     "/api/oidc/introspect"
Error #01: client id or secret not provided
```

### Response 
```json
{
    "error": "Client id or secret not provided"
}
```

stonith404:
Fixes #700


backend/internal/service/oidc_service.go
```diff
@@ -490,6 +490,11 @@ func (s *OidcService) createTokenFromRefreshToken(ctx context.Context, input dto
 }
 
 func (s *OidcService) IntrospectToken(ctx context.Context, creds ClientAuthCredentials, tokenString string) (introspectDto dto.OidcIntrospectionResponseDto, err error) {
+	client, err := s.verifyClientCredentialsInternal(ctx, s.db, creds)
```
ItalyPaleAle: I think this change will break federated credentials, because the client ID is not available in creds. We need to read it from the token first

kmendell: Thats what this does no? 

Starting at line 1389

```go
	// Determine the client ID based on the authentication method
	switch {
	case input.ClientAssertionType == ClientAssertionTypeJWTBearer && input.ClientAssertion != "":
		// Extract client ID from the JWT assertion's 'sub' claim
		clientID, err = s.extractClientIDFromAssertion(input.ClientAssertion)
		if err != nil {
			slog.Error("Failed to extract client ID from assertion", "error", err)
			return nil, &common.OidcClientAssertionInvalidError{}
		}
	case input.ClientID != "":
		// Use the provided client ID for other authentication methods
		clientID = input.ClientID
	default:
		return nil, &common.OidcMissingClientCredentialsError{}
	}
```

ItalyPaleAle: I see it now, it should be fine!

kmendell (approved):

backend/internal/service/oidc_service.go
```diff
@@ -1385,46 +1382,61 @@ func clientAuthCredentialsFromCreateTokensDto(d *dto.OidcCreateTokensDto) Client
...
 		if err != nil {
 			log.Printf("Invalid assertion for client '%s': %v", client.ID, err)
 			return nil, &common.OidcClientAssertionInvalidError{}
 		}
-		return &client, nil
+		return client, nil
 
 	// There's no credentials
 	// This is allowed only if the client is public
 	case client.IsPublic:
```
ItalyPaleAle: As discussed on Discord, this still allows public clients to make calls without passing a client secret, just a client ID. Adding this comment to make sure PR isn't merged without this change.

stonith404: Thanks, should be fixed now.

ItalyPaleAle (approved):
LGTM, thanks!

PR #709, opened 2025-07-02
feat: add "key-rotate" command

ItalyPaleAle:
Follow-up from #682 (see: https://github.com/pocket-id/pocket-id/pull/682#issuecomment-3025838190 )

Adds a new CLI command "pocket-id key-rotate" which replaces the token signing key with a new one. The command allows passing the key algorithm, allowing the use of RSA keys of different sizes (RS384, RS512 - in addition to the default RS256), ECDSA keys (ES256, ES384, ES512), and EdDSA (with curve Ed25519)

This also adds [spf13/cobra](https://github.com/spf13/cobra) to handle the various sub-commands. Cobra makes it easier to parse CLI flags, and also automatically generates help pages with the `--help` flag.


kmendell (approved):
Look good to me, @stonith404 ill let you do the final review.

stonith404 (approved):
Thank you! I really like the new structure.

PR #716, opened 2025-07-03
feat: distroless container additional variant + healthcheck command

### LINKED ISSUE #680: 🚀 Feature: Distroless, rootless pocket-id
### Feature description

Provide additional features in the form of

- distroless docker
- rootless configuration

### Pitch

https://github.com/11notes/docker-pocket-id

Merge the changes in this docker-pocket-id repostiory! :)

ItalyPaleAle:
Fix #680

Adds a `-distroless` variant for the container image, which is based on `gcr.io/distroless/static-debian12:nonroot`

Because the distroless image doesn't have curl, in order to perform healthchecks I have added a new sub-command to the pocket-id binary `pocket-id healthcheck`, which performs a healthcheck on a running instance of Pocket ID. This is a simpler alternative to shipping the curl binary. (Note that curl is still bundled in the non-distroless image, for backwards compatibility)

### DIFF
--- a/.github/workflows/build-next.yml
+++ b/.github/workflows/build-next.yml
@@ -73,10 +73,24 @@ jobs:
           push: true
           tags: ${{ env.DOCKER_IMAGE_NAME }}:next
           file: Dockerfile-prebuilt
-
+      - name: Build and push container image (distroless)
+        uses: docker/build-push-action@v6
+        id: container-build-push-distroless
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ env.DOCKER_IMAGE_NAME }}:next-distroless
+          file: Dockerfile-distroless
       - name: Container image attestation
         uses: actions/attest-build-provenance@v2
         with:
           subject-name: "${{ env.DOCKER_IMAGE_NAME }}"
           subject-digest: ${{ steps.build-push-image.outputs.digest }}
+          push-to-registry: true
+      - name: Container image attestation (distroless)
+        uses: actions/attest-build-provenance@v2
+        with:
+          subject-name: "${{ env.DOCKER_IMAGE_NAME }}"
+          subject-digest: ${{ steps.container-build-push-distroless.outputs.digest }}
           push-to-registry: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,14 +29,12 @@ jobs:
         uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-
       - name: Set DOCKER_IMAGE_NAME
         run: |
           # Lowercase REPO_OWNER which is required for containers
           REPO_OWNER=${{ github.repository_owner }}
           DOCKER_IMAGE_NAME="ghcr.io/${REPO_OWNER,,}/pocket-id"
           echo "DOCKER_IMAGE_NAME=${DOCKER_IMAGE_NAME}" >>${GITHUB_ENV}
-
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
         with:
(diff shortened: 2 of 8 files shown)


kmendell:
Not sure if its a concern, but your image is 120MB image (the bottom one), the top 3 are my attempts at it, again just something i noticed when building it, as i know one of the points was small container size. 
image

ItalyPaleAle:
@kmendell can you share your Dockerfile please? and can you try re-building since we've added a lot of new code?

By the way, if I build it locally, it's 57MB for me
image

A bit surprised it's double the size :O

kmendell:
https://github.com/pocket-id/pocket-id/tree/feat/distroless-image, was my initial testing havent touched in it 7 days though, how were you building it? I just built the binaries then built the container.

kmendell:
Docker desktop could be buggin too though, however the command line still shows the same 
image

ItalyPaleAle:
Is that build multi-arch? Mine is single-arch, and would explain why yours is roughly double.

As for the size difference, seems that your Dockerfile adds this:

```
upx --best --lzma ${BUILD_BIN}
```

Looking at [upx](https://github.com/upx/upx), it's a compressor that reduces the size of the binary. Would explain why the single-arch build could be a few MBs smaller.

But I don't know how I feel about compressing the executable. Our binaries are pre-built outside of Docker (because it's a lot faster, especially for multi-arch), and then they are signed. If a binary is re-compressed, it's not the same as the original one, so the signature wouldn't match. And IMHO I think having reproducible, signed builds is very important for a security-related app like pocket-id.

kmendell:
Thats probably the one i closely modified from 11notes, as my starter , but theres alot going on, so i started a sperate one, which is what those images are based on

stonith404 (approved):
Looks good to me, thanks!

As I remember, the only downside of the distroless image is that if the user wants to use a bind mount for the `data` folder, they have to create the folder and change the permissions on the host first, right? Did I miss any downsides or additional steps that we need to document?

We might want to use the distroless image as default image in the next breaking release, what are your thoughts on this @kmendell @ItalyPaleAle. It would make the set up a bit more complicated but we could recommend the users to use named volumes instead of bind mounts which don't need additional creation and permission steps.

ItalyPaleAle:
> As I remember, the only downside of the distroless image is that if the user wants to use a bind mount for the `data` folder, they have to create the folder and change the permissions on the host first, right? Did I miss any downsides or additional steps that we need to document?

That is correct and the biggest downside. Another downside is that there's no shell so users can't `docker exec -it bash pocket-id` and navigate inside the container.

> We might want to use the distroless image as default image in the next breaking release, what are your thoughts on this @kmendell @ItalyPaleAle. It would make the set up a bit more complicated but we could recommend the users to use named volumes instead of bind mounts which don't need additional creation and permission steps.

I don't have an opinion on the next breaking release. I am ok offering this as an option (for example in the "advanced setup" section) for now. We can also see how users respond to that and decide later based on feedback too.

kmendell:
@stonith404 We could offer the distroless as default in 2.0, we should always have the old version as maybe a `debug` tag so if we need the shell access for something we can always have those extra tools maybe? Thinking outloud on that one

ItalyPaleAle:
Since it's been approved, I'm gonna merge this just so we can test the release pipelines and make sure it's all good, sounds good?

11notes:
Maybe [this guide](https://github.com/11notes/RTFM/blob/main/linux/container/image/distroless.md#user-content-fn-1-e8f44bdc8e44362cc416ad7f4e86d6ef) helps you to understand how to debug distroless containers. You can always ask when you need or want to know more how to create proper and secure images @ItalyPaleAle, @stonith404 and @kmendell.

Thanks for adding the internal healthcheck @ItalyPaleAle, could remove static curl from my image thanks to that, shrinking it to less than 15MB 😊.

ovizii:
The changelog for v1.6.0 states links here and states:

> add distroless container additional variant + healthcheck command

Any documentation on which tag to use for the distroless variant?

ItalyPaleAle:
> The changelog for v1.6.0 states links here and states:
> 
> > add distroless container additional variant + healthcheck command
> 
> Any documentation on which tag to use for the distroless variant?

Thanks for reminding about the docs. I've opened pocket-id/website#126

You can add `-distroless` to the container image name. However, because distroless containers are non-root, you'll need to set the permissions on the filesystem as explained in https://pocket-id.org/docs/advanced/container-security-hardening#prerequisites

PR #724, opened 2025-07-06
fix: ensure user inputs are normalized

ItalyPaleAle:
User inputs such as names can contain UTF-8 sequences, and they may not be in a normalized form, so they may not show up as identical when they should be.

This PR makes sure that all strings that are not identifiers (like UUIDs) are normalized before being stored in the database.

> For more info on Unicode normalization, you can refer to [this article](https://withblue.ink/2019/03/11/why-you-need-to-normalize-unicode-strings.html) I wrote a few years ago


kmendell:
i'm curious if Zod already does this in the front end, or maybe it can?, i haven't looked into it yet, but since that where the input originates wouldn't it make more sense to normalize it there?

ItalyPaleAle:
I think this stuff is best done on the server. People can add the data using the UI, LDAP, the APIs directly, maybe terraform in the future...

stonith404:
Thanks, that's interesting. I didn't know this could be an issue. I have three questions:
-  Couldn't we use reflection for normalization? I think this would make it easier to maintain these models, even though it might give us a slight performance penalty.
-  I think it would make more sense to create a custom `ShouldBindJSON` function that calls the original `ShouldBindJSON` and also the `normalize` function. The call to `normalize` could easily be overlooked if we add new endpoints.
-  Maybe we can also trim all the inputs if we are doing normalization anyway?

ItalyPaleAle:
I think those are good ideas, I'll update

kmendell:
@ItalyPaleAle Mark this as ready once you get all of that updated :)

ItalyPaleAle:
Changes made!

kmendell:
Ill try to review this later today, but will probably wait to merge till elias gets back so he can look over it too.

kmendell (approved):
This looks good to me, I'd like to hold off on merging to Elias gets back though just to make sure he agrees.

Issue #544, opened 2025-05-18
🚀 Feature: Add ability to require re-authentication for specific clients


RealOrangeOne:
### Feature description

When authenticating with a service, a user should need to explicitly click "sign in", rather than automatically being authenticated and redirected. This can be sort of implemented by setting a very small `$SESSION_DURATION`, however that has other UX downsides, notably when managing a Pocket-ID account itself.

Explicit consent gives the user more control over the services they log in with, but can also increase security for their accounts.

### Pitch

With an active session, users logging in to a service are automatically authenticated and redirected back to the service. This could cause unintended logins, potentially used as part of tricking a user into authentication a session for someone else.

Authentik has the ability to set a given OIDC client as requiring explicit or implicit consent, which would work nicely here. I think for compatibility the "explicit" variant should be opt-in (implicit by default).

Additionally, given the relative ease of authenticating with a passkey, perhaps the "explicit" variant should not only require user interaction, but also re-authentication with the passkey (perhaps  3rd state of "explicit-with-auth").

stonith404:
Yes, I think we can combine this with re-authentication. 

I suggest adding a checkbox to the OIDC client called "Requires re-authentication." If enabled, users must authenticate with their passkey before being redirected, regardless of whether they are currently signed in to Pocket ID.

barryp:
I wonder if Pocket-ID's own web UI should have the option to require re-authentication - so that if a user logs into some other website (setup as an OIDC Client) using Pocket-ID, their browser can't then be used to get into the Pocket-ID management interface without some consent or interaction.

PR #747, opened 2025-07-11
feat: adds option to OIDC client to require reauthentication

MorrisMorrison:
Hi everyone,
this PR adds an option for OIDC Clients to enforce reauthentication.

Changes:
- adds db migrations for postgres and sqlite to add requires_reauthentication field to oidc_clients
- updates oidc dtos and models with RequiresReauthentication field
- updates oidc_service.HasAuthorizedClient to evaluate if client always requires reauthentication
- updates authorization page to check and enforce reauthentication
- adds form controls in OIDC client settings
- adds translations
- adds e2e test to verify reauthentication is triggered and the user can succesfully sign in afterwards

Fixes: #544

## Summary by CodeRabbit

- New Features
  - Per‑client “Requires Re‑Authentication” option for OIDC.
  - Authorization flow supports WebAuthn reauthentication with token handling.
  - New endpoint to perform passkey reauthentication.
- UI/Style
  - Admin OIDC client form and details show the new setting; widened labels for readability.
- Localization
  - Added English and German strings for the new setting and its description.
- Tests
  - Added end‑to‑end test verifying enforced reauthentication.
- Chores
  - Database migrations for reauthentication flag and tokens; scheduled cleanup of expired tokens.


kmendell:
once ready, feel free to mark as ready for review and ill take a look at it.

kmendell:
This looks good to me overall, I would like to wait till elias gets back before approving/merging though. To make sure he agrees as well.

backend/internal/service/oidc_service.go
```diff
@@ -187,6 +187,16 @@ func (s *OidcService) Authorize(ctx context.Context, input dto.AuthorizeOidcClie
 
 // HasAuthorizedClient checks if the user has already authorized the client with the given scope
 func (s *OidcService) HasAuthorizedClient(ctx context.Context, clientID, userID, scope string) (bool, error) {
+	var client model.OidcClient
+	err := s.db.WithContext(ctx).First(&client, "id = ?", clientID).Error
```
ItalyPaleAle: Please start a transaction here, and execute this query in the transaction, as well as passing the transaction to `hasAuthorizedClientInternal`

backend/resources/migrations/sqlite/20250710011023_requires_reauthentication_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE oidc_clients ADD COLUMN requires_reauthentication BOOLEAN DEFAULT FALSE;
```
ItalyPaleAle:
```suggestion
ALTER TABLE oidc_clients ADD COLUMN requires_reauthentication BOOLEAN NOT NULL DEFAULT FALSE;
```

backend/resources/migrations/postgres/20250710011023_requires_reauthentication_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE oidc_clients ADD COLUMN requires_reauthentication BOOLEAN DEFAULT FALSE;
```
ItalyPaleAle:
```suggestion
ALTER TABLE oidc_clients ADD COLUMN requires_reauthentication BOOLEAN NOT NULL DEFAULT FALSE;
```

frontend/src/routes/authorize/+page.svelte
```diff
@@ -38,8 +38,8 @@
 	async function authorize() {
 		isLoading = true;
 		try {
-			// Get access token if not signed in
-			if (!$userStore?.id) {
+			// Get access token if not signed in OR if client requires re-authentication
+			if (!$userStore?.id || client?.requiresReauthentication === true) {
```
ItalyPaleAle: Don't think we need a strictly-typed check here
```suggestion
			if (!$userStore?.id || client?.requiresReauthentication) {
```

backend/internal/service/oidc_service.go
MorrisMorrison: Thanks for your review! What is the purpose of wrapping both reads in a transaction here? To be safe against potential updates in between both read operations?

ItalyPaleAle: Yes that's one reason indeed.

Another reason is that when not in an explicit transaction, Gorm wraps each query in its own transaction [automatically](https://gorm.io/docs/transactions.html#Disable-Default-Transaction), so it will use 2 transactions for this, locking the DB twice (and performing 2 additional round-trips each time)

MorrisMorrison:
Added the requested changes. Please have a look.

backend/internal/service/oidc_service.go
```diff
@@ -187,7 +187,22 @@ func (s *OidcService) Authorize(ctx context.Context, input dto.AuthorizeOidcClie
...
+		tx.Rollback()
+	}()
+
+	var client model.OidcClient
+	err := tx.WithContext(ctx).First(&client, "id = ?", clientID).Error
+	if err != nil {
+		return false, err
+	}
+
+	if client.RequiresReauthentication {
```
kmendell: This shows the scopes again, im not sure if this was you intended behavior or not, but id think we just want the users to reauthenciate with the passkey again not completely reauthorize the scopes of the application?

frontend/src/routes/authorize/+page.svelte
```diff
@@ -38,8 +38,8 @@
 	async function authorize() {
 		isLoading = true;
 		try {
-			// Get access token if not signed in
-			if (!$userStore?.id) {
+			// Get access token if not signed in OR if client requires re-authentication
+			if (!$userStore?.id || client?.requiresReauthentication) {
```
kmendell: With my previous comment this would then do the intended behavior, Prompt the user to authenticate with the passkey again, but not have to reaccept the scopes of the application. 

If i understood the issue right, we just want the use to explicitly click the sign in button again, we should still only prompt for a passkey once.

kmendell (changes requested):

backend/internal/dto/oidc_dto.go
```diff
@@ -31,6 +33,7 @@ type OidcClientCreateDto struct {
 	LogoutCallbackURLs []string                 `json:"logoutCallbackURLs"`
 	IsPublic           bool                     `json:"isPublic"`
 	PkceEnabled        bool                     `json:"pkceEnabled"`
+  RequiresReauthentication bool               `json:"requiresReauthentication"`
```
ItalyPaleAle: Please run gofmt (you can also configure your IDE to format Go files automatically on save)

backend/internal/service/oidc_service.go
ItalyPaleAle: Indeed. I don't know if full re-auth with passkey is required, or just forcing the user to press a button? That would still solve the problem of enforcing user interaction. WDYT?

frontend/src/routes/authorize/+page.svelte
bluewalk: The issue states:
> Additionally, given the relative ease of authenticating with a passkey, perhaps the "explicit" variant should not only require user interaction, but also re-authentication with the passkey (perhaps 3rd state of "explicit-with-auth").

I'd say make sure you have to re-auth with the passkey again as clicking a simple button to consent doesn't add value/security.

backend/internal/dto/oidc_dto.go
MorrisMorrison: got added by [merge](https://github.com/pocket-id/pocket-id/pull/747/commits/c6173c4dec0f3aa537331c130011bb3a5f604a0c), but I will make sure to format again.

frontend/src/routes/authorize/+page.svelte
MorrisMorrison: agree with bluewalk, repo owner also mentioned explicitly having to re-auth with passkey in the linked issue. just lmk your decision @kmendell and I will have a look again.

kmendell: Yeah you should have to reauthenticate once, but right now it prompts you to use the passkey twice, once before and after re-accetping the scopes. removing authoirzation to the client is redundant, we should just force the user to reauth with a passkey, but not make them re accept the scopes of everything. If that makes any sense, my brain is not awake quite yet today lol

MorrisMorrison: yes having to accept the scopes again wasnt intended tbh. just wanted to know if user should only press sign in again or re-auth with passkey. thanks for checking, will update it.

ItalyPaleAle: I think re-accepting the scopes is confusing. All other IdPs I'm familiar with only show the scopes again if they have changed. Showing it every time would make it hard to know if they have changed

MorrisMorrison:
added the requested changes, please have a look:
- reverted changes in oidc_service, so users don't have to accept scopes again
- updated e2e test to verify reauthentication occurs by intercepting `/api/webauthn/login/start` call + confirming users don't have to accept scopes again

kmendell:
Ill try to take a look today!

kmendell:
@MorrisMorrison This looks much better, and i think is the intended behavior. That we are going for. Ill do a more in depth review ina bit to make sure everything looks good.

frontend/src/routes/authorize/+page.svelte
```diff
@@ -38,8 +38,8 @@
 	async function authorize() {
 		isLoading = true;
 		try {
-			// Get access token if not signed in
-			if (!$userStore?.id) {
+			// Get access token if not signed in OR if client requires re-authentication
```
ItalyPaleAle: With the latest changes, the "requiresReauthentication" check is only enforced on the client-side. I think this really needs to be enforced on the server side too, or it can be bypassed more easily than it should be.

stonith404 (changes requested):
Thanks for your PR!

As @ItalyPaleAle has stated currently the re-authentication only happens on client side. The backend would have to make sure that the user has really re-authenticated itself. 

One solution would be create an re-authentication endpoint which returns a very short-lived token similar to then authentication endpoint. The short-lived token then has to be sent with the client authorization request.

backend/internal/controller/oidc_controller.go
```diff
@@ -123,6 +124,33 @@ func (oc *OidcController) authorizationConfirmationRequiredHandler(c *gin.Contex
...
+// reauthenticateHandler godoc
+// @Summary Create reauthentication token
+// @Description Generate a short-lived reauthentication token for clients that require fresh authentication before authorization
+// @Tags OIDC
+// @Accept json
+// @Produce json
+// @Param request body object{clientID=string} true "Client ID requiring reauthentication"
+// @Success 200 {object} object "{ \"reauthenticationToken\": \"string\" }"
+// @Router /api/oidc/reauthenticate [post]
+func (oc *OidcController) reauthenticateHandler(c *gin.Context) {
```
stonith404: I would move this handler to the `webauthn_controller.go` because we might want to use this function in the future for other purposes, for example for confirming a sensitive action. If we make it more generic, the client ID parameter is not necessary.

backend/internal/dto/oidc_dto.go
```diff
@@ -3,19 +3,21 @@ package dto
...
-	CallbackURLs       []string                 `json:"callbackURLs"`
-	LogoutCallbackURLs []string                 `json:"logoutCallbackURLs"`
-	IsPublic           bool                     `json:"isPublic"`
-	PkceEnabled        bool                     `json:"pkceEnabled"`
-	Credentials        OidcClientCredentialsDto `json:"credentials"`
+	CallbackURLs             []string                 `json:"callbackURLs"`
+	LogoutCallbackURLs       []string                 `json:"logoutCallbackURLs"`
+	IsPublic                 bool                     `json:"isPublic"`
+	PkceEnabled              bool                     `json:"pkceEnabled"`
+	RequiresReauthentication bool                     `json:"requiresReauthentication"`
```
stonith404: `RequiresReauthentication` is not necessary here because `OidcClientMetaDataDto` is embedded into this struct.

backend/internal/model/oidc.go
```diff
@@ -146,3 +147,15 @@ type OidcDeviceCode struct {
 	ClientID string
 	Client   OidcClient
 }
+
+type OidcReauthenticationToken struct {
+	Base
+	Token     string
+	ExpiresAt datatype.DateTime
+	Used      bool
```
stonith404: I would prefer to delete the token instead of setting the `Used` flag.

backend/internal/service/oidc_service.go
```diff
@@ -1757,3 +1768,42 @@ func (s *OidcService) IsClientAccessibleToUser(ctx context.Context, clientID str
 
 	return s.IsUserGroupAllowedToAuthorize(user, client), nil
 }
+
+func (s *OidcService) CreateReauthenticationToken(ctx context.Context, userID string, clientID string) (string, error) {
```
stonith404: Also same as with the controller here, I would put  `CreateReauthenticationToken` and 
`ValidateReauthenticationToken` function into the `webauth_service.go`.

backend/resources/migrations/postgres/20250810012030_oidc_reauthentication_tokens.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE oidc_reauthentication_tokens (
+    id TEXT PRIMARY KEY,
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    token TEXT NOT NULL UNIQUE,
+    expires_at TIMESTAMPTZ NOT NULL,
+    used BOOLEAN NOT NULL DEFAULT FALSE,
```
stonith404: `used` can be removed, see my comment above. Additionally, please change the timestamp to the current date because we created a new release after the date of this timestamp, which means that the migration doesn't get applied.

backend/resources/migrations/postgres/20250710011023_requires_reauthentication_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE oidc_clients ADD COLUMN requires_reauthentication BOOLEAN NOT NULL DEFAULT FALSE;
```
stonith404: You can merge the `requires_reauthentication_column` and `oidc_reauthentication_tokens` migrations into a single migration.

backend/internal/service/oidc_service.go
stonith404: This function doesn't check if a passkey was used for re-authentication. An attacker could easily bypass the re-authentication by calling this endpoint with curl and receive a re-authentication token.

You would need to solve it similarly to the `VerifyLogin` function.

stonith404 (changes requested):
Thanks for your changes, I've left some comments.

backend/internal/service/oidc_service.go
```diff
@@ -1757,3 +1768,42 @@ func (s *OidcService) IsClientAccessibleToUser(ctx context.Context, clientID str
...
+
+	err = s.db.WithContext(ctx).Create(&reauthToken).Error
+	if err != nil {
+		return "", err
+	}
+
+	return token, nil
+}
+
+func (s *OidcService) ValidateReauthenticationToken(ctx context.Context, tx *gorm.DB, token string, userID string, clientID string) error {
```
ItalyPaleAle: nit: this method is only used internally so can be called `validateReauthenticationToken`

backend/resources/migrations/postgres/20250810012030_oidc_reauthentication_tokens.up.sql
```diff
@@ -0,0 +1,12 @@
...
+    user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
ItalyPaleAle: Nit: can probably remove the FK here. If the user is deleted, this becomes unusable, and eventually it gets garbage-collected. FKs make the schema more complex and operations like deleting users slower.

MorrisMorrison:
Hi everyone, thanks for the review!
I added the suggested changes, please have a look.

- moved the reauth endpoint to /webauthn
- moved the logic to webauthn_service
- removed the obsolete RequiresReauthentication flag from oidc_dto
- we directly delete the reauthentication token when consumed, instead of handling it via `used` flag
- renamed `OIDCReauthenticationToken` to just `ReauthenticationToken`
- merge db migrations into one
- updated timestamps of db migrations
- removed obsolete fk on users in migrations
- extended the reauthentication logic to verify passkey was used for reauth, similar to `VerifyLogin`

kmendell:
@coderabbitai review

backend/internal/controller/webauthn_controller.go
```diff
@@ -25,6 +25,8 @@ func NewWebauthnController(group *gin.RouterGroup, authMiddleware *middleware.Au
 
 	group.POST("/webauthn/logout", authMiddleware.WithAdminNotRequired().Add(), wc.logoutHandler)
 
+	group.POST("/webauthn/reauthenticate", authMiddleware.WithAdminNotRequired().Add(), wc.reauthenticateHandler)
+
```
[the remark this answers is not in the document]
kmendell: @stonith404 @MorrisMorrison I think this is a valid point no? I could be wrong.

backend/internal/service/webauthn_service.go
```diff
@@ -334,3 +334,91 @@ func (s *WebAuthnService) UpdateCredential(ctx context.Context, userID, credenti
...
+
+	var storedSession model.WebauthnSession
+	err := tx.
+		WithContext(ctx).
+		Where("id = ? AND expires_at > ?", sessionID, datatype.DateTime(time.Now())).
+		First(&storedSession).
+		Error
+	if err != nil {
+		return "", &common.ReauthenticationRequiredError{}
+	}
```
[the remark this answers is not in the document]
kmendell: I agree with this statment as well, since we do do the same thing for registration. This would keep it consistent

backend/internal/model/webauthn.go
```diff
@@ -45,6 +45,15 @@ type PublicKeyCredentialRequestOptions struct {
...
 }
 
+type ReauthenticationToken struct {
+	Base
+	Token     string
+	ExpiresAt datatype.DateTime
+
+	UserID string
+	User   User
+}
```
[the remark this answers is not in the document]
ItalyPaleAle: I think this is a valid point too

backend/internal/service/webauthn_service.go
```diff
@@ -334,3 +334,91 @@
...
+
+	if user.Disabled {
+		return "", &common.UserDisabledError{}
+	}
+
+	token, err := s.createReauthenticationToken(ctx, tx, userID)
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
ItalyPaleAle: I agree with this too!

backend/resources/migrations/postgres/20250814121300_requires_reauthentication.up.sql
```diff
@@ -0,0 +1,12 @@
...
+CREATE TABLE reauthentication_tokens (
+    id TEXT PRIMARY KEY,
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    token TEXT NOT NULL UNIQUE,
+    expires_at TIMESTAMPTZ NOT NULL,
+    user_id TEXT NOT NULL
+);
+
+CREATE INDEX idx_reauthentication_tokens_token ON reauthentication_tokens(token);
```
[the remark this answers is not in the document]
ItalyPaleAle: This is what we had removed before :D I don't think we need this, we have a GC

stonith404:
@MorrisMorrison Thanks for your changes. You can let me know when I should review the PR again. Please mark the comments that you've addressed as "resolved", this makes it easier to check when we should review the PR again.

MorrisMorrison:
@stonith404 Hi, I didnt add all the suggestions yet. marked the ones that are done as resolved. will let you know once ready

stonith404:
@MorrisMorrison Oh alright, sorry. Take your time.

MorrisMorrison:
@stonith404 Oh no worries at all! I’ve added the remaining suggestions. Thanks for checking!

stonith404 (approved):
I had to adapt the re-authentication logic, because I've fixed a potential vulnerability on the main branch in this commit fe003b927ce7772692439992860c804de89ce424.

Looks good to me now, thanks!

kmendell (approved):
Looks Good to me as well :)

MorrisMorrison:
ok great, thanks guys !

backend/internal/controller/webauthn_controller.go
```diff
@@ -25,6 +25,8 @@ func NewWebauthnController(group *gin.RouterGroup, authMiddleware *middleware.Au
 
 	group.POST("/webauthn/logout", authMiddleware.WithAdminNotRequired().Add(), wc.logoutHandler)
 
+	group.POST("/webauthn/reauthenticate", authMiddleware.WithAdminNotRequired().Add(), rateLimitMiddleware.Add(rate.Every(10*time.Second), 5), wc.reauthenticateHandler)
```
ItalyPaleAle, line 28: Shouldn't this re-use the same instance of the rate limit middleware used by the other routes? Otherwise, each instance would maintain its own rate-limit, which both allows more requests than what oen may expect, and also will use a lot more memory

backend/resources/migrations/postgres/20250814121300_requires_reauthentication.up.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE oidc_clients ADD COLUMN requires_reauthentication BOOLEAN NOT NULL DEFAULT FALSE;
+
+CREATE TABLE reauthentication_tokens (
+    id TEXT PRIMARY KEY,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
ItalyPaleAle, line 5: Do we need to store the time zone? The common practice is to always store times in the DB in UTC, and then format in the appropriate time zone when displaying

ItalyPaleAle (commented):
A couple of non-blocking comments from me!

backend/resources/migrations/postgres/20250814121300_requires_reauthentication.up.sql
stonith404: We are using `TIMESTAMPTZ ` for all timestamp columns in Postgres because of that I would keep it for the moment. But I think it would make sense to refactor this because in Sqlite we are not storing the timezone but we use just the unix timestamp.

backend/internal/controller/webauthn_controller.go
stonith404: For most of the routes we are using the global rate limiter but for a few critical endpoints we create a new instance with a custom limit. IMO it makes sense to keep the rate limit independent for the routes where we set a custom limit because else the limit could get unpredictable.

Issue #753, opened 2025-07-15
🚀 Feature: Api to get a user by username


elee1766:
### Feature description

I would like an api endpoint to be able to get a user given a username exists. 

### Pitch

When developing an application that uses pocket-id, it may be advantageous to be able to get the user id, or even the user information, of a user given a username.

The username is already a unique field in the table - so it should be no more trouble than user by id.

this would be useful in applications that want to store things and control access by user_id as an underlying key, but to be able to reference that key through their username.  This is useful because this means that a users storage, if keyed by their id, would stay with them seamlessly through migration, and also that there would be no issue where someone could access someone elses stuff when a username changed. 

note that it's already possible to see if a username already exists through trying to create a user with the username.

elee1766:
fwiw, my use case for this is: 

i would like to store documents for each user,  i would like to store them by by their user_id so that i dont need to move their documents if they change their username, but i would like to access these documents by their username

but to do this, i have to store my own mapping of username <-> user_id. 

it would be really helpful therefore instead of trying to sync this data in my local database to just hit pocket id, to get the user_id, and also would allow me to check sub against user_id for authenticated requests.

kmendell:
Instead of a separate endpoint for this, you can use the existing `/api/users/` endpoint which allows the client to filter by username by providing the "search" query param. Let me know if you would like an example. Otherwise you can refer to the API Documentation on how to do this.

PR #754, opened 2025-07-15
feat: get user by username via api

kmendell:
Modifies the existing /api/users/{id} endpoint to be able to get users by their `username` or their `id`

<details>
<summary>Issues Closed by this PR</summary>

Fixes: https://github.com/pocket-id/pocket-id/issues/753

</details>

### DIFF
--- a/backend/internal/service/user_service.go
+++ b/backend/internal/service/user_service.go
@@ -55,16 +55,20 @@ func (s *UserService) ListUsers(ctx context.Context, searchTerm string, sortedPa
 }
 
 func (s *UserService) GetUser(ctx context.Context, userID string) (model.User, error) {
-	return s.getUserInternal(ctx, userID, s.db)
+	return s.getUserInternal(ctx, "id", userID, s.db)
 }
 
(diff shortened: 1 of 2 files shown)


backend/internal/service/user_service.go
```diff
@@ -55,16 +55,20 @@ func (s *UserService) ListUsers(ctx context.Context, searchTerm string, sortedPa
...
+}
+
+func (s *UserService) getUserInternal(ctx context.Context, key string, userID string, tx *gorm.DB) (model.User, error) {
 	var user model.User
 	err := tx.
 		WithContext(ctx).
 		Preload("UserGroups").
 		Preload("CustomClaims").
-		Where("id = ?", userID).
+		Where(key+" = ?", userID).
```
ItalyPaleAle, line 71: The problem with the approach above is that you're assuming that everything that looks like a UUID, is an ID, and not a username. This may not always be true, for example in the case of users imported from LDAP you may have users whose username is a UUID-like.

You could probably simplify the entire logic by changing this query to:

```suggestion
		Where("id = ? OR username = ?", userID, userID).
```

(And maybe a UNIQUE index on the table so that there can't be a username or ID that are equivalent, just in case...)

Issue #294, opened 2025-03-03
🚀 Feature: Log levels


mihakrumpestar:
### Feature description

It would be great if we could set container log level as currently GIN logs every single request, which in my home-lab produces almost 7,000 log entries per hour.

### Pitch

I always set log level to warning, so I can clearly see if anything is wrong, as finding an error entry among 1000-ts of logs is a lot harder.

kmendell:
@mihakrumpestar  What is your `APP_ENV` variable set too? It should cut down on logging if its set to: `APP_ENV=production`

mihakrumpestar:
It is unset, as I checked the code where "production" is the default.

mihakrumpestar:
Well, GIN logs are completely irrelevant, and produce a ton of output.

I would appreciate a log level `warning`, where I can see if any problems arise, without any irrelevant info data like successful requests.

kmendell:
I dont see a default way to do this in GIN, unless you use the `Skipper` function, so if implemented, it would a be a bit of work. @stonith404  what are your thoughts?

stonith404:
@kmendell It seems like `gin.New()` creates a logging middleware. For supporting log levels we would have to create a custom logger.

I think this would make sense to implement but it's definitely not at the top of the priority list. But we are open for pull requests or upvotes (👍) which would prioritize the issue more.

kklee998:
Seems like the default Gin logger does not support logging levels. How about using this logging middleware from `gin-contrib`: 



Here is a rough sketch of the code:
```go
// backend/internal/bootstrap/router_bootstrap.go

import (
        ....
	"github.com/gin-contrib/logger"
        ....
)


func initRouterInternal(db *gorm.DB, svc *services) (utils.Service, error) {
       ....
	r := gin.New()
	r.Use(gin.Recovery())
        // TODO: get the LOG_LEVEL from environment and call .SetLogger() with the correct config
	r.Use(logger.SetLogger()) 
       ....
}
```

Logs look like this:

![Image](https://github.com/user-attachments/assets/02ba4d84-6b54-47a4-b939-d02a5293a34c)

If you guys are okay with taking on an extra dependency, I can open a PR to properly implement this

PR #760, opened 2025-07-20
feat: Support OTel and JSON for logs (via log/slog)

ItalyPaleAle:
This PR adds support for sending logs to a collector using OpenTelemetry (OTel). Additionally, it allows printing log lines as JSON.

## OTel for logs

We added support for OTel for traces and metrics with #495. The missing piece for observability was logs, which did not support OTel yet.

- This PR adds the ability to send logs to an OTel collector, configured using the typical env vars for OpenTelemetry (such as `OTEL_LOGS_EXPORTER=otlp` and `OTEL_EXPORTER_OTLP_ENDPOINT`).
- In order to support OTel, this PR adds a log exporter to the `log/slog` package, and configures slog as the default logger. 
- By making slog the default log handler, all log calls that use the "old" `log. Print` functions are **automatically converted to slog calls** (and sent to an OTel collector, if configured)
- Both gorm and gin required using wrappers to make them print logs that work well with slog and can be sent to OTel

> **Note:** one of the most useful features of slog and OTel is the ability to have logs automatically correlated to traces. When we use `log.Print`, the correlation isn't sent to OTel, so logs aren't correlated.
> In order for logs to be correlated, we will need to convert all log calls to use slog and receive the context (at least in methods that have a trace, such as those that are invoked by a HTTP request).
> I have shown an example here: 
> https://github.com/pocket-id/pocket-id/pull/760/files#diff-21ae94b87213fba831ae50c5ff9a5c85c28396c549269cfd1e33a0561c158298R260
> I have **not** converted other log calls, so those will need to be done in subsequent PRs.

## Log as JSON

This was an easy add since we were implementing slog support.

You can now set `LOG_JSON=true` and logs are printed out as JSON! JSON-formatted logs are often preferred when logs are sent to tools that parse them.

JSON logging is disabled by default.

> When the Pocket ID binary runs in a terminal (such as the user manually running `pocket-id` in the shell), logs are colored; otherwise, such as when running in a container or piping logs to a file, colors are disabled.


kmendell:
I'll try to look over this tomorrow, my first glance it seemed okay though.

backend/internal/bootstrap/bootstrap.go
```diff
@@ -14,13 +14,17 @@ import (
...
-	initApplicationImages()
-
-	// Initialize the tracer and metrics exporter
-	shutdownFns, httpClient, err := initOtel(ctx, common.EnvConfig.MetricsEnabled, common.EnvConfig.TracingEnabled)
+	// Initialize the observability stack, including the logger, distributed tracing, and metrics
+	shutdownFns, httpClient, err := initObservability(ctx, common.EnvConfig.MetricsEnabled, common.EnvConfig.TracingEnabled)
 	if err != nil {
 		return fmt.Errorf("failed to initialize OpenTelemetry: %w", err)
 	}
+	slog.InfoContext(ctx, "Pocket ID is starting")
```
kmendell, line 22: Sort of a nit pick, but since we have the starting text, should we add the version to it too? Just thinking it could be useful in the event of a denug situation.

ItalyPaleAle: It's definitely useful, but the version is added automatically to all log lines: 

https://github.com/pocket-id/pocket-id/pull/760/files#diff-93738a6cb75e9cfa4258f38081bdda1e190438d2ef5ea968a45ad0fb7b976eceR135-R137

kmendell:
Also can you make a PR on the `preview` branch of the website adding the new ENV Variable you added as well please?, I can do it later if you dont get around to it.

ItalyPaleAle:
>  dont have way to test the OTEL stuff though at the current moment,

Here's an example of logs in Grafana Cloud:

image

App is configured with these env vars:

```
OTEL_LOGS_EXPORTER=otlp
TRACING_ENABLED=true
METRICS_ENABLED=true
OTEL_SERVICE_NAME=pocket-id
OTEL_RESOURCE_ATTRIBUTES=deployment.environment=dev,service.version=dev
OTEL_EXPORTER_OTLP_ENDPOINT=https://otlp-gateway-prod-us-east-0.grafana.net/otlp
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic ..."
```


> So if i understand right we dont need to chnage all of the other fmt.Println statments? Should we change them though to have a consistancy in the codebase?

I assume you mean the `log. Print` (and `log.Println` and `log.Printf`).

- It is **not required** to change them, because they are automatically converted into `slog. Info` calls
- ... however we may want to change them (over time) for:
   - Ability to have the logs correlated to traces. Logs sent by OTel have a "trace_id" if there's a trace, so logs can be correlated to each request
      - For a trace to exist, the logs must use `slog.InfoContext` (or `slog.WarnContext`, `slog.ErrorContext`, `slog.DebugContext`...) and must have a context passed that contains tracing info (contexts that come from the HTTP server do have the trace attributes)
   - Ability to use structured logging. As an example, see how I've changed `log.Printf("Attempt %d: Failed to initialize database. Retrying...", i)` into `slog.Info("Failed to initialize database", slog.Int("attempt", i))`
   - Consistency in the codebase, as you mention

backend/internal/bootstrap/bootstrap.go
kmendell: Sure, what if the app doesnt get that far though? Im thinking outloud honestly, just a idea that popped into my head.

ItalyPaleAle: It's the first thing that is executed, so it should always be present!

kmendell: Oh i see what your saying now, man my brain is not working today 😅

kmendell:
It looks good to me, Just waiting for all feature merges till elias gets back though.

kmendell:
@ItalyPaleAle Im tempted to say to just migrate everything to slog now since we have the time. That was we can just have the consistency done and one less thing to worry about. 

Edit: idk how i didn't notice the wall of text I had typed before my bad :D

stonith404:
I agree with @kmendell, this shouldn't be hard to migrate, right?

Additionally, we might want to implement log levels #294 with this PR. What do you think @ItalyPaleAle?

ItalyPaleAle:
> I agree with @kmendell, this shouldn't be hard to migrate, right?

I spoke with @kmendell yesterday. If that's ok with you, I'd prefer to do that in a follow-up PR. This way, this PR will be smaller  (it's already fairly large) and easier to review and merge as the "foundation". I am concerned that changing _all_ log statements here will make this PR harder to review and meanwhile it will be easier to get never-ending merge conflicts.

> we might want to implement log levels https://github.com/pocket-id/pocket-id/issues/294 with this PR.

I think this PR will provide the foundation for that, but there's a few other things that need to be done.

- Using slog we have the ability to control the log level. You can see in a couple of the sample logs I've already converted that I've used `slog.Warn` and `slog. Error`. Part of the refactoring of all `log. Print` statements will also have to be deciding the level, but at this stage there's nothing else "foundational" that is needed here
- The second aspect is about controlling the log level. For that we'd need to implement a switch in the config to determine what level to use. This is not implemented here yet, but can be done in a follow-up relatively easily.
- In #294 they are specifically asking about Gin logs. Currently this PR does change the Gin logger to one that uses slog, but it uses the default INFO level. If we want to support a different log level for Gin, we'd probably need a separate config option.

kmendell:
@ItalyPaleAle @stonith404 While this looks good to me, ill merge it, we can then start the building blocks of expanding to migrating the rest of the logging over. I throughly tested this and could nto get it to break so we should be fine :D

kmendell (approved):
Approving, so we can then start on migrating the rest of the logging over.

PR #774, opened 2025-07-23
feat: add apple-touch-icon support

James18232:
added apple-touch-icon support for ios devices.

used the dark logo, because it is better.


savely-krasovsky:
180x180 for Apple icon seems like a very low resolution. AFAIK 512x512 is the minimum for the recent iPhones, iPads could require even more (1024x1024).

savely-krasovsky:
Also in 2025 `apple-touch-icon` is basically deprecated. Modern browsers including Safari use json manifest: https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/icons

kmendell:
I think the json manifest would work, i didnt know apple-touch-icon was deprecated though. 

Either way we should use the api url for getting the icon i think, as that way were not adding a static asset see this line: https://github.com/pocket-id/pocket-id/blob/2c78bd1b46f3fed949b9f27aad23371db40323d9/frontend/src/app.html#L5

ItalyPaleAle:
I would make it an actual static asset. This is the same manifest that would be used to configure a PWA, so it should be generated by the Svelte Kit builder.

kmendell:
But it should use the Logo that the user has defiend in their settings no? adding a static assets would not do this.

ItalyPaleAle:
Sorry, I misread your comment.

Agree, the logo should come from `/api/application-configuration/favicon` !

Just the manifest itself should be statically-generated (by Svelte Kit)

kmendell:
the favicon isnt the right logo url but it was the first one i could grab at ease. It would have to use the /api/application-configuration/logo. This would support both light and dark technically, not sure if thats suported, but i need to readup more on that, ive never really messed with PWA apps before.

kmendell:
@James18232 Can you try my latest commit i think i implemented the manifest support in a basic way, but im not in a place where i can fully test it.

James18232:
This does not work.

it may need a fixed asset in the manifest (png file) and/or a purpose of any..

 AFAIK  manifest use is only supported on ios for PWA’s, not apple touch icons which still rely separately on a link in the head per my commit or a apple-touch-icon.png at the root - could be wrong. 
 
Priority is the link tag in a page then the root. Not familiar if this can be fetched from /api instead, but it needs a fixed png asset. Not sure this would support light or dark dynamically.

you could load it dynamically based on the api i guess and just use local storage to avoid fetching all the time…i believe it is just grabbed at whatever asset is in the link tag at the time.

This feature is just to set an icon on the homescreen when a user clicks share —> add to homescreen on ios devices (not treated as a PWA - effectively just a bookmark icon)

as far as size AFAIK the current max size per specs is still 180px. There are smaller specs for older ios devices but not relevant to support imo.

kmendell:
Okay i reverted those changes, ill look more into if the api route can be used, we do something simalr for caching the icon elsewhere so it should be possible.

ItalyPaleAle:
> AFAIK manifest use is only supported on ios for PWA’s, not apple touch icons

Isn't that the same situation? Adding a website to the home screen should be equivalent to installing it as PWA if available, or adding as a bookmark if not.

Please correct me if I'm wrong here.

savely-krasovsky:
It's a pity that Apple doesn't have any modern documentation on this.

I only found this on MDN:
> Safari 17 (Release date: 2023-09-18)
> Only used when no apple-touch-icon is present and either "purpose" is set to "any" or "purpose" is not specified.

James18232:
I have had a mess around and think i have it working now. had to move and reuse the logo component logic so it can dynamically set a manifest file. it now registers (i think) as a PWA and hides the url bar as a basic enhancement. 

tested on my end, but i cant seem to make it grab the regular logo, it always seems to grab the dark logo..

frontend/static/manifest.json
ItalyPaleAle, line 1: Why this file if you're generating it dynamically?

frontend/src/routes/manifest.json.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function GET({ request }: RequestEvent) {
+  const logoUrl = getLogoUrl();
+  const manifest = {
+    name: "PocketID",
+    icons: [
+      {
+        src: logoUrl
+      }
+    ],
+    display: "standalone",
```
ItalyPaleAle:
```suggestion
    display: "browser",
```

frontend/src/routes/manifest.json.ts
```diff
@@ -0,0 +1,23 @@
+import { getLogoUrl } from '$lib/utils/logo-util';
+import type { RequestEvent } from '@sveltejs/kit';
+
+export async function GET({ request }: RequestEvent) {
+  const logoUrl = getLogoUrl();
+  const manifest = {
+    name: "PocketID",
```
ItalyPaleAle: This should probably get the name of the app from `{$appConfigStore.appName}`

frontend/static/manifest.json
James18232: No longer needed thanks - deleted

frontend/src/routes/manifest.json.ts
James18232: Up to you guys re display mode. Standalone gives a nice pwa feel to me and i dont think any browser objects are needed to navigate or interact with. Browser will leave it as effectively just a bookmark i guess.

frontend/src/routes/manifest.json.ts
James18232: yes makes sense

frontend/src/routes/manifest.json.ts
ItalyPaleAle: I will leave it to @kmendell to have the final word, but IMHO here I'd keep "browser" since we don't have a proper PWA

kmendell:
I will take a look at this in a little bit, been busy today 😅

James18232:
Thanks. Still an issue actually. The extra static manifest i had to test was being used and hiding an issue with the dynamic manifest. Will do some further troubleshooting.

kmendell:
@James18232 I know the issue sveltekit cant render routes like that it uses + syntax, im working on it now, at leats to get it working, as this is just sort of tiedious.

kmendell:
Okay @James18232 Try that i think it should work now with the dynamic serving part, at least edge is detecting the manifest for it.. I havent added the dynamic app name yet , but this should work if it does then well handle all that.

James18232:
The manifest is being set now yes, but it is not pulling any logo from it - just the default.

ItalyPaleAle:
Note: I've pushed an update to change the extension and content-type of the manifest as per recommendation, fixing a warning from the dev tools ([see this Stack Overflow](https://stackoverflow.com/questions/52129281/manifest-json-vs-manifest-webmanifest))

ItalyPaleAle:
Ok I think I got it to work, on both Edge (Chrome) and Safari
image
image


The issue was that Safari was making requests for `/apple-touch-icon.png` and `/apple-touch-icon-precomposed.png`, which didn't exist. However, since Pocket ID's server was returning the SvelteKit app, the response was a 200 one with some HTML in the body, so not a valid image that Safari could render.

James18232:
Interesting. Works on mac os but still not on ios? Agree the issue is ios expects only a png in the src, a link to the api may not work. Hence why the static manifest and client side image work just fine. Is it possible to set the image src and also send the correct logo to the client at the same time from the server side approach? This would allow the light/dark change but fake the traditional client side manifest/asset when the browser goes looking for it (if im understanding).

ItalyPaleAle:
> Agree the issue is ios expects only a png in the src, 

Oh, that may explain it then. The default icon a SVG...

kmendell:
i think the eaisest way to accomplish this is to just use the standard pocket id icon. While its not ideal, i dont think the dynamic route will work. i need to do more research, and talk to elias about this too. Cause im not a fan of the prerendered router in all honesty as its not clean, but im not sold on it.

James18232:
Yes agree kiss. Not sure if a static manifest is preferable and then serve a regular png from the server instead of client side.. i think having a client side script to check for light/dark mode (api query to server) then replace the png src to the dark png src (also kept on the server side) could be a simple option. 

Or just slap a png client side and the  good ol rel link and send it 🫡

savely-krasovsky:
To make it properly we should accept SVG/PNG, then resize it into various sizes and cache it somewhere. Generate manifest and serve those cached logos.

ItalyPaleAle:
If we're just putting a PNG that's static, how about we use [`monochrome`](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/icons#monochrome) so then the OS can change it between light and dark automatically?

James18232:
sounds reasonable. i'll look into it further.

James18232:
i have made the main logo served properly now as a png.  

I have tried nearly all combinations, and have come to the conclusion there is no way to serve a light/dark mode pwa icons. ios just doesnt respect manifest conventions.

The closest i got was a monochrome of black with transparent background, which is fine, but when dark mode is used ios makes the background dark and its then black on black.

the dark mode also doesnt work in monochrome as ios seems to default to a white background for pwa's with no way i can see to change this. this means the icon appears totally white, but looks fine with dark mode turned on.

good enough i think..

backend/internal/controller/app_config_controller.go
```diff
@@ -35,6 +35,7 @@ func NewAppConfigController(
 	group.PUT("/application-configuration", authMiddleware.Add(), acc.updateAppConfigHandler)
 
 	group.GET("/application-configuration/logo", acc.getLogoHandler)
+	group.GET("/application-configuration/AppleIcon", acc.getAppleIconHandler)
```
savely-krasovsky: I think it violates current naming convention. `/application-configuration/apple-icon` would suite better.

backend/resources/images/AppleTouchIcon.png
ItalyPaleAle, line 1: Can you rename this so it doesn't have "Apple" in the name? Because it's not just for Apple devices, for example you can add the app on Windows, Chrome, etc.

frontend/static/manifest.webmanifest
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "PocketID",
```
ItalyPaleAle: Why did you revert back to a static file? As per my comment, we would need the `name` to come from the app's name, which is dynamic.

kmendell:
No worries on all the comments, ill fix this a bit later today

kmendell:
@James18232 See if that works, it should use the same logic but also the dynamic app name now.

ItalyPaleAle:
I was wrong, actually. The manifest needs to be static because we can't have `+server.ts` routes in SveltKit, because we build using a static adapter so there's no server to render them...

James18232:
yes the static adapter threw me too. Not familiar with sveltkit. Theoretically in future though you could run a sveltkit backend dynamically as well as the current backend? That way the +server.ts files could be much more useful..

i can also add in a transparent logo that fits within the safe area as another icon for use with android - they mask to a circle (why?????). Not sure if this will break the apple config but i think it will work if the purpose is set to just maskable and monochrome.. would needs another api route to serve it. 

May as well add the svg also.

ItalyPaleAle:
> Theoretically in future though you could run a sveltkit backend dynamically as well as the current backend

Ah, funny you say that :D Prior to 1.0, there was a separate backend running SvelteKit. We switched it in 1.0 so to have a single binary.

I have a fix incoming, and also making the icon maskable since the current one doesn't look great on the iOS home screen.

Give me a few mins! Sorry this PR is being quite a lot of work...

ItalyPaleAle:
Ok I think I got it. This is what it looks like on the iOS home screen
IMG_86CCFEDF7D8D-1

The way it works now:

- The manifest in the SvelteKit app is static. This way, we can add more fields using the Svelte build system in the future if needed
- The backend parses the manifest and replaces the `name` field when serving it

Can you please help double-checking that everything works on your end too?

James18232:
@ItalyPaleAle yes this functions for me also.

two comments:
 1. the manifest has a duplicate png icon twice (error i think)- i believe the svg will need its own entry (type image/svg+xml, size any, maskable) as well as it own api route to serve it.

2. consider an additional monochrome png (with the background removed/transparent)? for use on devices that respect the manifest better (eg android). this will enable the background to be set by the OS responding to light/dark mode. i think you could set a default background with a 'theme_color" or "background_color" in the manifest but not 100% sure on that - ios ignores these and just picks white. i think if this was a third option with a purpose of "maskable monochrome (not any)" this would ensure ios devices grab the default (purpose any) png while others could grab the specific monochrome. i can only test the ios side of it though. if ios grabs the monochrome png it looks terrible as i described above.

ItalyPaleAle:
@James18232 thanks for confirming!

> 1. the manifest has a duplicate png icon twice (error i think)- i believe the svg will need its own entry (type image/svg+xml, size any, maskable) as well as it own api route to serve it.

Not an accident. I wanted to have one entry with `maskable`, and one with `any` because Safari doesn't support anything besides `any`. See MDN:
image

> consider an additional monochrome png (with the background removed/transparent)? 

As for this, I think it could be a good idea. Wanna take a stab at implementing it? Here are the SVG and PNG with alpha that you can use 



> Tip: use https://squoosh.app to make images much smaller, including lossless PNG compression!

James18232:
Learn something new everyday.  The joys of cross platform compatibility. 

I think the svg entry is still missing in the manifest though.. 

I wasn’t going to have a stab at the monochrome - i did already for ios and determined the support just isn’t there. 

I  also dont know of a way to reliably test the implementation on android, so would be shooting from the hip so to speak. Perhaps others with android devices could so it can be tested on real hardware.

James18232:
i have added the svg. unfortunately adding an additional 'any' purpose svg breaks the ios icon support (really not sure why), so i have just added a maskable monochrome one so at least it gets served.

ItalyPaleAle (approved):

ItalyPaleAle:
LGTM

> i have added the svg. unfortunately adding an additional 'any' purpose svg breaks the ios icon support (really not sure why), so i have just added a maskable monochrome one so at least it gets served.

I don't want to get back to square one, but maybe.... we add back the apple-touch-icon image for iOS, and use the PWA manifest for all other sane browsers (i.e. not Safari)...

James18232:
I hate to admit the logic in that 😂. if enough people want it its a simple tweak. Hopefully doing nothing works and apple gets with the program in the meantime.

kmendell:
Trying to clean this up a bit, as it took a turn to be a little to complex, in my opinion. I think this approach is working, at least the icons show up on my phone and the manifest screen in edge. Though my icon doesnt look like the screenshot, idk if thats cause im on iOS 26 or what.

James18232:
horses for courses - still works/looks fine.

not sure the 301 redirect to /api/application-configuration/web-manifest is needed though? could this just replace the href directly in src/app.html?

kmendell:
It could be, this is not by any means final, or cleaned up. But i do think this is more reasonable. Feel free to tr yand clean it up, otherwise il look more when i have time.

James18232:
I prefer the previous concept from @ItalyPaleAle of a static manifest on the front end with only the name dynamically handled by the backend. 

Seems to fit better with the static site frontend/api backend logic that has been adopted. Burying the manifest in the backend just to input the app name seems strange to me.

up to you guys as maintainers though - they both work just fine 👍.

ItalyPaleAle:
I would +1 on keeping the manifest as static in the FE (but not a blocker). The reasoning for me is that if in the future we want to add other PWA-related things to the manifest, it makes more sense to have it in the FE.

kmendell:
Alright thats fine, i wasn't opposed to that route, but theres something inside me that thinks its can be done much cleaner or a different way, honestly this whole thing was way more complicated than it should be LOL. Im gonna wait till @stonith404 gets back to discuss that see what approach he likes better.

shaqagain:
Is this fix just for iPhone? I can add Pocket-ID to homescreen but can't "install" this as a PWA on Android  16. I tested on Chrome and Brave. 

Screenshot_20250827-155104

stonith404:
I'm closing this in favor of #938 which is much simpler but doesn't support the custom logo and name. Thanks for your draft @James18232, the credit goes to you.

PR #776, opened 2025-07-23
fix: set input type 'email' for email-based login

### LINKED ISSUE #770: 🐛 Bug Report: Email comparison is case sensitive
### Reproduction steps

When using the "Email login" (requesting a one-time code per email), the end-user has to type in their email, which is then searched in the database for a match, and if found, it will send the code to the corresponding email.

Now the code doesn't make a case-insensitive search, and therefore any case difference will fail.

This makes the login process quite _fragile_, inconsistent, in particular for end-users who won't be able to login on this new system (they are already lost with just the _access keys_ concept).

This is particularly true since on Android mobile, the system **will automatically change the first typed-in character of the user-input/email in Uppercase**, making the process fail in 100% of the cases.

### Expected behavior

I don't know any system that makes use of case-sensitive emails, it should not make any difference.

PocketId should make **case-insensitive** emails comparison, to ensure a robust, resilient login mechanism.

### Actual Behavior

The code makes a simple search (assuming case-sensitive):
https://github.com/pocket-id/pocket-id/blob/main/backend/internal/service/user_service.go#L358

(I also quickly searched into the frontend code, and haven't found a _tolowercase_ or any similar handling).

### Version and Environment

latest

### Log Output

_No response_

ItalyPaleAle:
Fixes #770 because inputs of type "email" do not use autocapitalization

### DIFF
--- a/frontend/src/routes/login/alternative/code/+page.svelte
+++ b/frontend/src/routes/login/alternative/code/+page.svelte
@@ -61,7 +61,7 @@
 		<p class="text-muted-foreground mt-2">{m.enter_the_code_you_received_to_sign_in()}</p>
 	{/if}
 	<form onsubmit={preventDefault(authenticate)} class="w-full max-w-[450px]">
-		<Input id="Email" class="mt-7" placeholder={m.code()} bind:value={code} type="text" />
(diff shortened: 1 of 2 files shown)


kmendell (approved):
Thanks!

PR #783, opened 2025-07-25
fix: delete WebAuthn registration session after use

ItalyPaleAle:
During the WebAuthn registration process, make sure to delete the session after it's verified and the passkey is saved. This prevents the same session from being used twice.

### DIFF
--- a/backend/internal/service/webauthn_service.go
+++ b/backend/internal/service/webauthn_service.go
@@ -9,6 +9,7 @@ import (
 	"github.com/go-webauthn/webauthn/protocol"
 	"github.com/go-webauthn/webauthn/webauthn"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 
 	"github.com/pocket-id/pocket-id/backend/internal/common"
 	"github.com/pocket-id/pocket-id/backend/internal/model"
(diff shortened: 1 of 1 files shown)


backend/internal/service/webauthn_service.go
```diff
@@ -70,8 +71,7 @@ func (s *WebAuthnService) BeginRegistration(ctx context.Context, userID string)
 		Find(&user, "id = ?", userID).
 		Error
 	if err != nil {
-		tx.Rollback()
```
ItalyPaleAle, line 73: This rollback was not needed here, since there's a `defer`'d rollback above

kmendell:
Doesn't this do exactly what the job does every 24 hours? 
https://github.com/pocket-id/pocket-id/blob/12a7a6a5c5bec4f209260b90fc06e2e2201aecdf/backend/internal/job/db_cleanup_job.go#L23C3-L23C86

ItalyPaleAle:
> Doesn't this do exactly what the job does every 24 hours? 

Yes, but that's after at most 24 hours.

If a session's been used, it should be removed from the database. Otherwise, there's the risk it could be used twice to register a second passkey.

kmendell (approved):

PR #787, opened 2025-07-27
refactor: complete conversion of log calls to slog

ItalyPaleAle:
A few tiny bugs were fixed in the process :)

### DIFF
--- a/backend/internal/dto/validations.go
+++ b/backend/internal/dto/validations.go
@@ -1,7 +1,8 @@
 package dto
 
 import (
-	"log"
+	"log/slog"
+	"os"
 	"regexp"
 
 	"github.com/gin-gonic/gin/binding"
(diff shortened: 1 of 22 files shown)


backend/internal/service/oidc_service.go
```diff
@@ -1211,16 +1223,12 @@ func (s *OidcService) VerifyDeviceCode(ctx context.Context, userCode string, use
 	deviceAuth.UserID = &userID
 	deviceAuth.IsAuthorized = true
 
-	if err := tx.WithContext(ctx).Save(&deviceAuth).Error; err != nil {
-		log.Printf("Error saving device auth: %v", err)
-		return err
-	}
-
-	// Verify the update was successful
```
ItalyPaleAle, line 1219: I can't really understand what this was for, but didn't look like it did anything, so just removed it

backend/internal/dto/validations.go
```diff
@@ -18,9 +19,16 @@ var validateUsername validator.Func = func(fl validator.FieldLevel) bool {
...
 
 func init() {
-	if v, ok := binding.Validator.Engine().(*validator.Validate); ok {
-		if err := v.RegisterValidation("username", validateUsername); err != nil {
-			log.Fatalf("Failed to register custom validation: %v", err)
-		}
+	v, ok := binding.Validator.Engine().(*validator.Validate)
+	if !ok {
+		// Indicates a development-time error
+		panic("binding.Validator.Engine is not *validator.Validate")
```
kmendell: Do we really want to panic here? i know this is probably rare, and a development issue but a panic seems extreme

ItalyPaleAle: Yeah, it indicates a development-time error. This is in an `init` method so if this fails, the developer did something wrong

kmendell: Does its segfault then? like i guess im saying wouldnt the program panic anyways lol

kmendell: its a nit pick, otherwise its looks good to me.

kmendell (approved):
Looks good besides that one comment, Fell free to merge it

backend/internal/dto/validations.go
ItalyPaleAle: Yeah It would panic anyways. I'll remove that

Issue #790, opened 2025-07-27
🚀 Feature: Include code_challenge_methods_supported in ./well-known endpoint


mzgaljic:
### Feature description

An application I use, [Outline](https://github.com/outline/outline), determines if the oauth provider supports the PKCE flow by inspecting the `./well-known` response. To be specific, it looks for key `code_challenge_methods_supported`. If it doesn't exist, it assumes PKCE is not supported. Outline is doing the right thing, [according to the spec](https://datatracker.ietf.org/doc/html/rfc8414#section-2):

> code_challenge_methods_supported
      OPTIONAL.  JSON array containing a list of Proof Key for Code
      Exchange (PKCE) [[RFC7636](https://datatracker.ietf.org/doc/html/rfc7636)] code challenge methods supported by this
      authorization server.  Code challenge method values are used in
      the "code_challenge_method" parameter defined in Section 4.3 of
      [RFC7636].  The valid code challenge method values are those
      registered in the IANA "PKCE Code Challenge Methods" registry
      [[IANA.OAuth.Parameters](https://datatracker.ietf.org/doc/html/rfc8414#ref-IANA.OAuth.Parameters)].  **If omitted, the authorization server
      does not support PKCE.**

Emphasis on `If omitted, the authorization server does not support PKCE.`

Here is how outline specifically handles this (if you're curious): https://github.com/outline/outline/pull/9478/files

----------------

### Pitch

Pocket ID does not currently include `code_challenge_methods_supported` in the `./well-known` response. It would be great if it did, so that I can use Outline (and other apps that work in a similar way) with PKCE. Since PKCE is more secure. 

If you're looking for a real example, I see google returns it in their response: https://accounts.google.com/.well-known/openid-configuration

kmendell:
@mzgaljic This should be in the ghcr.io/pocket-id/pocket-id:next` image in a few minutes, if you would like to make sure it works.

mzgaljic:
> [@mzgaljic](https://github.com/mzgaljic) This should be in the ghcr.io/pocket-id/pocket-id:next` image in a few minutes, if you would like to make sure it works.

That was fast. Just tested it out, PKCE works with [Outline](https://github.com/outline/outline) now. Thank you! Looking forward to seeing this in the next release.

PR #794, opened 2025-07-28
feat: add support for `code_challenge_methods_supported`

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/790

Added both "plain" and "S256" from here https://datatracker.ietf.org/doc/html/rfc7636#section-6.2.2

### DIFF
--- a/backend/internal/controller/well_known_controller.go
+++ b/backend/internal/controller/well_known_controller.go
@@ -87,6 +87,7 @@ func (wkc *WellKnownController) computeOIDCConfiguration() ([]byte, error) {
 		"subject_types_supported":                        []string{"public"},
 		"id_token_signing_alg_values_supported":          []string{alg.String()},
 		"authorization_response_iss_parameter_supported": true,
+		"code_challenge_methods_supported":               []string{"plain", "S256"},
 	}
 	return json.Marshal(config)
 }


backend/internal/controller/well_known_controller.go
```diff
@@ -87,6 +87,7 @@ func (wkc *WellKnownController) computeOIDCConfiguration() ([]byte, error) {
 		"subject_types_supported":                        []string{"public"},
 		"id_token_signing_alg_values_supported":          []string{alg.String()},
 		"authorization_response_iss_parameter_supported": true,
+		"code_challenge_methods_supported":               []string{"plain", "S256"},
```
ItalyPaleAle, line 90: We only support S256 in Pocket ID

https://github.com/pocket-id/pocket-id/blob/01bf31d23d9200f9c7f44efe83ba5ac0a953efeb/backend/internal/service/oidc_service.go#L1014


```suggestion
		"code_challenge_methods_supported":               []string{"S256"},
```

kmendell: https://github.com/pocket-id/pocket-id/blob/01bf31d23d9200f9c7f44efe83ba5ac0a953efeb/backend/internal/service/oidc_service.go#L1043-L1045

These are the lines of why i put plain.

ItalyPaleAle: Ahhh ok

quagsirus:
This seems to have fixed PKCE compatibility with Immich! Thank you!

PR #799, opened 2025-07-29
feat: support reading secret env vars from _FILE

### LINKED ISSUE #685: 🚀 Feature: Enable use of Docker Secrets
### Feature description

I couldn't find any reference in the documentation for using docker secrets for sensitive information (Maxmind key, smtp password). I hope I'm not missing it.

### Pitch

Most other apps I deploy with docke I can change the environment variable to XXX_FILE=<secretsname>. I think it would be a good feature to support. Ideally it would support this for all environment variables just because? but at the very least sensitive information.

MAXMIND_LICENSE_KEY_FILE=/run/secrets/<secretname>
SMTP_PASSWORD_FILE=/run/secrets/<secretname>

ItalyPaleAle:
Fixes #685

Env vars that contain secret values now can be read from file. To do that, pass the env var `*_FILE` containing the path to a file on disk. This works with Docker (and K8s) secrets too, since they are mounted as files.

These env vars are currently supported:

- EnvConfig:
   - `DB_CONNECTION_STRING_FILE`
   - `MAXMIND_LICENSE_KEY_FILE`
   - Additionally, `ENCRYPTION_KEY_FILE` was already available
- AppConfig (when UI configuration is disabled): all values that have the `sensitive` tag in the struct, which currently includes:
   - `SMTP_PASSWORD_FILE`
   - `LDAP_BIND_PASSWORD_FILE`

### DIFF
--- a/backend/internal/common/env_config.go
+++ b/backend/internal/common/env_config.go
@@ -12,7 +12,7 @@ import (
 	_ "github.com/joho/godotenv/autoload"
 )
 
-func resolveStringOrFile(directValue, filePath, varName string) (string, error) {
+func resolveStringOrFile(directValue string, filePath string, varName string, trim bool) (string, error) {
 	if directValue != "" {
 		return directValue, nil
 	}
@@ -21,7 +21,11 @@ func resolveStringOrFile(directValue, filePath, varName string) (string, error) 
 		if err != nil {
 			return "", fmt.Errorf("failed to read secret '%s' from file '%s': %w", varName, filePath, err)
 		}
-		return strings.TrimRight(string(content), " \t\n\r"), nil
+
+		if trim {
+			return strings.TrimSpace(string(content)), nil
+		}
+		return string(content), nil
 	}
 	return "", nil
 }
@@ -117,6 +121,7 @@ func parseEnvConfig() error {
 		EnvConfig.DbConnectionString,
 		EnvConfig.DbConnectionStringFile,
 		"DB_CONNECTION_STRING",
+		true,
 	)
 	if err != nil {
 		return err
@@ -127,6 +132,7 @@ func parseEnvConfig() error {
 		EnvConfig.MaxMindLicenseKey,
 		EnvConfig.MaxMindLicenseKeyFile,
 		"MAXMIND_LICENSE_KEY",
+		true,
 	)
 	if err != nil {
 		return err
@@ -165,6 +171,8 @@ func parseEnvConfig() error {
 			EnvConfig.EncryptionKey,
 			EnvConfig.EncryptionKeyFile,
 			"ENCRYPTION_KEY",
+			// Do not trim spaces because the file should be interpreted as binary
+			false,
 		)
 		if err != nil {
 			return err
--- a/backend/internal/utils/jwk/utils.go
+++ b/backend/internal/utils/jwk/utils.go
@@ -15,7 +15,6 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"os"
 
 	"github.com/lestrrat-go/jwx/v3/jwa"
 	"github.com/lestrrat-go/jwx/v3/jwk"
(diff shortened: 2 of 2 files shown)


kmendell:
Can you add these to the docs please? Otherwise, it looks good to me at a glance, i need to pull and test it still though.

ItalyPaleAle:
Yes, I will document this (and the JSON logs) before we make the release

kmendell:
@ItalyPaleAle What do you think of that change? i think this looks cleaner than a whole bunch of if statements, as if we have a ton of sensitive varibales at some point it could get messy. Could go one ste futher and extract that into a util too. Also had to make sure the white space was trimmed from the files, as i was runnign into that issue.

giodamelio:
I can't tell for sure that this is not already done, but from a quick scan of the PR I don't think it is:

Would it be possible to allow the values for the files to be able to have environment vars in the paths that are resolved? I know that adds an extra layer of complexity, but it would be great for using [systemd-creds](https://systemd.io/CREDENTIALS/) to load the secrets into services.

Example for the Encryption Key.

I would set this env var on my service:

    ENCRYPTION_KEY_FILE=${CREDENTIALS_DIRECTORY}/pocket-id-encryption-key
    
Then in my service definition I would have:

    LoadCredentialEncrypted="pocket-id-encryption-key:/var/lib/credstore/pocket-id-encryption-key"

That would have systemd-creds decrypt the secret from `/var/lib/credstore/pocket-id-encryption-key` and mount it in a place that only that service can read, passing that path in as `$CREDENTIALS_DIRECTORY`.

If that is too much trouble I can work around it easily enough, I am just a systemd-creds evangelist apparently...

backend/internal/common/env_config.go
```diff
@@ -12,6 +12,20 @@ import (
...
+func resolveStringOrFile(directValue, filePath, varName string) (string, error) {
+	if directValue != "" {
+		return directValue, nil
+	}
+	if filePath != "" {
+		content, err := os.ReadFile(filePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to read secret '%s' from file '%s': %w", varName, filePath, err)
+		}
+		return strings.TrimSpace(string(content)), nil
```
ItalyPaleAle: I had considered this, but was wondering if this could be dangerous? What if the space is on purpose? We do not trim spaces from values read from env vars directly, so not sure it'd be safe to do this here

backend/internal/common/env_config.go
```diff
@@ -145,10 +160,21 @@ func parseEnvConfig() error {
 	case "":
 		EnvConfig.KeysStorage = "file"
 	case "database":
-		// If KeysStorage is "database", a key must be specified
-		if EnvConfig.EncryptionKey == "" && EnvConfig.EncryptionKeyFile == "" {
+		// Resolve encryption key using the same pattern
+		encryptionKey, err := resolveStringOrFile(
```
ItalyPaleAle, line 170: I had not done this for the encryption key because it's managed separately: https://github.com/pocket-id/pocket-id/blob/d12c1a0dc386f70c582821968ec0ed554bad0aa1/backend/internal/utils/jwk/utils.go#L49

I would prefer to keep this separate because it's very delicate. For example, the code added above for trimming whitespace (see my other comment) is very dangerous here, as would cause key files to be read differently, which would make keys fail to decrypt.

ItalyPaleAle (commented):
@kmendell thanks for the review, agree a helper is nice!

Not super sold about the whitespace trimming however - see my comments.

If you do want to keep trimming whitespaces, I would strongly recommend to not do that for the encryption key (and keep the logic that reads the file in the existing util function)

ItalyPaleAle:
@giodamelio that's interesting. I have to admit I wasn't very familiar with system-creds.

Not sure what is required to make it work. I assume it depends on where the env expansion happen. When you pass something like `ENCRYPTION_KEY_FILE=${CREDENTIALS_DIRECTORY}/pocket-id-encryption-key`, what value does the Pocket ID binary receive?

kmendell:
> @kmendell thanks for the review, agree a helper is nice!
> 
> Not super sold about the whitespace trimming however - see my comments.
> 
> If you do want to keep trimming whitespaces, I would strongly recommend to not do that for the encryption key (and keep the logic that reads the file in the existing util function)

Oh that makes sense, VSCode was just adding a new Line to the end of the file i was testing with and it does not load correctly if theres a new line, maybe we can only trim the ending maybe?

kmendell:
@ItalyPaleAle I changed it to only trim the right, and only trim spaces, tabs, returns, and new lines. This should solve both issues no?

ItalyPaleAle:
> @ItalyPaleAle I changed it to only trim the right, and only trim spaces, tabs, returns, and new lines. This should solve both issues no?

Not for the encryption key file. Its contents _could_ be binary, so must not be trimmed.

kmendell:
@ItalyPaleAle Ahh oky i understand, makes sense.

kmendell (approved):
Looks good to me, Thanks!

Issue #766, opened 2025-07-21
🐛 Bug Report: PocketID gets indexed by web crawlers


Etienne-bdt:
### Reproduction steps

Not really related to the app itself but, have a pocketid instance for some time and bad luck.
Google your domain
Tadah !

### Expected behavior

It should not be indexed using robots.txt (iirc)

### Actual Behavior

It gets crawled and indexed on google

### Version and Environment

Caddy reverse proxy with the latest pocketid

bluewalk:
Don't really think this is a bug though.

I've added a route block to Caddyfile for robots.txt, just to be sure it wouldn't get indexed.

```
route /robots.txt {
   header Content-Type "text/plain; charset=utf-8"
   respond <<EOT
User-agent: *
Disallow: /
EOT
}
```

Etienne-bdt:
Yeah, it might not really qualify as a bug, didn't know what to put in there
Thing is I never served robots.txt and never got indexed on my other services, but thanks I'll give your solution a try !

ElioDiNino:
It would be nice if there was an environment variable for whether to serve robots.txt or, alternatively, serve a default with the option to override. Most other self-hosted services block indexing by default.

stonith404:
Pocket ID shouldn't be indexed by crawlers. Are you sure your instance gets indexed?

https://github.com/pocket-id/pocket-id/blob/main/frontend/src/app.html#L7

ElioDiNino:
Ah yeah, the `noindex` meta tag should be enough to prevent indexing (although it doesn't stop Google and others from crawling). Not sure how @Etienne-bdt had his instance shown on Google 🧐

Etienne-bdt:
Right, I just saw the `noindex` tag.
I'm positive it got indexed however.

<img width="1080" height="257" alt="Image" src="https://github.com/user-attachments/assets/1698f304-da93-459d-8988-cbea54871a32" />

Google doesn't provide a lot of info aside that they indexed the website...

ItalyPaleAle:
I guess we could still create a static `robots.txt` to include in our app... Wouldn't hurt, right?

Etienne-bdt:
I added my domain to the google search console and got "some" insight. According to the console indexation and exploration are both allowed. 
If it's alright I'll open a PR to add the robots.txt with both disallow and no index directives.

kmendell:
Added in https://github.com/pocket-id/pocket-id/pull/806

PR #806, opened 2025-08-01
feat: add robots.txt to block indexing

Etienne-bdt:
Linked to #766

Serve a robots.txt which forbids both exploring and indexing by crawlers of all user agent
Hopefully this should stop indexing.

Note : If any other user gets their site indexed they can edit search results and delete them via the google search console if they own the domain under which their instance is served.

### DIFF
--- a/frontend/static/robots.txt
+++ b/frontend/static/robots.txt
@@ -1,0 +1,3 @@
+User-agent: *
+Disallow: /
+Noindex: /


ElioDiNino:
I'm pretty sure `Noindex` is not a valid `robots.txt` directive; it's only `Allow` and `Disallow`

Etienne-bdt:
Oh you're right, Noindex is [no longer supported](https://developers.google.com/search/blog/2019/07/a-note-on-unsupported-rules-in-robotstxt?hl=en). It's still a bit strange to me how my instance got indexed though.

Issue #755, opened 2025-07-16
🚀 Feature: Default user group for new users


dfunkt:
### Feature description

Select a specific group that gets assigned by default for new users. (or perhaps even create a new one if none exist)
### Pitch

In order for the login to be authorized, [Audiobookshelf](https://github.com/advplyr/audiobookshelf) requires every user to be part of a group (user/admin etc) if the groups claim is enabled in the settings.

dfunkt:
Implemented in https://github.com/pocket-id/pocket-id/commit/182d8090286f9953171c6c410283be679889aca7

PR #812, opened 2025-08-05
feat(signup): add default user groups and claims for new users

zeedif:
### Summary

This Pull Request resolves issue #755 by implementing a feature that allows administrators to define default user groups and custom claims for new users who register via Open Signup or Signup Tokens.

The primary motivation is to streamline the onboarding process for applications integrated via OIDC, such as Kavita, which may require specific roles (like a `Login` role) to be present for a user to be granted access. Previously, an administrator had to manually assign a user to a group after they registered, creating a delay and an extra administrative step.

With these changes, new users are automatically configured with the necessary permissions upon creation, allowing them to log into third-party services immediately.

### Implementation Details

-   **Backend**:
    -   Two new settings, `signupDefaultUserGroupIDs` and `signupDefaultCustomClaims`, have been added to the `AppConfig` model.
    -   The `UserService`'s `createUserInternal` method has been updated to read these settings and apply them to the new user.
    -   The `CustomClaimService` has been refactored to allow its claim update logic to be executed within an existing database transaction, ensuring the user creation process remains atomic.
-   **Frontend**:
    -   A new "Signup Defaults" card has been added to the "Application Configuration" page in the admin panel.
    -   This new section allows administrators to:
        -   Select one or more default user groups from a multi-select list.
        -   Define a list of default custom claims (key-value pairs).
    -   The `AppConfigService` on the frontend has been updated to correctly handle the serialization and deserialization of these new settings, which are stored as JSON strings in the database.

### A Note on Contribution

I would like to note that I implemented these changes with the assistance of AI, as I am not deeply proficient in Go or the Svelte framework.

I have tested these changes locally, and they appear to function as expected, resolving the issue for my use case. However, I would greatly appreciate a thorough review from the maintainers to identify any potential improvements, style inconsistencies, or bugs I may have missed.

I am fully open to feedback and ready to make any required changes to get this PR merged. Thank you for considering my contribution

## Summary by CodeRabbit

* New Features
  * New Signup settings in Admin > Application Configuration to control signup mode, default user groups, and default custom claims.
  * New users automatically receive configured groups and claims on signup (non-LDAP).

* Changes
  * Signup mode moved from General configuration to the new Signup settings card.
  * Updated success messages for Signup settings.
  * Translation updates/additions for EN and ES.

* Tests
  * Reorganized and expanded signup end-to-end tests; added stability improvements and broader flow coverage.

### DIFF
--- a/backend/internal/bootstrap/services_bootstrap.go
+++ b/backend/internal/bootstrap/services_bootstrap.go
@@ -58,6 +58,7 @@ func initServices(ctx context.Context, db *gorm.DB, httpClient *http.Client) (sv
 	}
 
 	svc.userGroupService = service.NewUserGroupService(db, svc.appConfigService)
+	svc.userService = service.NewUserService(db, svc.jwtService, svc.auditLogService, svc.emailService, svc.appConfigService, svc.customClaimService)
 	svc.ldapService = service.NewLdapService(db, httpClient, svc.appConfigService, svc.userService, svc.userGroupService)
 	svc.apiKeyService = service.NewApiKeyService(db, svc.emailService)
 


kmendell:
I'll try to look at this in a little bit, something we probably should add is Playwright E2E Tests for this since its a UI Element/Flow, if you wanna try to work on that, whle you are waiting for us to review it.

kmendell:
Oh im not sure if you just added them or if i was completly blind and missed them the first time.... I got side tracked with some stuff but still planning on looking over this as soon as i can.

frontend/src/routes/settings/admin/application-configuration/forms/app-config-signup-defaults-form.svelte
```diff
@@ -0,0 +1,71 @@
...
+	}
+</script>
+
+<form class="space-y-6" onsubmit={preventDefault(onSubmit)}>
+	<div>
+		<Label for="default-groups">{m.user_groups()}</Label>
+		<p class="text-muted-foreground mt-1 mb-2 text-xs">
+			{m.signup_defaults_groups_description()}
+		</p>
+		<MultiSelect
```
kmendell: We should only show this if groups exsist other wise it looks a little out of place: 

Image

frontend/src/routes/settings/admin/application-configuration/forms/app-config-signup-defaults-form.svelte
kmendell, line 1: I think we could move all the signup logic into one card, so the enable switch as well, not saying you have to but more putting this here so we remember

backend/internal/service/custom_claim_service.go
```diff
@@ -74,11 +92,6 @@ func (s *CustomClaimService) updateCustomClaims(ctx context.Context, idType idTy
 		seenKeys[claim.Key] = struct{}{}
 	}
 
-	tx := s.db.Begin()
```
kmendell, line 77: I may be missing something, but is there a reason for removing the rollbacks here?

frontend/src/routes/settings/admin/application-configuration/forms/app-config-signup-defaults-form.svelte
kmendell: Another thought going into this would be, to only show these new fields when signups are in the Open or With Tokens states?

backend/internal/service/custom_claim_service.go
zeedif: Previously, `updateCustomClaims` managed its own transaction. However, the new feature requires creating a user, assigning them to groups, and applying custom claims all within a single, atomic operation. If any of these steps fail, the entire user creation process should be rolled back.

To achieve this, I've refactored the function to `updateCustomClaimsInternal`, which now accepts a `*gorm.DB` transaction object (`tx`) from its caller. The transaction itself is now initiated and managed higher up in the call stack, specifically within the `UserService` methods like `SignUp` and `CreateUser`. This ensures that if applying the default claims fails, the newly created user is also rolled back, preventing partial or inconsistent data from being saved.

The new public methods, `UpdateCustomClaimsForUser` and `UpdateCustomClaimsForUserGroup`, now wrap the call to the internal method within their own transaction to maintain the original behavior for direct API calls.

kmendell:
@coderabbitai review

backend/internal/service/custom_claim_service.go
```diff
@@ -55,16 +55,34 @@ const (
...
+	var updatedClaims []model.CustomClaim
+	err := s.db.Transaction(func(tx *gorm.DB) error {
+		var txErr error
+		updatedClaims, txErr = s.updateCustomClaimsInternal(ctx, UserID, userID, claims, tx)
+		return txErr
+	})
+	if err != nil {
+		return nil, err
+	}
+	return updatedClaims, nil
```
stonith404:
```suggestion
	tx := s.db.Begin()
	defer func() {
		tx.Rollback()
	}()
	
	return  s.updateCustomClaimsInternal(ctx, UserID, userID, claims, tx)
```

I would prefer this syntax it's easier to read.

backend/internal/service/custom_claim_service.go
```diff
@@ -55,16 +55,34 @@ const (
...
+	var updatedClaims []model.CustomClaim
+	err := s.db.Transaction(func(tx *gorm.DB) error {
+		var txErr error
+		updatedClaims, txErr = s.updateCustomClaimsInternal(ctx, UserGroupID, userGroupID, claims, tx)
+		return txErr
+	})
+	if err != nil {
+		return nil, err
+	}
+	return updatedClaims, nil
```
stonith404:
```suggestion
func (s *CustomClaimService) UpdateCustomClaimsForUserGroup(ctx context.Context, userGroupID string, claims []dto.CustomClaimCreateDto) ([]model.CustomClaim, error) {
	tx := s.db.Begin()
	defer func() {
		tx.Rollback()
	}()

	return s.updateCustomClaimsInternal(ctx, UserGroupID, userGroupID, claims, tx)
}
```

Same here.

PR #815, opened 2025-08-07
refactor: use reflection to mark file based env variables

stonith404:
These changes make it simpler to set up environment variables that can be read from a file. This also helps avoid mistakes, like when a developer might accidentally try to access `MaxMindLicenseKeyFile` instead of `MaxMindLicenseKey` in the code.

Now, you can use `options:"file"` to mark environment variables. When marked this way, Pocket ID will try to read from `ANY_ENV_VARIABLE_FILE` instead of `ANY_ENV_VARIABLE`. If the file version isn't defined, it will use `ANY_ENV_VARIABLE` as a fallback.

### DIFF
--- a/backend/internal/cmds/key_rotate_test.go
+++ b/backend/internal/cmds/key_rotate_test.go
@@ -141,7 +141,7 @@ func testKeyRotateWithDatabaseStorage(t *testing.T, flags keyRotateFlags, wantEr
 	// Set up database storage config
 	envConfig := &common.EnvConfigSchema{
 		KeysStorage:   "database",
-		EncryptionKey: "test-encryption-key-characters-long",
+		EncryptionKey: []byte("test-encryption-key-characters-long"),
 	}
 
 	// Create test database


backend/internal/common/env_config.go
```diff
@@ -166,23 +126,9 @@ func parseEnvConfig() error {
...
-		encryptionKey, err := resolveStringOrFile(
-			EnvConfig.EncryptionKey,
-			EnvConfig.EncryptionKeyFile,
-			"ENCRYPTION_KEY",
-			// Do not trim spaces because the file should be interpreted as binary
-			false,
-		)
-		if err != nil {
-			return err
+		if EnvConfig.EncryptionKey == "" {
```
ItalyPaleAle: I just noticed this in the PR that was merged. For the encryption key, when we read it from file, we need to treat it as binary, so we shouldn't be putting it into a `string`.

Env vars can be treated as strings, but when we read from file it should be treated as `[]byte` - this is because it's possible people pass binary files (like with `head -c 32 /dev/random > key_file`).

backend/internal/common/env_config.go
```diff
@@ -191,3 +137,52 @@ func parseEnvConfig() error {
...
+		if envVarFileValue == "" {
+			continue
+		}
+
+		fileContent, err := os.ReadFile(envVarFileValue)
+		if err != nil {
+			return fmt.Errorf("failed to read file for env var %s: %w", envVarFileName, err)
+		}
+
+		field.SetString(string(fileContent))
```
ItalyPaleAle: There was a conversation with @kmendell where he was feeling strongly about the fact that for values read from files we should call `strings.TrimSpace` (but, importantly, **not** for the encryption key, because that's binary - see above)

kmendell: Its not that it wont work, I just was having issues with VScode being dumb and trying to format the file to add white space at the end, This normally wouldnt happen but got me thinking that is most circumstances we shouldnt need that whitespace.

ItalyPaleAle: Well, it's not uncommon for people to add a newline at the end. Even simple things like using `echo "foo" > file` will add a newline at the end (unless you do `echo -n "foo" > file`)

kmendell: To my point, Pocket id did not load when there was whitespace at the end of the file. Hence why i made that change.

backend/internal/common/env_config.go
stonith404: Thanks, it should now handle binary files correctly for the encryption key variable.

backend/internal/common/env_config.go
stonith404: Whitespaces should now be trimmed for non binary files.

ItalyPaleAle (approved):
Thanks for the changes!

PR #830, opened 2025-08-13
feat: add ability to set custom client ID

### LINKED ISSUE #508: 🚀 Feature: allow to set artbitrary Cliend-ID
### Feature description

Currently you cannot do it, since Pocket-ID automatically generates random UUID. Unfortunately there are apps which require to create exact Client IDs to work. Excellent example is OpenCloud: their desktop and mobile app has hard-coded IDs which they list in the doc: https://docs.opencloud.eu/docs/admin/configuration/authentication-and-user-management/external-idp

### Pitch

It will allow to support some setups I described above.

James18232:
this PR includes changes to support a custom client id/client secret to be used for a client.

This is required for services such as owncloud ocis, who use fixed client id/client secrets for some elements of their services.

I believe this is not strictly in accordance with the oidc specs, so this is proposed as an advanced option only, after a client is setup in the usual manner. 

happy for input if things are not quite right - i have done preliminary testing and think it is ok. 

fixes #508.


stonith404:
Thanks for your PR. I haven't looked at the code yet because it's still a draft, but based on your description I have the following note: The client secret shouldn't be definable through the UI. This could lead to insecure client IDs when an inexperienced user creates a client ID. The implementation of OIDC by ownCloud OCIS is incorrect in this case; a client should never define a client secret because then the client secret isn't secret anymore.

However, I would keep the option to define it through the API because that allows integrations like the [Terraform provider](https://github.com/Trozz/terraform-provider-pocketid), which enables a declarative approach to defining OIDC clients. In that context, it makes sense to define client secrets manually.

TLDR the option to define a client secret manually should be removed from the UI but kept in the API.

James18232:
Thanks.

I agree in principle, but not in practice.

Best not let perfect be the enemy of good.

Imo it is better to provide passkey only oidc support via pocketID than leave other less secure options like basic auth. A secure passkey implementation save for a (not secret) client secret  is still better than basic auth where end users can make poor passwords that are far less secure.

Leaving the api option is just adding an extra step.

This saves those ( like me ) from having to do sql surgery to make the shoe fit or rock another auth.

stonith404:
I understand that it's annoying to manually set the client secret with the API or even worse with SQL. 

Besides setting up OIDC for Owncloud, would you ever want to set the client secret manually? 

If not, this definitely has to be fixed by Owncloud by e.g using PKCE without a client secret. I just don't see a reason to implement something that probably gets misused by inexperienced Pocket ID users, only to fix an issue that isn't caused by us but by Owncloud.

James18232:
No there is no other purpose i can think of. As you say an actually secret client secret or the pkce public flow is a better approach - no arguments here 🤷‍♂️.

this is simply to support apps like OCIS that do not follow the specs perfectly - i am sure there are others.

Nevertheless in this instance OCIS use private+pkce so this is better than nothing. 

it still shouldn’t be the reason a client secret cannot be set by the user if there is an api that can do it with slightly more effort..

Perhaps an additional requirement that if a custom client secret is used it then requires the additional pkce on the assumption the secret is not secret but these seems a bit much.

i cant see it being misused when the default client setup is correct - its just to handle edge cases.

stonith404:
The main concern with allowing users to set the client secret manually is that inexperienced users might choose an insecure client ID or fail to understand the risks associated with a hard-coded client secret. While manually adding the client secret via the API is inconvenient, I believe this extra step is necessary because setting it manually is inherently insecure.

As you mentioned, aside from OwnCloud, we don’t know of any valid use cases for manually setting the client secret,  so enabling this feature would essentially be supporting an insecure practice. It's not a edge-case it's an invalid and insecure implementation by Owncloud.

To clarify why hard-coding a client secret is dangerous: it makes it easy for someone to create a phishing site that impersonates an OwnCloud client without the user realizing it. Moreover, if an attacker gains access to an authorization code, they could use the exposed secret to generate a valid access token.

James18232:
makes sense.

i can just remove it from the UI, or alternatively have a drop down selection to pick from a preset list of secrets for 'services of OIDC shame' that could be curated with the common clients like OCIS until they get with the program?

I think OCIS could also be fixed on pocketID end potentially anyway, as it does send a PKCE code, but evidently sends the client secret to the token endpoint. if the user sets the auth flow as public in the client settings, i think PocketID should ignore any client secret sent to the token endpoint no? this way you could essentially ignore the fixed secret and just treat it as a public client as it should be..

ItalyPaleAle:
Have you considered opening an issue with ownCloud first? This does sound like a bug in their implementation first of all...

As for an other option, how about a command for the pocket-id CLI to set the secret for a client? This way it's less cumbersome than an API call (because that requires passing an access token), but still not immediate in the UI.

James18232:
i have just removed any client secret input directly for now - sounds like it is generally not a good idea anyway. 

owncloud are aware of having a contant client secret based on the issued on their forum - there does not appear any motivation to fix it internally.

i believe there is still a bug on the PocketID side though, as it checks any client secret sent when requesting a token even when the client is set up as a public client. it should be ignoring any client secret for these clients, as it is worthless anyway.

frontend/src/routes/settings/admin/oidc-clients/oidc-client-form.svelte
```diff
@@ -105,6 +110,10 @@
 				return e;
 			});
 	}
+
+	function onReplaceClientId() {
+			replaceClientIdMessage = "(API integration TBA).";
```
kmendell: Just doing a quick overview, but is this still planning on being implemented? It looks like it's for the client id, so I assume so, but just putting this comment here as I know things get missed.

James18232: this can be deleted - sveltekit threw me a bit when messing around. the api call begins in replace-client-credentials to take an approach similar to federated identies - although i have reduced this to only the client id, so probably should rename this to replace-client-id.svelte..

James18232:
Thanks @stonith404. i was not sure i could piggy back off the existing `UPDATE /oidc/clients/:id` endpoint? the gorm save did not play nice with updating primary keys and added new entries. I also needed to replace all instances of client id that appears in the other tables that rely on positive reference, hence the new api route just to handle this task..

i should clarify this PR is now for a custom client ID  front-end input only, with additional backend API for client secret updates.

stonith404:
@James18232I think we could solve updating the foreign keys directly in the database with cascading referential integrity. But the current solution has some drawbacks anyways, for example we would have to make sure that the custom client ID doesn't contain any dangerous characters like `/` and `.`. I have to think a bit about it sorry, I'll let you know as soon as I find a good solution.

James18232:
@stonith404  sure. I did think about input sanitation but ultimately didn’t think it necessary as it is created by an admin who presumably already has control of the db. sanitation on the backend service would not hurt i suppose (say to alphanumeric + -).

i did not know you could cascade updates in sql in that way, so went the manual route. This does have the obvious drawback of having to update the service when the db structure changes lest they fall out of line.

a middle ground may be to only allow this advanced feature on new client creation (like the client secret being shown). This way the updates to the db remain controlled.

James18232:
simplified the front end as requested (just to tidy this up for now).

stonith404:
I'm closing this in favor of #864, it was just faster to create a new PR instead of editing this one. Thank you for your PR though, it provided a good foundation, the credit definitely goes to you.

James18232:
No worries, your PR is 👍 . Thanks.

PR #840, opened 2025-08-17
feat: support for url based icons

kmendell:
This adds support for oidc client logos to be retrived from a URL vs a file upload.


kmendell:
@coderabbitai review

frontend/src/routes/settings/admin/oidc-clients/oidc-client-list.svelte
```diff
@@ -71,16 +71,21 @@
...
+					variant="outline"
+					aria-label={m.edit()}><LucidePencil class="size-3 " /></Button
+				>
+				<Button
+					onclick={() => deleteClient(item)}
+					size="sm"
+					variant="outline"
+					aria-label={m.delete()}><LucideTrash class="size-3 text-red-500" /></Button
+				>
+			</div>
```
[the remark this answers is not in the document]
kmendell: on:click is svelte 4 syntax onclick is correct for svelte 5 update your memory.

savely-krasovsky:
I just want to highlight that adding such a feature usually leads to increased security risks. This is because both the link and its content are user inputs, which can expose the system to the following threats:

1. A provided link will eventually be fed into a Go or third-party HTTP client. For example, a specially crafted link could be used to enumerate internal resources that are only available on the intranet. In other cases, vulnerabilities in the link parser could be exploited to cause a DoS, enable the use of obscure protocols, and so on. It's something to keep in mind.
2. The downloaded file itself could be dangerous and could be used to exploit vulnerabilities within an image decoder. I would also send `Accept` header and check `Content-Type` response header. Checking image header and its magic bytes also makes sense IMO.
3. Saving arbitrary files from user provided URL should be also extra cautious. Don't trust user filenames, it could lead to path traversal vulnerabilities.

I would also add an option to disable this feature completely using environment variables.

stonith404 (approved):

stonith404:
@coderabbitai review

backend/internal/common/env_config.go
```diff
@@ -180,6 +181,25 @@ func validateEnvConfig(config *EnvConfigSchema) error {
...
 
+	// Validate LOCAL_IPV6_RANGES
+	ranges := strings.Split(config.LocalIPv6Ranges, ",")
+	for _, rangeStr := range ranges {
+		rangeStr = strings.TrimSpace(rangeStr)
+		if rangeStr == "" {
+			continue
+		}
+
+		_, ipNet, err := net.ParseCIDR(rangeStr)
```
savely-krasovsky, line 192: @stonith404 new `netip` std package suits better IMO: https://pkg.go.dev/net/netip#Addr.IsPrivate

stonith404: My knowledge of networking is very limited, but in IPv6, couldn’t every IP range theoretically be considered a private range? Is that why we implemented this environment variable, or am I mistaken?

savely-krasovsky: @stonith404 yes, you are right IPv6 ULA addresses have only few valid cases and building real networks using them is most probably a mistake. I commented the wrong line, I just wanted to hint that newer `netip` is more handy and versatile than `net`.

kmendell:
LGTM :), Glad we went with the Popover approach looks alot better.

PR #846, opened 2025-08-18
tests: use proper async calls for cleanupBackend function

kmendell:
Also fixed the warning in the jwt.util about top level async.


stonith404:
What warning did you receive about top-level await, and where did you get it? I think top-level await should be fine, we might just need to specify a higher Node.js version since this is a relatively new feature.

kmendell:
@stonith404 I cant remember the exact error, But it was like "top level await can only be used when module is `nodenext` ....... etc" with more options, i tried changing that in package.json but it was not working for me even after restarting the TS Server

PR #854, opened 2025-08-20
fix: move audit log call before TX is committed

ItalyPaleAle:
Partially reverts fe003b927ce7772692439992860c804de89ce424 . Calls to `CreateNewSignInWithEmail` accept a transaction, so they should be made before the tx is committed


kmendell (approved):

stonith404:
Thanks for catching that!

PR #855, opened 2025-08-20
fix: for one-time access tokens and signup tokens, pass TTLs instead of absolute expiration date

### LINKED ISSUE #804: 🐛 Bug Report: Login Code too long
### Reproduction steps

occasionally (1% of the time) when you click create login code, the app generates a login code longer than 6 digits (16 digits).

Image

the app will continue to generate long login codes until the container is restarted, after which it will return to regular length.

I cannot reliably reproduce the bug.

### Expected behavior

it should create a 6 digit login code.

### Actual Behavior

creates a login code longer than 6 digits.

### Version and Environment

pocketid: v1.6.4
reverse proxy: caddy
docker: (rootless)

### Log Output

none available.

ItalyPaleAle:
Fixes #804

We perform computations on expiration times on the server, which means that:

1. If there's a clock drift with the client, it doesn't change the length of the token (see #804)
2. We don't need to trust what the client says. Users won't be allowed to create tokens for themselves that last longer than 15 mins.

Also, enforces a maximum TTL, server-side, of 31 days, even for people making direct API calls (the UI was not allowing more than 30 days already)


backend/internal/controller/user_controller.go
```diff
@@ -331,10 +336,20 @@ func (uc *UserController) createOneTimeAccessTokenHandler(c *gin.Context, own bo
...
 	}
 
+	var ttl time.Duration
 	if own {
 		input.UserID = c.GetString("userID")
-	}
-	token, err := uc.userService.CreateOneTimeAccessToken(c.Request.Context(), input.UserID, input.ExpiresAt)
+		ttl = defaultOneTimeAccessTokenDuration
+	} else {
+		ttl = input.TTL.Duration
```
stonith404, line 344: I would prefer to handle this validation with a binding in the DTO. You can create a custom validator in the [`validation.go`](https://github.com/pocket-id/pocket-id/blob/7ef1bdd5d02da4a295587f44cd49dd0df3702047/backend/internal/dto/validations.go) file.

stonith404:
Do you think it would make sense to use TTL instead of a date for the sign-up tokens as well?

backend/internal/controller/user_controller.go
ItalyPaleAle: Done!

ItalyPaleAle:
> Do you think it would make sense to use TTL instead of a date for the sign-up tokens as well?

Done!

PR #856, opened 2025-08-20
feat: strip debug symbol from backend binary

maximerobine:
Stripping debug symbols from backend binary to reduce binary (and docker image) size.
````
~/git/pocket-id main ❯ docker image ls
REPOSITORY                 TAG       IMAGE ID       CREATED          SIZE
pocket-id-opti             latest    eb413f6a47d5   22 minutes ago   89.7MB
pocket-id                  latest    9affda5ce232   24 minutes ago   121MB

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -34,7 +34,7 @@ RUN VERSION=$(cat /build/.version) \
   GOOS=linux \
   go build \
     -tags "${BUILD_TAGS}" \
-    -ldflags="-X github.com/pocket-id/pocket-id/backend/internal/common.Version=${VERSION} -buildid=${VERSION}" \
+    -ldflags="-X github.com/pocket-id/pocket-id/backend/internal/common.Version=${VERSION} -buildid=${VERSION} -s" \
```
ItalyPaleAle: This Dockerfile is primarily used by our test infra, so I'd leave the debug symbols here

ItalyPaleAle:
I think this is generally good, save for my comment above.

Do you think we could/should add `-w` too?

Dockerfile
maximerobine: Ok my bad, reverting

maximerobine:
I did not explicitly add the `-w` because it is included in `-s`.
From https://pkg.go.dev/cmd/link:
> -s
	Omit the symbol table and debug information.
	Implies the -w flag, which can be negated with -w=0.
	
I can explicitly add it if you want ?

ItalyPaleAle (approved):

Issue #857, opened 2025-08-20
🚀 Feature: Different base url for openid configuration


DerSteph:
### Feature description

Add possibility to have different base urls for the openid configuration endpoints via env variables, at least for those who are called by the OIDC client. 

These are (at least according to chatgpt based on the openid-configuration)
- token_endpoint
- userinfo_endpoint
- introspection_endpoint
- jwks_uri

For those, either have one env variable INTERNAL_APP_URL or for each one an own env variable.

The url would then only be used, when INTERNAL_APP_URL is defined, otherwise it would still use APP_URL as default.

If you want, i can also provide an pull request with my changes.

### Pitch

Im currently using pocket-id in my homelab in a small kubernetes cluster and im using mDNS annoucements via a customized controller to publish my dns records in my local network. Also im using traefik-oidc-auth as oidc client.

In general this mean my pocket-id instance has pocket-id.local as domain in my home network.

When im trying to use pocket-id for adding authentication for one of my services, i cannot continue after logging in successfully, as i get `Failed to exchange auth code`.
<img width="421" height="145" alt="Image" src="https://github.com/user-attachments/assets/2cd310a3-0d30-4d7a-8302-8261357bd0a4" />

When investigating, why its not working, i found that traefik-oidc-auth cannot access pocket-id.local, as it cannot resolve the mDNS domain in the pod itself.
<img width="1873" height="111" alt="Image" src="https://github.com/user-attachments/assets/e70aab2a-c417-4b84-87e4-f59e8a2b80e0" />

By digging in a bit deeper, i found out that pocket-id is providing the `.well-known/openid-configuration` which is used by traefik-oidc-auth to find out which urls have to be called.

`
{
  //...
  "token_endpoint": "https://pocket-id.local/api/oidc/token",
  "userinfo_endpoint": "https://pocket-id.local/api/oidc/userinfo"
}
`

Ive then changed the source code of pocket-id, so that can define via env variables custom base urls for some of the endpoints and then my login workflow worked as expected.

`
APP_URL: https://pocket-id.local
INTROSPECTION_ENDPOINT: http://pocket-id.pocket-id-system.svc.cluster.local:1411
JWKS_URI: http://pocket-id.pocket-id-system.svc.cluster.local:1411
TOKEN_ENDPOINT: http://pocket-id.pocket-id-system.svc.cluster.local:1411
USERINFO_ENDPOINT: http://pocket-id.pocket-id-system.svc.cluster.local:1411
`

[...]

PR #858, opened 2025-08-20
feat: add custom base url

DerSteph:
Fixes: https://github.com/pocket-id/pocket-id/issues/857

Adds env variables which can customize the base urls used in well_known_controller.

Can be used for setups, where the oicd client cannot resolve the app url.


ItalyPaleAle:
Hey! Thanks for this PR.

Could you provide some more context on the issue you're trying to solve?

DerSteph:
Hey,
Ive written my issue here https://github.com/pocket-id/pocket-id/issues/857 :)

Burnett01:
nice feature! would love to see this

kmendell:
@coderabbitai review

backend/internal/common/env_config.go
```diff
@@ -133,6 +136,18 @@ func parseEnvConfig() error {
 		return errors.New("APP_URL must not contain a path")
 	}
 
+	// only check if internal app url is not default value
+	if EnvConfig.InternalAppURL != EnvConfig.AppURL {
```
ItalyPaleAle: By making this change, you are requiring users to always set `INTERNAL_APP_URL` if they are setting `APP_URL`.

Instead:

1. Leave the default value of `InternalAppURL` (L94) empty
2. Here:
    - if the value of `InternalAppURL` is empty, set it to the value of `AppURL`
    - if it's not empty, do the validation below

ItalyPaleAle (changes requested):

ItalyPaleAle:
I have left a review comment on this PR, but please see my comment [here](https://github.com/pocket-id/pocket-id/issues/857#issuecomment-3230160752) as to why I'm fundamentally not supportive of this change, which I believe goes against the OIDC specs

kmendell (approved):

DerSteph:
Ah thanks for changes and merge! Was very busy at the weekend.

PR #863, opened 2025-08-22
fix: enable foreign key check for sqlite

stonith404:
By default Sqlite doesn't enforce foreign key checks which can lead to orphan rows in the database and confusing behavior. This changes the default connection string of Sqlite so that the connections enforce foreign key checks. 

For this change I have added a migration, that cleans all potential orphan rows, adds missing foreign keys and brings the `ON DELETE` behavior of Postgres and Sqlite in sync.


ItalyPaleAle (approved):
LGTM

My only concern is that we've put in the docs the connection string without FK. We will need to update the docs, of course, but if users have manually modified the conn string, they may miss the memo about foreign keys being required now.

Should we do some more parsing on the connstring and enforce that FKs are enabled?

stonith404:
Thanks for the review. 

I mean the foreign key check is basically opt-in, it won't break anything if a user doesn't enable the check, it could just lead to orphan rows. Because of that I would just mention in the release notes that it's recommended to add the new parameter to the connection string. Or would you really force the users to use this option?

ItalyPaleAle:
> I mean the foreign key check is basically opt-in, it won't break anything if a user doesn't enable the check, it could just lead to orphan rows

But if the user decides to enable it later, and there are orphaned rows, that will fail, right? That's why you've added all the DELETE queries in here

stonith404:
Oh yes, you're right, I missed that. In that case I think it makes sense to force this parameter, regardless which connection string the user is using.

kmendell:
I think forcing this would be a good idea as well.

backend/internal/bootstrap/db_bootstrap.go
```diff
@@ -86,11 +86,9 @@ func connectDatabase() (db *gorm.DB, err error) {
 		if common.EnvConfig.DbConnectionString == "" {
 			return nil, errors.New("missing required env var 'DB_CONNECTION_STRING' for SQLite database")
 		}
-		if !strings.HasPrefix(common.EnvConfig.DbConnectionString, "file:") {
-			return nil, errors.New("invalid value for env var 'DB_CONNECTION_STRING': does not begin with 'file:'")
-		}
 		sqliteutil.RegisterSqliteFunctions()
 		connString, err := parseSqliteConnectionString(common.EnvConfig.DbConnectionString)
+		fmt.Println("CONNSTRING", connString)
```
ItalyPaleAle: Whoops
```suggestion
```

PR #874, opened 2025-08-24
feat: support automatic db migration rollbacks

stonith404:
This PR adds support for automatic rollbacks of database migrations when the user downgrades to a older release. Pocket ID fetches the necessary "down migrations" from GitHub because they are obviously not included in the older versions.


ItalyPaleAle:
I like this idea, but I'm also a bit nervous about allowing the app to execute code (queries) straight from a web-based source.

How about, if we detect a downgrade:

1. Refuse to start up
2. Allow doing a downgrade of the DB only if the user passes an explicit flag, like the env var `ALLOW_DOWNGRADE=1` (and call out that it will fetch code from GitHub)

ItalyPaleAle (approved):

PR #884, opened 2025-08-24
fix(i18n): missing translations fix

savely-krasovsky:


frontend/src/lib/utils/audit-log-translator.ts
```diff
@@ -0,0 +1,28 @@
+import { m } from '$lib/paraglide/messages';
+
+/**
+ * Translates an audit log event type using paraglide messages.
+ * Falls back to a formatted string if no specific translation is found.
+ * @param event The event type string from the backend (e.g., "CLIENT_AUTHORIZATION").
+ * @returns The translated string.
+ */
+export function translateAuditLogEvent(event: string): string {
```
stonith404, line 17: With Paraglide, dynamically accessing translation strings isn’t recommended. Paraglide optimizes performance by only loading the messages needed for each page. However, if you access strings dynamically, it has no way of knowing which ones will be used, so it ends up loading all translations on page load.

Instead, I recommend defining all possible event translation strings in a single object. This way, Paraglide can efficiently load only what’s required.

stonith404 (changes requested):
Thanks for your PR, please see my comment above.

frontend/src/lib/utils/audit-log-translator.ts
savely-krasovsky: Pushed a fix. Not sure if this is optimal, but it works.

Issue #814, opened 2025-08-07
🚀 Feature: Add display_name to users


Elyviere:
### Feature description

A new field added to the user, "Display Name", which can be used by apps which separate usernames and display_name, and that can be synchronised from LDAP.

### Pitch

I'm using the ldap-integration in PocketID, and in my ldap (LLDAP) the field display_name is included by default on users. I'm also using an app, Reitti, which I'm using PocketID OAuth to sign in to. Reitti has a separate field for username and display_name. 

So, it stands to reason that I should be able to get my display_name which I've input in lldap and give it to Reitti. However, because PocketID doesn't get the display_name from ldap, it doesn't get included when I sign in to Reitti.

stonith404:
The display name is just the first and last name combined, right? If so, Pocket ID returns a "name" claim which includes the full name.

dragonfly8677:
No, it's a separate field. It's used to decide which name to show other users, as sometimes you don't want to use your username or first name. Here are the user fields in LLDAP:

<img width="691" height="539" alt="Image" src="https://github.com/user-attachments/assets/4978fc91-3593-446a-bbe0-545d5ab1a8f0" />

stonith404:
Okay, thanks for the clarification. I'm marking this as "needs more upvotes" to check if other users want this too.

PR #898, opened 2025-08-29
feat: user display name

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/814

### DIFF
--- a/backend/internal/dto/user_dto.go
+++ b/backend/internal/dto/user_dto.go
@@ -10,6 +10,7 @@ type UserDto struct {
 	Email        string           `json:"email" `
 	FirstName    string           `json:"firstName"`
 	LastName     string           `json:"lastName"`
+	DisplayName  string           `json:"displayName"`
 	IsAdmin      bool             `json:"isAdmin"`
 	Locale       *string          `json:"locale"`
 	CustomClaims []CustomClaimDto `json:"customClaims"`
@@ -19,14 +20,15 @@ type UserDto struct {
 }
 
 type UserCreateDto struct {
-	Username  string  `json:"username" binding:"required,username,min=2,max=50" unorm:"nfc"`
-	Email     string  `json:"email" binding:"required,email" unorm:"nfc"`
-	FirstName string  `json:"firstName" binding:"required,min=1,max=50" unorm:"nfc"`
-	LastName  string  `json:"lastName" binding:"max=50" unorm:"nfc"`
-	IsAdmin   bool    `json:"isAdmin"`
-	Locale    *string `json:"locale"`
-	Disabled  bool    `json:"disabled"`
-	LdapID    string  `json:"-"`
+	Username    string  `json:"username" binding:"required,username,min=2,max=50" unorm:"nfc"`
+	Email       string  `json:"email" binding:"required,email" unorm:"nfc"`
+	FirstName   string  `json:"firstName" binding:"required,min=1,max=50" unorm:"nfc"`
+	LastName    string  `json:"lastName" binding:"max=50" unorm:"nfc"`
+	DisplayName string  `json:"displayName" binding:"max=100" unorm:"nfc"`
+	IsAdmin     bool    `json:"isAdmin"`
+	Locale      *string `json:"locale"`
+	Disabled    bool    `json:"disabled"`
+	LdapID      string  `json:"-"`
 }
 
 type OneTimeAccessTokenCreateDto struct {
--- a/backend/internal/model/user.go
+++ b/backend/internal/model/user.go
@@ -13,14 +13,15 @@ import (
 type User struct {
 	Base
 
-	Username  string `sortable:"true"`
-	Email     string `sortable:"true"`
-	FirstName string `sortable:"true"`
-	LastName  string `sortable:"true"`
-	IsAdmin   bool   `sortable:"true"`
-	Locale    *string
-	LdapID    *string
-	Disabled  bool `sortable:"true"`
+	Username    string `sortable:"true"`
+	Email       string `sortable:"true"`
+	FirstName   string `sortable:"true"`
+	LastName    string `sortable:"true"`
+	DisplayName string `sortable:"true"`
+	IsAdmin     bool   `sortable:"true"`
+	Locale      *string
+	LdapID      *string
+	Disabled    bool `sortable:"true"`
 
 	CustomClaims []CustomClaim
 	UserGroups   []UserGroup `gorm:"many2many:user_groups_users;"`
@@ -31,7 +32,12 @@ func (u User) WebAuthnID() []byte { return []byte(u.ID) }
 
 func (u User) WebAuthnName() string { return u.Username }
 
-func (u User) WebAuthnDisplayName() string { return u.FirstName + " " + u.LastName }
+func (u User) WebAuthnDisplayName() string {
+	if strings.TrimSpace(u.DisplayName) != "" {
+		return u.DisplayName
+	}
+	return u.FirstName + " " + u.LastName
+}
 
 func (u User) WebAuthnIcon() string { return "" }
 
--- a/backend/internal/service/custom_claim_service.go
+++ b/backend/internal/service/custom_claim_service.go
@@ -25,6 +25,7 @@ func isReservedClaim(key string) bool {
 		"name",
 		"email",
 		"preferred_username",
+		"display_name",
 		"groups",
 		TokenTypeClaim,
 		"sub",
--- a/backend/internal/service/oidc_service.go
+++ b/backend/internal/service/oidc_service.go
@@ -1808,6 +1808,9 @@ func (s *OidcService) getUserClaimsFromAuthorizedClient(ctx context.Context, aut
 		claims["given_name"] = user.FirstName
 		claims["family_name"] = user.LastName
 		claims["name"] = user.FullName()
+		if strings.TrimSpace(user.DisplayName) != "" {
+			claims["display_name"] = user.DisplayName
+		}
 		claims["preferred_username"] = user.Username
 		claims["picture"] = common.EnvConfig.AppURL + "/api/users/" + user.ID + "/profile-picture.png"
 
--- a/backend/internal/service/user_service.go
+++ b/backend/internal/service/user_service.go
@@ -245,12 +245,13 @@ func (s *UserService) CreateUser(ctx context.Context, input dto.UserCreateDto) (
 
 func (s *UserService) createUserInternal(ctx context.Context, input dto.UserCreateDto, isLdapSync bool, tx *gorm.DB) (model.User, error) {
 	user := model.User{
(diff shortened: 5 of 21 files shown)


backend/internal/model/user.go
```diff
@@ -31,7 +32,12 @@ func (u User) WebAuthnID() []byte { return []byte(u.ID) }
 
 func (u User) WebAuthnName() string { return u.Username }
 
-func (u User) WebAuthnDisplayName() string { return u.FirstName + " " + u.LastName }
+func (u User) WebAuthnDisplayName() string {
+	if strings.TrimSpace(u.DisplayName) != "" {
```
ItalyPaleAle: I recommend not trimming when getting the data. Instread, trim before storing into the dtabase


```suggestion
	if u.DisplayName != "" {
```

backend/internal/service/custom_claim_service.go
```diff
@@ -25,6 +25,7 @@ func isReservedClaim(key string) bool {
 		"name",
 		"email",
 		"preferred_username",
+		"display_name",
```
ItalyPaleAle, line 28: Could this be a breaking change if people have defined a claim called `display_name`?

backend/internal/service/oidc_service.go
```diff
@@ -1808,6 +1808,9 @@ func (s *OidcService) getUserClaimsFromAuthorizedClient(ctx context.Context, aut
 		claims["given_name"] = user.FirstName
 		claims["family_name"] = user.LastName
 		claims["name"] = user.FullName()
+		if strings.TrimSpace(user.DisplayName) != "" {
+			claims["display_name"] = user.DisplayName
+		}
```
ItalyPaleAle: Ditto
```suggestion
		claims["display_name"] = user.DisplayName
```

backend/internal/service/custom_claim_service.go
kmendell: I didnt think of that tbh, you may be right. I'll talk it over with elias to see what he says.

backend/internal/model/user.go
kmendell: Okay ill remove it, , the normalization you added (which i set the key for this) should handle that right? or am i mistaken?

stonith404:
@ItalyPaleAle Thanks for pointing out the potential breaking change. I think there is a second, even more breaking change with LDAP but I'm not 100% sure. @kmendell will it break existing LDAP installations that don't define the display name attribute? Because in the LDAP configuration form the display name attribute is required and all current installations won't have this defined.

kmendell:
@stonith404 We could set the default of it to the `cn` attribute? or just write some logic to fallback to the `cn` attribute if nothing is returned for the displayName. but the simpliest appraoch would be to just set the default value to `cn` simalr to how the group filters are being set by default.

stonith404:
@kmendell You're right, that makes sense, I've made the changes. I'll try to fix the e2e test ASAP but the issue is really strange, I have no clue why my changes broke that the Zod validation messages won't get loaded correctly anymore.

backend/resources/migrations/postgres/20250829120000_user_display_name.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE users ADD COLUMN display_name TEXT;
+
+UPDATE users
+SET display_name = trim(coalesce(first_name,'') || ' ' || coalesce(last_name,''));
+
+ALTER TABLE users ALTER COLUMN display_name SET NOT NULL;
```
ItalyPaleAle, line 6: Two things here:

1. I don't believe updating existing rows is necessary, since if there's no display_name, then we perform this at display time.
2. No need to first add the column, then set it to not null.


```suggestion
ALTER TABLE users ADD COLUMN display_name TEXT NOT NULL;
```

backend/resources/migrations/sqlite/20250829120000_user_display_name.up.sql
```diff
@@ -0,0 +1,44 @@
...
+       ldap_id,
+       locale,
+       disabled
+FROM users;
+
+DROP TABLE users;
+ALTER TABLE users_new
+    RENAME TO users;
+
+CREATE UNIQUE INDEX users_ldap_id ON users (ldap_id);
```
ItalyPaleAle: Same as postgres:

```suggestion
ALTER TABLE users ADD COLUMN display_name TEXT NOT NULL;  
```

backend/internal/model/user.go
```diff
@@ -13,14 +13,15 @@ import (
...
-	FirstName string `sortable:"true"`
-	LastName  string `sortable:"true"`
-	IsAdmin   bool   `sortable:"true"`
-	Locale    *string
-	LdapID    *string
-	Disabled  bool `sortable:"true"`
+	Username    string  `sortable:"true"`
+	Email       string  `sortable:"true"`
+	FirstName   string  `sortable:"true"`
+	LastName    *string `sortable:"true"`
```
ItalyPaleAle: Why changing this to a pointer? I don't think that's needed, and makes things more complex... A null last name or a zero value one should be treated the same
```suggestion
	LastName    string `sortable:"true"`
```

backend/internal/service/oidc_service.go
```diff
@@ -1840,6 +1840,8 @@ func (s *OidcService) getUserClaimsFromAuthorizedClient(ctx context.Context, aut
 		claims["given_name"] = user.FirstName
 		claims["family_name"] = user.LastName
 		claims["name"] = user.FullName()
+		claims["display_name"] = user.DisplayName
+
 		claims["preferred_username"] = user.Username
 		claims["picture"] = common.EnvConfig.AppURL + "/api/users/" + user.ID + "/profile-picture.png"
```
ItalyPaleAle: Suggest moving these profile claims *after* the custom claims, to avoid custom claims overriding them (see the point above about potential conflict on display_name)

backend/internal/service/user_service.go
```diff
@@ -735,11 +737,17 @@ func (s *UserService) SignUp(ctx context.Context, signupData dto.SignUpDto, ipAd
...
-		Username:  signupData.Username,
-		Email:     signupData.Email,
-		FirstName: signupData.FirstName,
-		LastName:  signupData.LastName,
+		Username:    signupData.Username,
+		Email:       signupData.Email,
+		FirstName:   signupData.FirstName,
+		LastName:    signupData.LastName,
+		DisplayName: displayName,
 	}
```
ItalyPaleAle: Can be reverted since the display name is computed at display time if empty. Better not store extra info in the DB if not needed


```suggestion
	userToCreate := dto.UserCreateDto{
		Username:    signupData.Username,
		Email:       signupData.Email,
		FirstName:   signupData.FirstName,
		LastName:    signupData.LastName,
	}

```

ItalyPaleAle (changes requested):

backend/internal/model/user.go
stonith404: `LastName` can be null, and Gorm suggests using pointers for [nullable fields](https://gorm.io/docs/models.html#Declaring-Models). I'm not a fan of this but there isn't really another way.

stonith404:
@ItalyPaleAle I’d suggest storing the default DisplayName in the database instead of making it nullable and computing it on the fly. While this does take up a few extra bytes, it simplifies the codebase. If we computed it dynamically, we’d need to handle that logic in multiple places, which introduces extra boilerplate and point of failures. This is especially relevant for DTO mapping. Right now, the display name can be mapped automatically from the DB model to DTOs, but if it were computed, we’d have to set it manually everywhere.

backend/internal/model/user.go
ItalyPaleAle: Does the column need to be nullable?

Since we treat a NULL last name the same as an empty one (in fact we convert from a null pointer to an empty string), we could simplify and make it non-nullable.

stonith404: If the last name isn’t set, it should be stored as NULL in my opinion. The only place we currently handle a nil last name as an empty string is [here](https://github.com/pocket-id/pocket-id/pull/898/files#diff-dce531fa46c42f2676562274eb8f6b83bda90b47e9b984904cc6826a98bc54d2R81), which is just for simplifying the code a bit.

I think it’s cleaner to make optional fields nullable in the database:
- Filtering becomes easier (WHERE ... IS NOT NULL vs. WHERE ... != "").
- The DTO mapper correctly returns null instead of an empty string.

ItalyPaleAle: I still don't agree with this point. IMHO the complexity of having to deal with pointers, nil values (Go), and nulls (JS) is significant. I see the reason for nullable values when there needs to be a difference between the unset state and the zero state (e.g. when the empty string has a different meaning than the null value...)

In any case, your call here :)

backend/resources/migrations/sqlite/20250829120000_user_display_name.up.sql
```diff
@@ -0,0 +1,44 @@
...
+        unique,
+    email        TEXT                  not null
+        unique,
+    first_name   TEXT,
+                     last_name TEXT,
+    display_name TEXT                  not null,
+    is_admin     NUMERIC default FALSE not null,
+    ldap_id      TEXT,
+    locale       TEXT,
+    disabled     NUMERIC default FALSE not null
```
ItalyPaleAle: nit... something's off with the indentation here

stonith404:
@coderabbitai review

stonith404:
@coderabbitai review

backend/internal/dto/user_dto.go
```diff
@@ -22,14 +23,15 @@ type UserDto struct {
...
-	FirstName string  `json:"firstName" binding:"required,min=1,max=50" unorm:"nfc"`
-	LastName  string  `json:"lastName" binding:"max=50" unorm:"nfc"`
-	IsAdmin   bool    `json:"isAdmin"`
-	Locale    *string `json:"locale"`
-	Disabled  bool    `json:"disabled"`
-	LdapID    string  `json:"-"`
+	Username    string  `json:"username" binding:"required,username,min=2,max=50" unorm:"nfc"`
+	Email       string  `json:"email" binding:"required,email" unorm:"nfc"`
+	FirstName   string  `json:"firstName" binding:"required,min=1,max=50" unorm:"nfc"`
+	LastName    string  `json:"lastName" binding:"required,min=1,max=50" unorm:"nfc"`
```
ItalyPaleAle: I assume the `required` here was a typo? Since we allow it to be empty/null

Plus, [not everyone has a last name](https://en.wikipedia.org/wiki/Mononym)

stonith404: Oops, thanks for pointing that out. I meant to add it to the display name because this one is required.

Baspla:
display_name is not a standard oidc field, right? the closest thing i could find is nickname.

PR #901, opened 2025-08-30
feat: initial client_credentials flow support

### LINKED ISSUE #890: 🚀 Feature: Client Credentials flow support
### Feature description

[Client Credentials flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) is a very useful flow for m2m integrations. Since #566 is already merged it would be nice to support getting access token using both methods:
1. With a pair of `client_id` and `client_secret`.
2. With a federated credential (`client_id` + `client_assertion`).

### Pitch

Client Credentials flow can be used to programmatically bypass IAPs (Identity-Aware Proxy) such as [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy) and many others. For example it will allow monitoring products like Uptime Kuma or Gatus to monitor services behind those proxies by providing access token they received from Pocket-ID using client credential flow.

savely-krasovsky:
`resource` parameter could be used by every flow according to RFC 9068 (https://datatracker.ietf.org/doc/html/rfc9068#name-requesting-a-jwt-access-tok), but it's out of the scope of that PR.

Closes #890


ItalyPaleAle:
This LGTM at a high level, but could you add some tests to validate the behavior?

savely-krasovsky:
@ItalyPaleAle sure, I wanted to write some tests, but the problem is that I don't see tests which validate complete methods like `createTokenFromRefreshToken` or `createTokenFromAuthorizationCode`. Maybe I am not looking in the right places. Sure I can copy the body of `createTokenFromClientCredentials` and create a test case, but I don't see a lot of value, since test and codebase will be out of sync very soon. It seems for me that token creation flow lacking integration tests as a thing right now. Would be happy to listen suggestions though.

savely-krasovsky:
@ItalyPaleAle well, I figured it out, IMO looks fine. I think it would be nice to probably write similar tests for other flows.

backend/internal/service/oidc_service.go
```diff
@@ -333,6 +336,34 @@ func (s *OidcService) createTokenFromDeviceCode(ctx context.Context, input dto.O
...
+		return CreatedTokens{}, err
+	}
+
+	// GenerateOAuthAccessToken uses user.ID as a "sub" claim. Prefix is used to take those security considerations
+	// into account: https://datatracker.ietf.org/doc/html/rfc9068#name-security-considerations
+	dummyUser := model.User{
+		Base: model.Base{ID: "client-" + client.ID},
+	}
+
+	audClaim := client.ID
```
kmendell, line 352: Based on spec the aud should be the Authorization Server Token endpoint 

> aud
> REQUIRED. Audience. The aud (audience) Claim. Value that identifies the Authorization Server as an intended audience. The Authorization Server MUST verify that it is an intended audience for the token. The Audience SHOULD be the URL of the Authorization Server's Token Endpoint.

Unless a `resource` value is present.

kmendell (changes requested):
The grant type also needs to be put in the well known controller to advertise the `client_credentials` grant type is supported.

https://github.com/pocket-id/pocket-id/blob/6188dc6fb7ae47011d62c3b6d6860fdc2cd296c7/backend/internal/controller/well_known_controller.go#L86

backend/internal/service/oidc_service.go
savely-krasovsky: I assume this quote is from there: https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication
This applies only to client authentication (e.g. what claims JWT assertion should contain when client come to AS to exchange authorization code). This PR is about client credentials flow, which is another beast.

savely-krasovsky: OIDC doesn't extend Client Credentials flow, so it's only covered by OAuth2 general specification. In our case we are interested in those ones:
https://datatracker.ietf.org/doc/html/rfc6749#section-4.4
https://datatracker.ietf.org/doc/html/rfc9068

backend/internal/service/oidc_service.go
```diff
@@ -333,6 +336,34 @@ func (s *OidcService) createTokenFromDeviceCode(ctx context.Context, input dto.O
...
+	}
+
+	accessToken, err := s.jwtService.GenerateOAuthAccessToken(dummyUser, audClaim)
+	if err != nil {
+		return CreatedTokens{}, err
+	}
+
+	return CreatedTokens{
+		AccessToken: accessToken,
+		ExpiresIn:   time.Hour,
```
ItalyPaleAle: nit (could be for a subsequent PR): should be moved to a constant like we have for RefreshTokenDuration

backend/internal/service/oidc_service_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/lestrrat-go/jwx/v3/jwa"
 	"github.com/lestrrat-go/jwx/v3/jwk"
 	"github.com/lestrrat-go/jwx/v3/jwt"
+	"github.com/pocket-id/pocket-id/backend/internal/model"
```
ItalyPaleAle: nit: please fix the imports - all pocket-id ones should be at the bottom

backend/internal/service/oidc_service_test.go
```diff
@@ -384,4 +394,119 @@ func TestOidcService_verifyClientCredentialsInternal(t *testing.T) {
...
+				input := dto.OidcCreateTokensDto{
+					ClientID:     confidentialClient.ID,
+					ClientSecret: "invalid-secret",
+				}
+				_, err := s.createTokenFromClientCredentials(t.Context(), input)
+				require.Error(t, err)
+				require.ErrorIs(t, err, &common.OidcClientSecretInvalidError{})
+			})
+
+			t.Run("Fails with public client", func(t *testing.T) {
```
ItalyPaleAle: nit:
```suggestion
			t.Run("Fails without client secret for public clients", func(t *testing.T) {
```

backend/internal/service/oidc_service.go
savely-krasovsky: All `createToken...` methods currently hardcode it, so yes, I would leave it like that for now. It makes sense to fix it after providing a way to set custom expiration for all tokens.

ItalyPaleAle: Yeah i was just suggesting making it a constant at the top, like we do here: https://github.dev/pocket-id/pocket-id/blob/0b381467ca349f76ea9e8ea638a0b34627fa2e31/backend/internal/service/oidc_service.go#L43-L44 Just to avoid having repeated "magic numbers". But just a nit

kmendell (approved):
LGTM

PR #908, opened 2025-09-02
feat: add CSP header

### LINKED ISSUE #894: 🚀 Feature: hardened CSP headers
### Feature Description

​Add a new section to the Pocket-ID documentation that provides a recommended Content Security Policy (CSP) header configuration for the Pocket-ID login page. This policy will provide a layer of protection against client-side attacks like Cross-Site Scripting (XSS) and data injection. The documentation should explain each directive in the policy and how it contributes to security.
​The documentation should recommend a basic, secure CSP header that can be easily implemented by users running Pocket-ID behind a reverse proxy. This is a common setup in homelabs.

​### Pitch

​This feature request is about defense in depth. While Pocket-ID's core functionality is secure, the login page itself is a potential attack vector. A well-defined CSP adds a critical layer of protection by instructing the user's browser to only load resources from trusted sources.
​
​This is a low-effort, high-reward change. It doesn't require any changes to the Pocket-ID codebase itself. The recommended policy can be added to the documentation, allowing users to configure it on their own reverse proxy (like Nginx, Caddy, or Traefik). By following the example of projects like Authentik, Pocket-ID can provide its users with the tools they need to harden their own self-hosted environments.

stonith404:
Fixes #894

### DIFF
--- a/backend/frontend/frontend_included.go
+++ b/backend/frontend/frontend_included.go
@@ -3,6 +3,7 @@
 package frontend
 
 import (
+	"bytes"
 	"embed"
 	"fmt"
 	"io/fs"
@@ -12,6 +13,7 @@ import (
 	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/pocket-id/pocket-id/backend/internal/middleware"
 )
 
 //go:embed all:dist/*
@@ -27,13 +29,45 @@ func RegisterFrontend(router *gin.Engine) error {
 	fileServer := NewFileServerWithCaching(http.FS(distFS), int(cacheMaxAge.Seconds()))
 
 	router.NoRoute(func(c *gin.Context) {
-		// Try to serve the requested file
 		path := strings.TrimPrefix(c.Request.URL.Path, "/")
-		if _, err := fs.Stat(distFS, path); os.IsNotExist(err) {
-			// File doesn't exist, serve index.html instead
-			c.Request.URL.Path = "/"
+
+		if strings.HasPrefix(path, "api/") {
+			c.JSON(http.StatusNotFound, gin.H{"error": "API endpoint not found"})
+			return
 		}
 
+		// If path is / or does not exist, serve index.html
+		if path == "" {
+			path = "index.html"
+		} else if _, err := fs.Stat(distFS, path); os.IsNotExist(err) {
+			path = "index.html"
+		}
+
+		if path == "index.html" {
+			b, err := fs.ReadFile(distFS, "index.html")
+			if err != nil {
+				_ = c.Error(fmt.Errorf("failed to read index.html: %w", err))
+				return
+			}
+
+			nonce := middleware.GetCSPNonce(c)
+			if nonce != "" {
+				// Replace the first <script> tag with a nonce-bearing tag
+				needle := []byte("<script>")
+				replacement := []byte(fmt.Sprintf("<script nonce=\"%s\">", nonce))
+				b = bytes.Replace(b, needle, replacement, 1)
+			}
+
+			// Do not cache the HTML shell, as it embeds a per-request nonce
+			c.Header("Content-Type", "text/html; charset=utf-8")
+			c.Header("Cache-Control", "no-store")
+			c.Status(http.StatusOK)
+			_, _ = c.Writer.Write(b)
+			return
+		}
+		
+		// Serve other static assets with caching
+		c.Request.URL.Path = "/" + path
 		fileServer.ServeHTTP(c.Writer, c.Request)
 	})
 
--- a/backend/frontend/frontend_included.go
+++ b/backend/frontend/frontend_included.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
+	"io"
 	"io/fs"
 	"net/http"
 	"os"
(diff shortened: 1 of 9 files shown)


backend/internal/middleware/csp_middleware.go
```diff
@@ -0,0 +1,54 @@
...
+		csp := "default-src 'self'; " +
+			"base-uri 'self'; " +
+			"object-src 'none'; " +
+			"frame-ancestors 'none'; " +
+			"form-action 'self'; " +
+			"img-src 'self' data: blob:; " +
+			"font-src 'self'; " +
+			"style-src 'self' 'unsafe-inline'; " +
+			"script-src 'self' 'nonce-" + nonce + "'; " +
+			"connect-src 'self' https://api.github.com"
```
ItalyPaleAle: The call to api.github.com is to check for updates, right? Perhaps that should be done in the backend and proxied (and cached), so we don't have to allowlist this?

backend/internal/middleware/csp_middleware.go
```diff
@@ -0,0 +1,54 @@
...
+		c.Next()
+	}
+}
+
+func generateNonce() string {
+	b := make([]byte, 16)
+	if _, err := rand.Read(b); err != nil {
+		return "" // if generation fails, return empty; policy will omit nonce
+	}
+	return base64.StdEncoding.EncodeToString(b)
```
ItalyPaleAle: - "Raw" = no `=` padding at the end (shorter)
- "URL" = makes this URL-safe

```suggestion
	return base64.RawURLEncoding.EncodeToString(b)
```

backend/frontend/frontend_included.go
```diff
@@ -27,13 +29,45 @@ func RegisterFrontend(router *gin.Engine) error {
...
+				_ = c.Error(fmt.Errorf("failed to read index.html: %w", err))
+				return
+			}
+
+			nonce := middleware.GetCSPNonce(c)
+			if nonce != "" {
+				// Replace the first <script> tag with a nonce-bearing tag
+				needle := []byte("<script>")
+				replacement := []byte(fmt.Sprintf("<script nonce=\"%s\">", nonce))
+				b = bytes.Replace(b, needle, replacement, 1)
```
ItalyPaleAle: Consider using a stream-based transformer, which avoids reading the file in-memory twice (before and after replacement).

For example: https://pkg.go.dev/github.com/icholy/replace#Bytes

The transformer could write directly to `c.Writer`, which implements `io.Writer`

stonith404: Good point but I'm not really sure if it's worth adding a new library just for this use-case. The `index.html` should only be around 2KB so it hasn't really a big memory fingerprint. Would you really recommend to use stream based transformer?

ItalyPaleAle: Not a strict requirement, just a suggestion possibly for the future. Also for reducing pressure on the GC

ItalyPaleAle: Fixed as per discord

backend/internal/service/version_service.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"sync/atomic"
+	"time"
+)
+
+const versionCacheTTL = 15 * time.Minute
```
ItalyPaleAle:
```suggestion
const (
	versionCacheTTL = 15 * time.Minute
	versionCheckURL = "https://api.github.com/repos/pocket-id/pocket-id/releases/latest"
)
```

backend/internal/service/version_service.go
```diff
@@ -0,0 +1,80 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet,
+		"https://api.github.com/repos/pocket-id/pocket-id/releases/latest", nil)
+	if err != nil {
+		return "", fmt.Errorf("create GitHub request: %w", err)
+	}
+
+	httpClient := *s.httpClient
+	httpClient.Timeout = 2 * time.Second
+
+	resp, err := httpClient.Do(req)
```
ItalyPaleAle: Avoiding creating a new HTTP client just to add a timeout:

```suggestion
	reqCtx, cancel := context.WithTimeout(ctx, 5 * time.Second)
	defer cancel()
	req, err := http.NewRequestWithContext(reqCtx, http.MethodGet, versionCheckURL, nil)
	if err != nil {
		return "", fmt.Errorf("create GitHub request: %w", err)
	}

	resp, err := s.httpClient.Do(req)
```

backend/internal/service/version_service.go
```diff
@@ -0,0 +1,80 @@
...
+	s := &VersionService{httpClient: httpClient}
+	s.cache.Store((*cacheEntry)(nil))
+	return s
+}
+
+// GetLatestVersion returns the latest available version from GitHub.
+// It caches the result for a short duration to avoid excessive API calls.
+func (s *VersionService) GetLatestVersion(ctx context.Context) (string, error) {
+	// Serve from cache if fresh
+	if entry, ok := s.cache.Load().(*cacheEntry); ok && entry != nil {
```
ItalyPaleAle: While this will work fine, there's a situation where if multiple calls to GetLatestVersion start at the same time, the request is made multiple (unbounded) times in parallel.

An option could be to keep using an atomic for cache, but then adding a simple semaphore before starting the request. For example (with `s.requesting = make(chan struct{}, 1)`):

```go
// Serve from cache if fresh
// ... same as before

// Try to acquire the lock
// This blocks until another goroutine releases it
s.requesting <- struct{}{}
defer func () {
    // Release the lock
    <-s.requesting
} ()

// Now that we acquired the lock, check the cache again before making the request, as another goroutine may have populated it
// Same logic as above
```

backend/internal/service/version_service.go
```diff
@@ -0,0 +1,80 @@
...
+const versionCacheTTL = 15 * time.Minute
+
+type cacheEntry struct {
+	version     string
+	lastFetched time.Time
+}
+
+type VersionService struct {
+	httpClient *http.Client
+	cache      atomic.Value
```
ItalyPaleAle: Suggest using the generic `atomic.Ptr[*cacheEntry]` which avoids the need to do type casting at runtime

backend/internal/service/version_service.go
```diff
@@ -0,0 +1,80 @@
...
+
+func NewVersionService(httpClient *http.Client) *VersionService {
+	s := &VersionService{httpClient: httpClient}
+	s.cache.Store((*cacheEntry)(nil))
+	return s
+}
+
+// GetLatestVersion returns the latest available version from GitHub.
+// It caches the result for a short duration to avoid excessive API calls.
+func (s *VersionService) GetLatestVersion(ctx context.Context) (string, error) {
```
ItalyPaleAle, line 33: One problem with this method is that if there's an error fetching from GitHub, it will not use the cache.

Suggestion: wrap this in another method that checks the cache. If there's an error fetching from GitHub, return the cached value if any, even if stale.

backend/internal/service/version_service.go
ItalyPaleAle: ( I guess a `sync.Mutex` would work too :) )

stonith404: I mean, it's really unlikely that this will happen, and if it does, it will just request GitHub multiple times, it won't cause a panic, right? I'm just not really sure if it's worth it to add "complexity" just for this edge that just causes multiple requests.

ItalyPaleAle: (I'm not going to say how "highly unlikely" generally happens with a large enough number of users... :) )

The request can take up to 5s, so there's a relatively large window of time where multiple requests can come at once. The biggest problem would be that with multiple requests in parallel, you may hit rate limiting in GitHub, causing responses to fail.

stonith404: 👍 Should be fixed with the cache util I've added.

backend/internal/service/version_service.go
stonith404: Thanks, should be fixed with the caching util as well.

ItalyPaleAle (approved):

PR #909, opened 2025-09-02
feat: allow uppercase usernames

### LINKED ISSUE #906: 🚀 Feature: Allow uppercase usernames
### Feature description

It would be really great if you could add the option to use uppercase usernames. Right now it forces us to use lowercase usernames like 'felitendo', but if i want to use 'Felitendo', then i would have to change the username on every single site that supports it (which are not a lot sadly).

### Pitch

This Feature would help me a lot with user management and I personally think there is no good reason to force all users to use lowercase usernames.

Felitendo:
Allows the use of uppercase usernames like described in this issue: #906 

This PR Adds a new toggle to the Administration Settings called "Allow uppercase usernames". When toggling it, it becomes possible to add/update usernames even when they start with an uppercase letter.

image

This option is disabled by default.

Also sorry for editing the .gitignore file, but i think it will be better to ignore .pnpm-store/ files anyways
fixes #906 

TODO:

- [x] update toggle name to "Allow Uppercase Usernames" to better fit the styling of the other toggles
- [x] add translations
- [x] add tests


Felitendo:
This PR is ready for review :)

ItalyPaleAle (changes requested):
So, the first thing is that by looking at this PR and how things are implemented in the codebase (specifically in the DTO validator: https://github.com/pocket-id/pocket-id/blob/901333f7e43b4e925ed6dfd890dee2caa1947934/backend/internal/dto/validations.go#L16-L19 ), it seems that there's a bug in Pocket ID today:

- The backend allows uppercase characters in usernames (so if someone were to create the user with the APIs, hypothetically, they could already today)
- The frontend requires lowercase
- In the case of users imported from LDAP, we don't do any validation and everything goes

That does need to be fixed, but I believe the **frontend** is correct, and uppercase characters should not be allowed (and LDAP should sanitize usernames too). Instead, I strongly recommend treating all usernames as **case-insensitive** and store them as lowercased in the database.

Allowing uppercase characters (and making the usernames case sensitive like in this PR) can **introduce security vulnerabilities**, both in Pocket ID and in apps that depend on them.

For example, imagine your admin user is named `admin`. All of a sudden, a user creates `ADMIN`. With this PR, they'd be treated as two different users, and to start this could be confusing to users in the Pocket ID admin portal: who's the real admin?

However, it gets worse if you consider that downstream apps may treat usernames in a case-insensitive way, so they may treat both `admin` and `ADMIN` with special privileges. Suddenly, you've introduced an attack vector to downstream apps. Given that most apps treat usernames in a case-insensitive way, this isn't far-fetched.

If we absolutely decide that we do want to allow uppercase characters in the usernames, I would at least implore that we do all comparisons in a case-insensitive way. This would at least disallow the creation of `admin` and `ADMIN` as two separate users, but would add quite a bit of complexity.

> As an aside, I don't think a toggle would be the right solution to this problem because it burdens the user, makes the UI more complex, and makes our testing more complex. It is also global, so if users deploy new apps in the future that don't treat usernames as case-sensitive, there could be security risk above.

ItalyPaleAle:
By the way, would #898 address your underlying problem without touching usernames?

Felitendo:
Is it storing the display name in preferred_username?

Felitendo:
> By the way, would #898 address your underlying problem without touching usernames?

Yeah that would probably be a better idea, but I'm not sure if every app supports changing the preferred_username claim to e.g. display_name... But then for those apps, I could just change them manually. So yep I'm fine with that. I hope that #898 gets merged soon

Felitendo:
#898 would theoretically fix the issue, but only if all apps would allow me to change the username claim from "preferred_username" to something like "display_name". That's sadly not the case for all Apps (like Komodo, Zipline, etc.).

Therefore I think it would be better if we implement a way for usernames to be "in-casesensitive" like you described. So that when somebody created the user "Admin", nobody else can create "admin" or "ADMIN". That would also get rid of the attack vector that you described...

As a sidenote, I completely agree that a toggle in the Admin Settings isn't the most elegant way to handle a case like this, I will remove it in the next commit.

stonith404:
I agree with @ItalyPaleAle, adding an option to allow uppercase usernames just bloats the UI. If we would allow users to customize every single part of the application the settings UI just gets too complicated and Pocket ID won't be "simple" anymore. 

I think the best solution would just be to allow uppercase characters in usernames but don't allow other usernames with the same characters but different casing.

On first glance these two changes are necessary to fix the issue:
- Fix the validation in the frontend by allowing uppercase characters
- Make sure that users can't use the same username as another user just with a different casing

Felitendo:
I completely agree with you @stonith404.
The first change should be already implemented.
Therefore the only thing left is to make usernames "in-casesensitive" and remove the toggle including all of the translations from the UI.

I will add a new commit that adresses all of this sometime this week

ItalyPaleAle:
@Felitendo on a practical sense, my advice would be:

- For Postgres, create a custom case-insensitive collation (note: because we allow ASCII only, build it on top of ASCII not full Unicode) and apply it to the column.
- For SQLite, you should be able to change the column to `text COLLATE NOCASE`

Note: we need to handle the case where users have added users via LDAP that have unicode characters, since we didn't disallow it earlier. We should talk about how to do the migration for them.

Felitendo:
@ItalyPaleAle @stonith404 PR is ready for review again. LMK if i missed anything

ItalyPaleAle:
@Felitendo please see my message above on recommendation on implementation.

In particular, this won't really work:  https://github.com/pocket-id/pocket-id/pull/909/files#diff-0d2aaf48f3d836e224594a0e106533af7c7229f91b6f4c051d13531c9915f670R36-R57

1. First you check for an exact match
2. Then you check for a match for lowercase

This won't work if the passed username is not lowercased, but has different casing. For example if the existing user is "FELITENDO" and you try to add "Felitendo", it will not match anything.

Felitendo:
@ItalyPaleAle will this work?

ItalyPaleAle:
@Felitendo come on, dude... I don't mean to be disprespectful, but I know you know this is a bit of a mess, at least on the backend. Nothing against using AI, I use it too, but please don't just submit AI slop to open source projects, for the maintainers to then have to deal with it...

My recommendation, at least for the backend, would be to start a bit over from scratch.

1. For the LDAP thing, either merge #923 into your PR, wait for that to be merged, or ignore it for now (it's getting fixed soon)
2. For the case-insensitive database comparison, the idea of using a case-insensitive collation for the database row is that then you can perform just simple equality checks (like `WHERE username = ?`) without having to add functions like `lowercase`. Not only this is simpler, but it's a lot faster (because there's an index on `username`, lookups for equality will use the index, while comparisons like `WHERE lowercase(username) = lowercase(?)` perform a full-table scan

Felitendo:
> please don't just submit AI slop to open source projects, for the maintainers to then have to deal with it...

This was not my intention at all, I'm really sorry for all the headache that i caused. I just wanted to get this feature into Pocket ID and I figured the best way to do it is to file a PR.

As for the Recomendations:
1. I think it's better if I wait until #923 gets merged.
2. I don't think I'm able to this alone and I really don't want to submit any more AI slop

I think the best way to proceed forward is for me to clean up the mess that i caused on the backend and for someone else (who actually knows what they are doing) to take over this PR.

kmendell:
@Felitendo 

Don’t be too hard on yourself :) contributing to open source and tackling new features is never easy, and we truly appreciate your initiative and effort in trying to help Pocket ID grow. 

That being said we would love to have you try and keep going with this feature if you feel comfortable, all we ask is you take the feedback given and try to implement it, if you do not know how, just ask we are all human and nobody is perfect :)

If you feel its easier to start over on a newer PR for this that is fine as well; or if you dont feel comforatble continuing we understand that as well.

Felitendo:
Hey, sorry for the late response. I've been thinking about the last few days what the best way to proceed forward would be and came to the conclusion that it would be better if someone is taking over this PR (or creates a new one)...
I initially thought that all it required to have uppercase usernames, is to change 1-2 values and maybe add some translations, but I completely underestimated the work and coding knowledge that actually has to go into this.

And because I really want to get this feature implemented and also to support all of you, I'm willing to sponsor $50 to whover is comfortable working on this.

Thank you to everyone that is working on Pocket ID and appologies again for any issues I've caused.

stonith404:
Closing this in favor of #958.

PR #923, opened 2025-09-07
fix: ensure users imported from LDAP have fields validated

ItalyPaleAle:
Fixes a side issue discovered in #909, where users imported from LDAP did not go through validation rules.

Note: after upgrading, if the database contains LDAP-imported users that fail the validation, those will not be updated anymore and users should see warnings in the logs.

### DIFF
--- a/backend/internal/service/ldap_service.go
+++ b/backend/internal/service/ldap_service.go
@@ -179,10 +179,16 @@ func (s *LdapService) SyncGroups(ctx context.Context, tx *gorm.DB, client *ldap.
 				}
 			}
 
+			username = norm.NFC.String(username)
+			if !dto.ValidateUsername(username) {
+				slog.WarnContext(ctx, "Username is not valid", slog.String("username", username))
+				continue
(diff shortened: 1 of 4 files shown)


backend/internal/service/ldap_service.go
```diff
@@ -179,10 +179,16 @@ func (s *LdapService) SyncGroups(ctx context.Context, tx *gorm.DB, client *ldap.
 				}
 			}
 
+			username = norm.NFC.String(username)
```
stonith404, line 182: Is the username validation here really necessary because it's just a read operation?

ItalyPaleAle: I guess not, I will remove it

stonith404:
Would it make sense to validate the user groups in this case as well?

ItalyPaleAle:
> Would it make sense to validate the user groups in this case as well?

Yes you're right. Done!

stonith404 (approved):

PR #929, opened 2025-09-09
fix: add validation for callback URLs

### LINKED ISSUE #916: 🐛 Bug Report: URLs are not validated
### Reproduction steps

When creating a new client the Callback URLs and the Logout Callback URLs are not validated.

### Expected behavior

I expected the URLs would be validated that they contain a https:// scheme, a domain and a path.

### Actual Behavior

No URL input validation occurs.

### Pocket ID Version

1.10.0

stonith404:
Fixes #916

### DIFF
--- a/backend/internal/dto/validations.go
+++ b/backend/internal/dto/validations.go
@@ -67,8 +67,8 @@ func ValidateClientID(clientID string) bool {
 
 // ValidateCallbackURL validates callback URLs with support for wildcards
 func ValidateCallbackURL(raw string) bool {
-	if raw == "" {
-		return false
+	if raw == "*" {
+		return true
 	}
 
 	// Replace all '*' with 'x' to check if the rest is still a valid URI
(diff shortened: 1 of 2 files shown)


backend/internal/dto/validations.go
```diff
@@ -60,3 +64,24 @@ func ValidateUsername(username string) bool {
 func ValidateClientID(clientID string) bool {
 	return validateClientIDRegex.MatchString(clientID)
 }
+
+// ValidateCallbackURL validates callback URLs with support for wildcards
+func ValidateCallbackURL(raw string) bool {
+	if raw == "" {
```
ItalyPaleAle: Do we need to allowlist `"*"` (allow all callbacks) too? I believe that's a valid one today

stonith404: You're right, thanks

kmendell (approved):

PR #933, opened 2025-09-11
chore: minify background image

ItalyPaleAle:
The current background image is a very large JPEG file that is ove 3MB in size. That is a lot of data, especially for users that are not in the LAN.

This PR converts the background image to a smaller one (1440p) and uses the WebP format (supported by all browsers that support WebAuthn). This reduces the size of the background image to just 292KB. That's a 92% reduction!

Because we did not have code in place to replace existing images only if they were the default, I added some support code that checks if file on disk represent legacy default images and deletes them - but only if they were the default ones (hashes are hardcode in the application)

Tested:

1. A fresh installation writes the webp file
2. If there's an existing background.jpg AND it's the default one (not uploaded by users), it's deleted
3. If there's a user-uploaded background file (no matter the extension), it's unchanged

### DIFF
--- a/backend/internal/bootstrap/application_images_bootstrap_test.go
+++ b/backend/internal/bootstrap/application_images_bootstrap_test.go
@@ -14,7 +14,7 @@ import (
 
 func TestGetBuiltInImageData(t *testing.T) {
 	// Get the built-in image data map
-	builtInImages := getBuiltInImageData()
+	builtInImages := getBuiltInImageHashes()
 
 	// Read the actual images directory from disk
 	imagesDir := filepath.Join("..", "..", "resources", "images")
@@ -34,7 +34,7 @@ func TestGetBuiltInImageData(t *testing.T) {
 		actualFilesMap[fileName] = struct{}{}
 
 		// Check if the file exists in the built-in data
-		builtInData, exists := builtInImages[fileName]
+		builtInHash, exists := builtInImages[fileName]
 		assert.True(t, exists, "File %s exists in images directory but not in getBuiltInImageData map", fileName)
 
 		if !exists {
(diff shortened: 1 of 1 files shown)


kmendell:
I think we should add support for all images to use webp images no? Not just the background image?

ItalyPaleAle:
> I think we should add support for all images to use webp images no? Not just the background image?

All other images are SVGs or ICOs so should be good!

kmendell:
I moreover meant allow uploads of webp images as it seems to be a growing standard, in addition to support the other ones as well. This can be a future enhancement though.

ItalyPaleAle:
Oh, I see now. I didn't try uploading a webp file, and noticed the mime type wasn't in the allowlist. PTAL now, should be fixed. Thanks!

stonith404:
Thanks, the current image is definitely too large. The PR looks good to me but I think the quality of the background image is too low, especially on larger screens. 

My suggestion would be to use a 2160 × 1440 webp image with a quality of 80%. Even though this image would be 400 KB, it almost looks identical to the current 4 MB image in terms of quality. Or do you have another suggestion?

ItalyPaleAle:
Ok changed to 1440p. Using squoosh.app I was able to make it to less than 300KB in size.

I think in the future we could make the image responsive, so smaller screens (like phones) would load a much smaller image. But that would have to be a separate PR.

PS: Also added a test to make sure that we don't forget to update the hashes in the file if we update an image

stonith404:
Thanks, one last thing. Is storing the size really necessary, wouldn't the hash be enough?

ItalyPaleAle:
No it isn't necessary (it was a relic from a previous attempt)

Removed

kmendell (approved):
LGTM, Letting elias do his final look before merging though.

stonith404 (approved):
Looks good to me now, thanks. The `imageData` struct is no longer needed because you've removed the size from it, so I've cleaned it up. If my changes look good to you, feel free to merge it.

PR #940, opened 2025-09-13
fix: key-rotate doesn't work with database storage

ItalyPaleAle:
If the key was already present in the database, key-rotate failed because the row wasn't replaced


kmendell (approved):
Thanks!

PR #954, opened 2025-09-16
fix: make environment variables case insensitive where necessary

### LINKED ISSUE #935: 🐛 Bug Report: APP_URL Case sensitive
### Reproduction steps

APP_URL variable is case sensitive, it might introduce issues in some environments where the url is build from variables, variables used for something else. Not a big issue, it took me 3 hours of troubleshooting but it doesn't make sense either to have this variable case sensitive

If proxydmz_pocketid_name2=Pocket-ID  it wont' work changing it to pocket-id works.... 

      - APP_URL=https://${proxydmz_pocketid_name2}.${mydomain}

I use the same scheme of https://${proxydmz_pocketid_name2}.${mydomain} in several containers and never had this problem

### Expected behavior

.

### Actual Behavior

.

### Pocket ID Version

distroless latest

### Database

default

### OS and Environment

docker

### Log Output

I get this in the logs, probably not related but would be nice some help to troubleshoot it

{"time":"2025-09-11T18:05:20.017Z","level":"ERROR","msg":"Job failed with error","app":"pocket-id","version":"1.10.0","name":"SendHeartbeat","id":"deabacf3-f191-4a24-95ce-a8153233d6f8","error":"heartbeat request failed: request failed with status code: 429"}

stonith404:
Fixes #935

### DIFF
--- a/backend/internal/common/env_config.go
+++ b/backend/internal/common/env_config.go
@@ -32,17 +32,17 @@ const (
 )
 
 type EnvConfigSchema struct {
-	AppEnv             string     `env:"APP_ENV"`
-	LogLevel           string     `env:"LOG_LEVEL"`
-	AppURL             string     `env:"APP_URL"`
-	DbProvider         DbProvider `env:"DB_PROVIDER"`
+	AppEnv             string     `env:"APP_ENV" options:"toLower"`
+	LogLevel           string     `env:"LOG_LEVEL" options:"toLower"`
+	AppURL             string     `env:"APP_URL" options:"toLower"`
+	DbProvider         DbProvider `env:"DB_PROVIDER" options:"toLower"`
 	DbConnectionString string     `env:"DB_CONNECTION_STRING" options:"file"`
 	UploadPath         string     `env:"UPLOAD_PATH"`
 	KeysPath           string     `env:"KEYS_PATH"`
 	KeysStorage        string     `env:"KEYS_STORAGE"`
 	EncryptionKey      []byte     `env:"ENCRYPTION_KEY" options:"file"`
 	Port               string     `env:"PORT"`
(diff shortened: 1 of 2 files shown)


savely-krasovsky:
I would test both cases: env passed in uppercase and env passed in lowercase, but overall LGTM.

backend/internal/common/env_config.go
```diff
@@ -32,17 +32,17 @@ const (
 )
 
 type EnvConfigSchema struct {
-	AppEnv             string     `env:"APP_ENV"`
-	LogLevel           string     `env:"LOG_LEVEL"`
-	AppURL             string     `env:"APP_URL"`
-	DbProvider         DbProvider `env:"DB_PROVIDER"`
+	AppEnv             string     `env:"APP_ENV" options:"toLower"`
+	LogLevel           string     `env:"LOG_LEVEL" options:"toLower"`
+	AppURL             string     `env:"APP_URL" options:"toLower"`
```
ItalyPaleAle, line 37: Wouldn't this be a breaking change if someone has case-sensitive paths?

In a URL:

- The host is case-insensitive
- The path is case-sensitive

Here you're lowercasing the entire URL so this would apply to all of it

https://stackoverflow.com/a/24314935/192024

ItalyPaleAle (changes requested):

backend/internal/common/env_config.go
stonith404: It's not supported to run Pocket ID on a path because of that we can make it case insesitive:
https://github.com/pocket-id/pocket-id/blob/c2c394a05bffacabd6699c36598ad14c3a805786/backend/internal/common/env_config.go#L153

PR #958, opened 2025-09-17
feat: allow uppercase usernames

### LINKED ISSUE #906: 🚀 Feature: Allow uppercase usernames

stonith404:
This PR allows users to use uppercase letters in their usernames. The database index prevents the same username with different cases to avoid username confusion.

Closes #906, #909


stonith404:
@ItalyPaleAle Oh okay that makes sense, thanks. I would use the datatype [`CITEXT`](https://www.postgresql.org/docs/current/citext.html) in this case then. I don't really see any downsides that would be relevant for us if we use this datatype. Or do you have a better solution in mind?

ItalyPaleAle:
Oh I think CITEXT would work great.

Maybe also consider adding `CITEXT COLLATE "C"` which disables Unicode (since we don't allow it in usernames anyways) and simplifies operations (since "lowercasing" is a complex, often unclear, topic in Unicode).

PS: Consider updating some E2E tests to use mixed case usernames and ensure it all works

stonith404:
@ItalyPaleAle Great, thanks. The migrations are now adapted and I've added e2e tests to verify that existing usernames with different casing get rejected.

frontend/src/lib/utils/zod-util.ts
```diff
@@ -26,3 +26,11 @@ export const callbackUrlSchema = z
 			message: m.invalid_redirect_url()
 		}
 	);
+
+export const usernameSchema = z
+	.string()
+	.min(2)
+	.max(30)
+	.regex(/^[a-zA-Z0-9]/, m.username_must_start_with())
+	.regex(/[a-zA-Z0-9]$/, m.username_must_end_with())
```
ItalyPaleAle, line 35: Nit: Any reason why it' can't end with _.- ?

ItalyPaleAle (approved):

Felitendo:
@stonith404 thank you so much for working on this issue.
And I also wanted to take a moment to apologise again to @ItalyPaleAle for the issues i've caused with my last PR. I promise it won't happen again

Thank you guys, you are truly the best :)

kmendell (approved):

ItalyPaleAle:
@Felitendo The message I sent on the other thread was harsher than it should have been, and I want to apologize for that.

While your PR didn't work out, it was helpful that you:

- Brought up the issue
- Proposed an initial solution, which sparked a conversation that highlighted issues with a number of approaches and possible solutions to them

There's an iterative aspect in software development where the first solution is often not the correct one. A lot of the times we realize that there are problems only after it's implemented; hopefully those problems are caught during the code review (like it happened this time). That doesn't mean the initial efforts are bad or useless... on the contrary, they're a learning opportunity for everyone (not just you, but also maintainers).

My only suggestion for the future would be to pay a bit more attention to the feedback provided and, when using AI tools, check the output a bit more closely. We've all been there where the AI just does what it wants and spins in circles sometimes... (and that's why us devs still have a job :) )

PR #961, opened 2025-09-18
feat: pagination improvements

### LINKED ISSUE #825: 🚀 Feature: Pagination improvements
### Feature description

Hello, I'm requesting that table pagination options on all pages could perhaps could become 'sticky' and remembered as an attribute for each admin.

Image

Right now every time a page reloads, you have to reset the pagination to say, show 100 records per page instead of the default.

I also noticed that when using the **< Back** option on various screens, pagination resets - and you are also returned to the _top_ of page 1, instead of where you previously were before editing the item.

For example, If I edit this: 

Image

When I click **< Back**, I am returned to the top of page 1, and pagination is reset to 20. Then I must navigate to the very end of the list to edit the next 'W' app.

### Pitch

Bulk editing OIDC clients is challenging without this feature as the admin is forced to reset pagination and find the page / spot in the list they were at, after editing each one.

stonith404:
This PR introduces several improvements to pagination:
- The “Items per page” setting is now persisted per table in local storage.
- The current page number is stored in the URL as a query parameter. This is useful when navigating back to a list after clicking on an item, since previously the current page was reset.

----

To enable these improvements, data fetching has been moved into the `AdvancedTable` components. Previously, fetching was handled at the page level. Although I generally prefer fetching data as high as possible in the component tree, this is no longer possible because only the `AdvancedTable` component is aware of its pagination state.

Additionally, I had to convert all service methods into arrow functions to avoid issues with `this` binding. For example, consider this code:
https://github.com/pocket-id/pocket-id/blob/3b5bfbf183ae1d4f76c14fba7cef2236a200c36d/frontend/src/routes/settings/admin/users/user-list.svelte#L99

If `userService.list ` were defined as a regular method, the value of `this` inside the function would refer to the `AdvancedTable` component instead of the `UserService` instance. As a result, accessing `this.api` would throw an error: `Cannot read properties of undefined (reading 'get')` because `this.api ` doesn’t exist on `AdvancedTable`.

---
Closes #825


kmendell:
@stonith404 Im looking over this, Hopefully will have a review done this weekend.

kmendell:
The only two points i see that are issues as dicussed in discord: 

1. No defult sorting applied
2. The persisted state is not applied to the filters.

Besides those things this looks good to me :)

stonith404:
@coderabbitai review

PR #965, opened 2025-09-19
fix: decouple images from app config service

### LINKED ISSUE #964: 🐛 Bug Report: Custom background image lost after update to v1.11.1
### Reproduction steps

After updating two instances to v1.11.1 (from v1.10.0) both no longer showed their custom background.  When I went to re-upload the custom background that operation failed with the following error recorded in the log:

`ERR Request with errors: Error #01: remove data/uploads/application-images/background.webp: no such file or directory`

In fact there was not (nor has there ever been for as far as back as I have backups?) a background.webp file in that folder.  Upon creating the file manually only then was I able to successfully re-upload my custom background.

### Expected behavior

To not lose a custom background as a result of an update (didn't see anything in the release notes about it, apologies if missed) and to not be blocked from uploading a custom background due to a file not existing.

### Actual Behavior

Lost custom background, could not re-upload without first manually creating placeholder/default (?) file.

### Pocket ID Version

v1.11.1

### Database

SQLite

### OS and Environment

Pocket ID v1.11.1 running in Docker v28.4.0 on Debian v12 running through Traefik v3.5.2.

### Log Output

`ERR Request with errors: Error #01: remove data/uploads/application-images/background.webp: no such file or directory`

stonith404:
This PR separates the handling of application images (such as the logo and background) from the `AppConfigService`. By decoupling this logic, the codebase becomes cleaner and avoids bugs caused by inconsistencies in file type management.

Currently, Pocket ID stores the file extension of uploaded images in the database at the time of upload. If a user later replaces images directly on the filesystem, the stored extension can fall out of sync with the actual file, preventing the image from being displayed. A similar issue occurred in #964, where users had uploaded a custom JPG image. Since the last release changed the default background image format in `AppConfigService` from JPG to WEBP, Pocket ID incorrectly assumed those users were using a WEBP file instead of their actual JPG, which led to display problems.

---
Closes #964

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1188 +1,1768 @@
-## [1.11.1](https://github.com/pocket-id/pocket-id/compare/v1.10.0...v1.11.1) (2025-09-18)
+## v1.11.1
 
 ### Bug Fixes
 
-- add missing translations ([8c9cac2](https://github.com/pocket-id/pocket-id/commit/8c9cac2655ddbe4872234a1b55fdd51d2f3ac31c))
+- add missing translations([8c9cac2](https://github.com/pocket-id/pocket-id/commit/8c9cac2655ddbe4872234a1b55fdd51d2f3ac31c) by @stonith404)
 
-## [1.11.0](https://github.com/pocket-id/pocket-id/compare/v1.10.0...v1.11.0) (2025-09-18)
+**Full Changelog**: https://github.com/pocket-id/pocket-id/compare/v1.11.0...v1.11.1
 
+## v1.11.0
+
+### Bug Fixes
+
+- update localized name and description of ldap group name attribute ([#892](https://github.com/pocket-id/pocket-id/pull/892) by @kmendell)
+- disable sign up options in UI if `UI_CONFIG_DISABLED`([1d7cbc2](https://github.com/pocket-id/pocket-id/commit/1d7cbc2a4ecf352d46087f30b477f6bbaa23adf5) by @stonith404)
+- ensure users imported from LDAP have fields validated ([#923](https://github.com/pocket-id/pocket-id/pull/923) by @ItalyPaleAle)
+- list items on previous page get unselected if other items selected on next page([6c696b4](https://github.com/pocket-id/pocket-id/commit/6c696b46c8b60b3dc4af35c9c6cf1b8e1322f4cd) by @stonith404)
+- add validation for callback URLs ([#929](https://github.com/pocket-id/pocket-id/pull/929) by @stonith404)
+- key-rotate doesn't work with database storage ([#940](https://github.com/pocket-id/pocket-id/pull/940) by @ItalyPaleAle)
+- make environment variables case insensitive where necessary ([#954](https://github.com/pocket-id/pocket-id/pull/954) by @stonith404)
+- my apps card shouldn't take full width if only one item exists([e7e53a8](https://github.com/pocket-id/pocket-id/commit/e7e53a8b8c87bee922167d24556aef3ea219b1a2) by @stonith404)
 
 ### Features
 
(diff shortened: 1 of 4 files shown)


backend/internal/service/app_images_service.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	mimeType := utils.GetImageMimeType(ext)
+	if mimeType == "" {
+		return "", "", fmt.Errorf("unsupported image type '%s'", ext)
+	}
+
+	imagePath := path.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", name, ext))
```
ItalyPaleAle: Use filepath, or this will have problems on Windows


```suggestion
	imagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", name, ext))
```

backend/internal/service/app_images_service.go
```diff
@@ -0,0 +1,82 @@
...
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	currentExt, ok := s.extensions[imageName]
+	if !ok {
+		return fmt.Errorf("unknown application image '%s'", imageName)
+	}
+
+	imagePath := path.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, fileType))
```
ItalyPaleAle: ditto

```suggestion
	imagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, fileType))
```

backend/internal/utils/file_util.go
```diff
@@ -24,6 +25,15 @@ func GetFileExtension(filename string) string {
 	return filename
 }
 
+// SplitFileName splits a full file name into name and extension.
+func SplitFileName(fullName string) (name, ext string) {
+	dot := strings.LastIndexByte(fullName, '.')
```
ItalyPaleAle: LastIndex is safer in case the file name contains non-ASCII characters

```suggestion
	dot := strings.LastIndex(fullName, ".")
```

backend/internal/utils/file_util.go
```diff
@@ -24,6 +25,15 @@ func GetFileExtension(filename string) string {
 	return filename
 }
 
+// SplitFileName splits a full file name into name and extension.
+func SplitFileName(fullName string) (name, ext string) {
```
ItalyPaleAle, line 29: Use [filepath.Ext](https://pkg.go.dev/path/filepath#example-Ext) ?

backend/internal/service/app_images_service.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if err := utils.SaveFile(file, imagePath); err != nil {
+		return err
+	}
+
+	if currentExt != "" && currentExt != fileType {
+		oldImagePath := path.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, currentExt))
```
ItalyPaleAle:
```suggestion
		oldImagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, currentExt))
```

ItalyPaleAle (commented):
A few of small comments but overall LGTM

For the future, thoughts on storing uploaded images directly in the database? They would then be loaded once the app starts up. It would solve all kinds of issues with keeping the filesystem in sync

backend/internal/utils/file_util.go
ItalyPaleAle: I guess if we need both the name and extension, easier to do it this way

stonith404:
@ItalyPaleAle thanks for the review, Personally I’m not a big fan of storing files in the database especially with SQLite, since it increases the DB size. I prefer keeping data and files separate.

With the changes in this PR, the filesystem is now the single source of truth, and no image related data is stored in the database anymore. This should resolve all the sync issues we had before.

What would make sense though, would be to to implement S3 support in the future, because this seems highly requested and with this feature Pocket ID should be ready to run in a serverless environment.

PR #974, opened 2025-09-22
feat: set pwa display type via environment variable

James18232:
This feature allows the user to set the 'display' type to be used in the app manifest for use when saving as a PWA. The default setting retains the existing "minimal-ui" value. 

based on earlier concepts by @ItalyPaleAle.


stonith404:
Thanks for the contribution. I’m not really convinced that making the display type configurable is the right direction. If we start down that path, we’d likely need to make many other manifest properties configurable as well. That would add unnecessary complexity and bloat to the configuration.

Since the manifest is just a static file, anyone who needs full control can already supply their own by pointing `/app.webmanifest` to a custom file in their reverse proxy. That way you can configure every part of the PWA without adding extra options here.

James18232:
All good, yes proxy is another option. 

imo the display is the only manifest item most users would really be interested in given its impact on the ui and hence worthy of special treatment.

up to y’all.

PR #989, opened 2025-09-29
fix: tokens issued with refresh token flow don't contain groups

### LINKED ISSUE #987: 🐛 Bug Report: initial id token contains `groups` claim but subsequent tokens generated using refresh token don't
### Reproduction steps

NOTE: this started happening with v1.11

When I use `kube oidc-login` with `--oidc-extra-scope=groups` against pocket-id and don't yet have a token, it calls `/authorize?access_type=offline...` and gets an id token that contains my groups, along with a refresh token. The k8s API server authorizes me using the `groups` claim. When that id_token expires and `kubectl oidc-login` tries to get a new one, the new id token doesn't contain my groups and authorization fails.

NOTE: this has nothing to do with Kubernetes. I can verify locally that the id token does not contain the groups.

### Expected behavior

Later id tokens generated with a refresh token also contain the `groups` claim.

### Actual Behavior

Later id tokens generated with a refresh token don't contain the `groups` claim.

### Pocket ID Version

v1.11.2

### Database

SQLite

### OS and Environment

On GCP using v1.12.2-distroless

### Log Output

_No response_

ItalyPaleAle:
Fixes #987

### DIFF
--- a/backend/internal/service/oidc_service.go
+++ b/backend/internal/service/oidc_service.go
@@ -469,7 +469,7 @@ func (s *OidcService) createTokenFromRefreshToken(ctx context.Context, input dto
 	var storedRefreshToken model.OidcRefreshToken
 	err = tx.
 		WithContext(ctx).
-		Preload("User").
+		Preload("User.UserGroups").
 		Where(
 			"token = ? AND expires_at > ? AND user_id = ? AND client_id = ?",
 			utils.CreateSha256Hash(rt),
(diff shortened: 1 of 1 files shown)


backend/internal/service/oidc_service.go
```diff
@@ -1829,7 +1829,7 @@ func (s *OidcService) getUserClaims(ctx context.Context, user *model.User, scope
 		claims["email_verified"] = s.appConfigService.GetDbConfig().EmailsVerified.IsTrue()
 	}
 
-	if slices.Contains(scopes, "groups") {
```
stonith404, line 1824: Does this check make sense? Wouldn't it be better to return an empty list if the user has no groups instead of completely omitting the claim?

ItalyPaleAle: I saw in my tests an empty array being returned and assumed it was an oversight as it felt "unusual". bUt I see your point. Up to you, feel free to revert this line if you prefer

backend/internal/service/oidc_service.go
```diff
@@ -1829,7 +1829,7 @@ func (s *OidcService) getUserClaims(ctx context.Context, user *model.User, scope
...
+	if slices.Contains(scopes, "groups") && len(user.UserGroups) > 0 {
```
stonith404:
```suggestion
	if slices.Contains(scopes, "groups") {
```

Issue #902, opened 2025-08-31
🐛 Bug Report: Federated client credentials RFC 7523 inconsistency


michaelbeaumont:
### Reproduction steps

Try to create an app with federated client credentials for use with Kubernetes as described in the docs.

Try to figure out what the federated client credentials Subject option does.

### Expected behavior

There's no reason to restrict the characters in a client ID, is there?

If we allow `sub != clientID`, we are violating the spec. So we probably can't.

### Actual Behavior

First of all, the docs mention using Kubernetes API server and even describe the format like of those JWT tokens as `system:serviceaccount:envoy-gateway-system:envoy-infra-services-571ea4e6`
but then I get the error `Client ID can only contain letters, numbers, underscores, and hyphens` when trying to create a client with this non-generated clientID.

There also seems to be some inconsistency with clientID vs sub. The RFC says the JWT sub must equal the clientID, but then we allegedly have the Federated Client Credentials optional Subject? ...why is it there? It seems like we both [enforce that the clientID](https://github.com/pocket-id/pocket-id/blob/802754c24c019abb099c0f1b7be45bac315ae1dd/backend/internal/service/oidc_service.go#L1688) is [in the JWT as per the RFC](https://github.com/pocket-id/pocket-id/blob/main/backend/internal/service/oidc_service.go#L1535-L1550) but also [we verify using the optional subject in some cases](https://github.com/pocket-id/pocket-id/blob/802754c24c019abb099c0f1b7be45bac315ae1dd/backend/internal/service/oidc_service.go#L1657).

### Pocket ID Version

v1.10.0

### Database

sqlite

### OS and Environment

Not relevant.

ItalyPaleAle:
I can comment on the willful violation of RFC 7523. Specifically, we have not been complying with this point on section 3:

> For client authentication, the subject MUST be the "client_id" of the OAuth client.

Up until Pocket ID 1.10.0 it was not possible to set a custom Client ID, so complying with this point was not possible.

It also seems that other IdPs are not conforming with this requirement. For example, Entra ID (which the feature was modeled after) doesn't require the subject to match the client ID either (something that wouldn't be possible in Entra ID either): https://learn.microsoft.com/en-us/graph/api/resources/federatedidentitycredentials-overview?view=graph-rest-1.0#set-up-federated-identity-credentials-through-microsoft-graph

I am personally not _too_ concerned with the incompliance with the RFC, given there's precedent in the industry.

[...]

PR #991, opened 2025-09-29
fix: lookup federated client configurations when using RFC 7523 assertions

michaelbeaumont:
See #902. This is more a proposal for one way to solve the problem. In particular, it solves the case where the clientID is left out of the request, since it's optional in the RFC. Unfortunately we don't have a uniqueness constraint on the federatedIdentities across client configurations so we just take the first match.

I was unable to figure out how to write the JSON SQL logic using `gorm`...

### DIFF
--- a/backend/internal/service/oidc_service.go
+++ b/backend/internal/service/oidc_service.go
@@ -1565,32 +1565,63 @@ func (s *OidcService) verifyClientCredentialsInternal(ctx context.Context, tx *g
 	isClientAssertion := input.ClientAssertionType == ClientAssertionTypeJWTBearer && input.ClientAssertion != ""
 
 	// Determine the client ID based on the authentication method
-	var clientID string
 	switch {
 	case isClientAssertion:
 		// Extract client ID from the JWT assertion's 'sub' claim
-		clientID, err = s.extractClientIDFromAssertion(input.ClientAssertion)
+		iss, sub, err := s.extractClientIDFromAssertion(input.ClientAssertion)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


stonith404:
Thanks for your PR.

I'll let @ItalyPaleAle to review your PR because he has implemented federated client credentials. He is currently on vacation so it may take a few weeks.

ItalyPaleAle:
I will not be able to test this for a couple of weeks, I’m traveling and only have an iPad with me.

At a glance it seems like it could work, however we would need to optimize the database. The current query performs a full-table scan and multiple JSON operations on each row, which are very slow. Perhaps creating a index on the result of the functions would help. (See [indexes on expression](https://www.postgresql.org/docs/current/indexes-expressional.html))

ItalyPaleAle:
@michaelbeaumont are you still interested in helping with this PR?

michaelbeaumont:
I'll take a shot at adding indexes one day this week but do feel free to take over.

Just trying a few things out with sqlite since that's my backend. While it's possible to create an index on the `json_extract(..)` expression, this still results in a `SCAN` according `EXPLAIN QUERY PLAN` because, I think, the `WHERE` clause ultimately still has to search through each entry in the `federatedIdentities` array.

michaelbeaumont:
@ItalyPaleAle With postgres, we can add a GIN index on `->federatedIdentities` to make the lookup efficient, apparently.

It's a little trickier with sqlite, as described in my last comment. Is there a way we can optimize for the case of a user having one federated issuer/sub?

PR #994, opened 2025-10-01
feat: add the ability to make email optional

### LINKED ISSUE #867: 🚀 Feature: Make email optional
### Feature description

Right now email is required to make an account, but it really should be optional, because not all apps require the email scope. 

### Pitch

In my use case, I want to run public federated platforms like Matrix and Mastodon. My user base is very privacy conscious, so they do not want to provide an email to make an account.

stonith404:
This PR adds the ability to make the email field optional:
Screenshot 2025-10-01 at 12 10 09

Closes #867


kmendell (approved):
LGTM :)

PR #998, opened 2025-10-02
feat: add CLI command for importing and exporting Pocket ID data

### LINKED ISSUE #721: 🚀 Feature: Export/Import app data
### Feature description

A new section probably in the Application Configuration or a CLI command that allows an admin to export/import a Pocket ID instance's data to/from a file.

### Pitch

One use case would be to export all app data and configuration (including users' data) from an instance using SQLite and then import them into a fresh instance configured to use Postgres.

stonith404:
Closes #721

### DIFF
--- a/backend/internal/cmds/export.go
+++ b/backend/internal/cmds/export.go
@@ -1,0 +1,248 @@
+package cmds
+
+import (
+	"archive/zip"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	"github.com/pocket-id/pocket-id/backend/internal/bootstrap"
+	"github.com/pocket-id/pocket-id/backend/internal/common"
+	"github.com/spf13/cobra"
+	"gorm.io/gorm"
+)
+
+type exportFlags struct {
+	Path string
+}
+
+type databaseJson struct {
+	Provider string                      `json:"provider"`
+	Version  uint                        `json:"version"`
+	Tables   map[string][]map[string]any `json:"tables"`
+}
+
+func init() {
+	var flags exportFlags
+
+	exportCmd := &cobra.Command{
+		Use:   "export",
+		Short: "Exports all data of Pocket ID into a ZIP file",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runExport(flags)
+		},
+	}
+
+	exportCmd.Flags().StringVarP(&flags.Path, "path", "p", "./pocket-id-export.zip", "Path to the ZIP file to export the data to")
+
+	rootCmd.AddCommand(exportCmd)
+}
+
+// runExport orchestrates the export flow
+func runExport(flags exportFlags) error {
+	db, err := bootstrap.NewDatabase()
+	if err != nil {
+		return fmt.Errorf("failed to connect to database: %w", err)
+	}
+
+	dbData, err := extractDatabase(db)
+	if err != nil {
+		return err
+	}
+
+	if err := createExportZip(flags.Path, dbData); err != nil {
+		return err
+	}
+
+	fmt.Printf("Exported data to %s\n", flags.Path)
+	return nil
+}
+
+// extractDatabase reads all tables into a databaseJson struct
+func extractDatabase(db *gorm.DB) (databaseJson, error) {
+	var tables []string
+	switch common.EnvConfig.DbProvider {
+	case common.DbProviderSqlite:
+		if err := db.Raw(`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%' AND name != 'schema_migrations';`).Scan(&tables).Error; err != nil {
+			return databaseJson{}, fmt.Errorf("failed to query sqlite tables: %w", err)
+		}
+
+	case common.DbProviderPostgres:
+		if err := db.Raw(`SELECT tablename FROM pg_tables WHERE schemaname='public' AND tablename != 'schema_migrations';`).Scan(&tables).Error; err != nil {
+			return databaseJson{}, fmt.Errorf("failed to query postgres tables: %w", err)
+		}
+	}
+
+	var version uint
+	if err := db.Raw("SELECT version FROM schema_migrations").Row().Scan(&version); err != nil {
+		return databaseJson{}, fmt.Errorf("failed to query schema version: %w", err)
+	}
+
+	out := databaseJson{
+		Provider: db.Name(),
+		Version:  version,
+		Tables:   map[string][]map[string]any{},
+	}
+
+	for _, table := range tables {
+		if err := dumpTable(db, table, &out); err != nil {
+			return databaseJson{}, err
+		}
+	}
+
+	return out, nil
+}
+
+// dumpTable selects all rows from a table and appends them to out.Tables
+func dumpTable(db *gorm.DB, table string, out *databaseJson) error {
+	rows, err := db.Raw(fmt.Sprintf("SELECT * FROM %s", table)).Rows()
+	if err != nil {
+		return fmt.Errorf("failed to read table %s: %w", table, err)
+	}
+	defer rows.Close()
+
+	cols, _ := rows.Columns()
+	for rows.Next() {
+		vals := make([]any, len(cols))
+		ptrs := make([]any, len(cols))
+		for i := range vals {
+			ptrs[i] = &vals[i]
+		}
+		if err := rows.Scan(ptrs...); err != nil {
+			return fmt.Errorf("failed to scan row in table %s: %w", table, err)
+		}
+
+		rowMap := make(map[string]any, len(cols))
+		for i, col := range cols {
+			rowMap[col] = normalizeForJSON(vals[i])
+		}
+		out.Tables[table] = append(out.Tables[table], rowMap)
+	}
+
+	return rows.Err()
+}
+
+// createExportZip writes the database JSON and file assets into a ZIP archive
+func createExportZip(path string, dbData databaseJson) error {
+	zipFile, err := os.Create(path)
+	if err != nil {
+		return fmt.Errorf("failed to create zip file: %w", err)
+	}
+	defer zipFile.Close()
+
+	zipWriter := zip.NewWriter(zipFile)
+	defer zipWriter.Close()
+
+	// Write DB JSON
+	if err := writeDatabaseJSON(zipWriter, dbData); err != nil {
+		return err
+	}
+
+	// Add filesystem assets (uploads + keys)
+	for _, basePath := range []string{common.EnvConfig.UploadPath, common.EnvConfig.KeysPath} {
+		if err := addDirectoryToZip(zipWriter, basePath); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// writeDatabaseJSON adds database.json to the ZIP archive
+func writeDatabaseJSON(zipWriter *zip.Writer, dbData databaseJson) error {
+	jsonWriter, err := zipWriter.Create("database.json")
+	if err != nil {
+		return fmt.Errorf("failed to create database.json in zip: %w", err)
+	}
+	if err := json.NewEncoder(jsonWriter).Encode(dbData); err != nil {
+		return fmt.Errorf("failed to encode database.json: %w", err)
+	}
+	return nil
+}
+
+// addDirectoryToZip recursively adds all files from basePath into the zip under the correct prefix
+func addDirectoryToZip(zipWriter *zip.Writer, basePath string) error {
+	if basePath == "" {
+		return nil
+	}
+	prefix := "uploads"
+	if basePath == common.EnvConfig.KeysPath {
+		prefix = "keys"
+	}
+
+	return filepath.Walk(basePath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		relPath, err := filepath.Rel(basePath, path)
+		if err != nil {
+			return err
+		}
+		zipPath := filepath.Join(prefix, relPath)
+
+		w, err := zipWriter.Create(zipPath)
+		if err != nil {
+			return fmt.Errorf("failed to create zip entry for %s: %w", zipPath, err)
+		}
+
+		f, err := os.Open(path)
+		if err != nil {
+			return fmt.Errorf("failed to open file %s: %w", path, err)
+		}
+		defer f.Close()
+
+		if _, err := io.Copy(w, f); err != nil {
+			return fmt.Errorf("failed to copy file %s into zip: %w", path, err)
+		}
+		return nil
+	})
+}
+
+// normalizeForJSON ensures DB values round-trip through JSON safely
+func normalizeForJSON(v any) any {
+	switch t := v.(type) {
+	case nil:
+		return nil
+	case []byte:
+		s := string(t)
+		// Try JSON
+		if isJSONObjectOrArray(s) {
+			var j any
+			if err := json.Unmarshal(t, &j); err == nil {
+				return j
+			}
+		}
+		// Try UTF-8 text
+		if utf8.Valid(t) {
+			return s
+		}
+		// Fallback: base64-encode as binary
+		return map[string]any{"__binary__": base64.StdEncoding.EncodeToString(t)}
+
+	case time.Time:
+		return t.UTC().Format(time.RFC3339Nano)
+
+	default:
+		return t
+	}
+}
+
+// isJSONObjectOrArray checks if a string looks like a JSON object or array
+func isJSONObjectOrArray(s string) bool {
+	s = strings.TrimSpace(s)
+	if len(s) < 2 {
+		return false
+	}
+	return (s[0] == '{' && s[len(s)-1] == '}') ||
+		(s[0] == '[' && s[len(s)-1] == ']')
+}
--- a/backend/internal/cmds/import.go
+++ b/backend/internal/cmds/import.go
@@ -1,0 +1,323 @@
+package cmds
+
+import (
+	"archive/zip"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/source/iofs"
+	"github.com/pocket-id/pocket-id/backend/internal/bootstrap"
+	"github.com/pocket-id/pocket-id/backend/internal/common"
+	"github.com/pocket-id/pocket-id/backend/internal/utils"
+	"github.com/pocket-id/pocket-id/backend/resources"
+	"github.com/spf13/cobra"
+	"gorm.io/gorm"
+)
+
+type importFlags struct {
+	Path string
+}
+
+func init() {
+	var flags importFlags
+
+	importCmd := &cobra.Command{
+		Use:   "import",
+		Short: "Imports all data of Pocket ID from a ZIP file",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runImport(flags)
+		},
+	}
+
+	importCmd.Flags().StringVarP(&flags.Path, "path", "p", "./pocket-id-export.zip", "Path to the ZIP file to import the data from")
+
+	rootCmd.AddCommand(importCmd)
+}
+
+// runImport handles the high-level orchestration of the import process
+func runImport(flags importFlags) error {
+	if !askForConfirmation() {
+		fmt.Println("Import aborted.")
+		return nil
+	}
+
+	r, err := zip.OpenReader(flags.Path)
+	if err != nil {
+		return fmt.Errorf("failed to open zip: %w", err)
+	}
+	defer r.Close()
+
+	// Extract metadata and files
+	dbData, err := processZipFiles(r.File)
+	if err != nil {
+		return err
+	}
+
+	if dbData.Provider != string(common.EnvConfig.DbProvider) {
+		return fmt.Errorf(
+			"import file is for %s, but current DB provider is %s",
+			dbData.Provider,
+			common.EnvConfig.DbProvider,
+		)
+	}
+
+	// Connect DB and reset schema
+	db, err := bootstrap.NewDatabase()
+	if err != nil {
+		return err
+	}
+	if err := resetSchema(db); err != nil {
+		return err
+	}
+
+	if err := runMigrations(db, dbData.Version); err != nil {
+		return err
+	}
+
+	if err := insertData(db, dbData); err != nil {
+		return err
+	}
+
+	fmt.Println("Import completed successfully.")
+	return nil
+}
+
+func askForConfirmation() bool {
+	fmt.Println("WARNING: This feature is experimental and may not work correctly. Please create a backup before proceeding and report any issues you encounter.")
+	fmt.Println()
+	fmt.Println("WARNING: Import will erase all existing data at the following locations:")
+	fmt.Printf("Database:      %s\n", absolutePathOrOriginal(common.EnvConfig.DbConnectionString))
+	fmt.Printf("Uploads Path:  %s\n", absolutePathOrOriginal(common.EnvConfig.UploadPath))
+
+	fmt.Printf("Keys Path:     %s\n", absolutePathOrOriginal(common.EnvConfig.KeysPath))
+
+	ok, err := utils.PromptForConfirmation("Do you want to continue?")
+	if err != nil {
+		panic(err)
+	}
+	return ok
+
+}
+
+// processZipFiles extracts database.json and file contents from the ZIP archive
+func processZipFiles(files []*zip.File) (databaseJson, error) {
+	var dbData databaseJson
+
+	for _, f := range files {
+		switch {
+		case f.Name == "database.json":
+			if err := readDatabaseJSON(f, &dbData); err != nil {
+				return dbData, err
+			}
+
+		case strings.HasPrefix(f.Name, "uploads/"):
+			if err := extractIntoBase(f, common.EnvConfig.UploadPath, "uploads/"); err != nil {
+				return dbData, err
+			}
+
+		case strings.HasPrefix(f.Name, "keys/"):
+			if err := extractIntoBase(f, common.EnvConfig.KeysPath, "keys/"); err != nil {
+				return dbData, err
+			}
+		}
+	}
+
+	return dbData, nil
+}
+
+// readDatabaseJSON parses database.json from the ZIP file
+func readDatabaseJSON(f *zip.File, dbData *databaseJson) error {
+	rc, err := f.Open()
+	if err != nil {
+		return fmt.Errorf("failed to open database.json: %w", err)
+	}
+	defer rc.Close()
+
+	if err := json.NewDecoder(rc).Decode(dbData); err != nil {
+		return fmt.Errorf("failed to decode database.json: %w", err)
+	}
+	return nil
+}
+
+// resetSchema clears the DB schema depending on the provider
+func resetSchema(db *gorm.DB) error {
+	switch common.EnvConfig.DbProvider {
+	case common.DbProviderPostgres:
+		return db.Exec(`DROP SCHEMA public CASCADE; CREATE SCHEMA public;`).Error
+
+	case common.DbProviderSqlite:
+		var tables []string
+		if err := db.Raw(`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%';`).
+			Scan(&tables).Error; err != nil {
+			return fmt.Errorf("failed to list SQLite tables: %w", err)
+		}
+
+		for _, t := range tables {
+			if err := db.Exec(fmt.Sprintf(`DROP TABLE IF EXISTS "%s";`, t)).Error; err != nil {
+				return fmt.Errorf("failed to drop SQLite table %s: %w", t, err)
+			}
+		}
+	}
+	return nil
+}
+
+// runMigrations migrates the DB schema to the appropriate version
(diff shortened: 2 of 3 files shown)


stonith404:
@coderabbitai review

kmendell:
@stonith404 Ill pull and test this later tonight and then do a review on it. Sorry for the delay

backend/internal/cmds/export.go
```diff
@@ -0,0 +1,248 @@
...
+	return nil
+}
+
+// writeDatabaseJSON adds database.json to the ZIP archive
+func writeDatabaseJSON(zipWriter *zip.Writer, dbData databaseJson) error {
+	jsonWriter, err := zipWriter.Create("database.json")
+	if err != nil {
+		return fmt.Errorf("failed to create database.json in zip: %w", err)
+	}
+	if err := json.NewEncoder(jsonWriter).Encode(dbData); err != nil {
```
ItalyPaleAle: I would recommend creating a `NewEncoder` with EncodeHTML disables

backend/internal/cmds/export.go
```diff
@@ -0,0 +1,248 @@
...
+			rowMap[col] = normalizeForJSON(vals[i])
+		}
+		out.Tables[table] = append(out.Tables[table], rowMap)
+	}
+
+	return rows.Err()
+}
+
+// createExportZip writes the database JSON and file assets into a ZIP archive
+func createExportZip(path string, dbData databaseJson) error {
```
ItalyPaleAle: Have you considered adding all files into the JSON document, base64-encoded? This simplifies the need to handle ZIP files, which does come with some potential security issues (see "zip bomb"). The base64-encoded data would be bigger, but not sure how much that would matter

PS: You can get the stream created by the JSON encoder and `io.Pipe` it into a gzip stream to compress on-the-fly

backend/internal/cmds/export.go
```diff
@@ -0,0 +1,248 @@
...
+
+// normalizeForJSON ensures DB values round-trip through JSON safely
+func normalizeForJSON(v any) any {
+	switch t := v.(type) {
+	case nil:
+		return nil
+	case []byte:
+		s := string(t)
+		// Try JSON
+		if isJSONObjectOrArray(s) {
```
ItalyPaleAle: Would it be safer to just keep JSON data as-is, as a JSON string?

backend/internal/cmds/export.go
```diff
@@ -0,0 +1,248 @@
...
+			if err := json.Unmarshal(t, &j); err == nil {
+				return j
+			}
+		}
+		// Try UTF-8 text
+		if utf8.Valid(t) {
+			return s
+		}
+		// Fallback: base64-encode as binary
+		return map[string]any{"__binary__": base64.StdEncoding.EncodeToString(t)}
```
ItalyPaleAle: What kind of colums we store today that are binary?

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,371 @@
...
+		return fmt.Errorf("failed to decode database.json: %w", err)
+	}
+	return nil
+}
+
+// resetSchema clears the DB schema depending on the provider
+func resetSchema(db *gorm.DB) error {
+	switch common.EnvConfig.DbProvider {
+	case common.DbProviderPostgres:
+		return db.Exec(`DROP SCHEMA public CASCADE; CREATE SCHEMA public;`).Error
```
ItalyPaleAle: I don't think it's safe to assume the schema name is always "public", isn't it? IIRC the connection string could define a different schema name.

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,371 @@
...
+			if b, err := json.Marshal(val); err == nil {
+				row[k] = string(b)
+			}
+		}
+	}
+}
+
+// extractIntoBase writes a file entry from the ZIP under baseDir, removing the given prefix
+func extractIntoBase(f *zip.File, baseDir, stripPrefix string) error {
+	const maxFileSize = 50 << 20 // 50 MiB is a very generous limit
```
ItalyPaleAle: What if someone has a lot of custom user profile images? I don't think we should put any max limit, since this isn't from a HTTP endpoint anyways

ItalyPaleAle (commented):
Just some initial comments, can’t do a full review (can’t run code right now)

backend/internal/cmds/import.go
stonith404: The limit is per file, so this shouldn't be an issue. The limit per file is just a cheap way to prevent some sorts of ZIP bombs, but a ZIP bomb with many files would still be possible.

backend/internal/cmds/export.go
stonith404: The `PublicKey` and `CredentialID` of `WebauthnCredential`.

backend/internal/cmds/import.go
stonith404: good catch, thanks.

backend/internal/cmds/export.go
stonith404: IMO a ZIP file is a better approach because it makes it easier for users to extract the export contents, for example if they just want to access profile pictures or other files directly.

But yeah ZIP bombs could be a concern. Maybe we should enforce a total uncompressed size limit and ask the user for confirmation if that limit is exceeded before continuing with extraction?

kmendell: I agree that the zip is a better option here, but we should have to enforce some sort of limits with it i think.

kmendell:
@stonith404 you could use dockerode to run the command, a bit of a heavy dependency but it works pretty well.

stonith404:
@coderabbitai review

backend/internal/service/export_service.go
```diff
@@ -0,0 +1,217 @@
...
+	var tables []string
+
+	switch common.EnvConfig.DbProvider {
+	case common.DbProviderSqlite:
+		if err := s.db.Raw(`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%' AND name != 'schema_migrations';`).Scan(&tables).Error; err != nil {
+			return nil, fmt.Errorf("failed to query sqlite tables: %w", err)
+		}
+
+	case common.DbProviderPostgres:
+		if err := s.db.Raw(`SELECT tablename FROM pg_tables WHERE schemaname='public' AND tablename != 'schema_migrations';`).Scan(&tables).Error; err != nil {
```
ItalyPaleAle: Nit, should probably make the schema of pg_tables explicit just to be extra safe
```suggestion
		if err := s.db.Raw(`SELECT tablename FROM pg_catalog.pg_tables WHERE schemaname='public' AND tablename != 'schema_migrations';`).Scan(&tables).Error; err != nil {
```

backend/internal/service/export_service.go
```diff
@@ -0,0 +1,217 @@
...
+		s := string(t)
+		// Try UTF-8 text
+		if utf8.Valid(t) {
+			return s
+		}
+		// Fallback: base64-encode as binary
+		return map[string]any{"__binary__": base64.StdEncoding.EncodeToString(t)}
+
+	case time.Time:
+		println("time")
```
ItalyPaleAle:
```suggestion
		println("time")
```

backend/internal/service/import_service.go
```diff
@@ -0,0 +1,341 @@
...
+			colType = schema[table][col]
+		}
+
+		if strings.Contains(colType, "timestamp") || strings.Contains(colType, "datetime") {
+			switch v := val.(type) {
+			case float64:
+				if v > 1e9 && v < 1e12 {
+					row[col] = time.Unix(int64(v), 0).UTC()
+				} else if v >= 1e12 {
+					row[col] = time.UnixMilli(int64(v)).UTC()
```
ItalyPaleAle: Could we normalize this when reading from the database? Always use milliseconds for example when creating the JSON. This feels a bit brittle IMHO

backend/internal/service/import_service.go
```diff
@@ -0,0 +1,341 @@
...
+	if _, err := io.CopyN(out, rc, int64(f.UncompressedSize64)); err != nil {
+		return fmt.Errorf("copy failed for %s: %w", f.Name, err)
+	}
+	return nil
+}
+
+// togglePostgresForeignKeyChecks enables/disables FK checks in Postgres
+func togglePostgresForeignKeyChecks(tx *gorm.DB, enable bool) error {
+	var tables []string
+	if err := tx.Raw(`SELECT tablename FROM pg_tables WHERE schemaname = 'public'`).
```
ItalyPaleAle:
```suggestion
	if err := tx.Raw(`SELECT tablename FROM pg_catalog.pg_tables WHERE schemaname = 'public'`).
```

backend/internal/service/import_service.go
```diff
@@ -0,0 +1,341 @@
...
+	if strings.HasSuffix(f.Name, "/") || name == "" {
+		return nil // skip directories
+	}
+
+	targetPath := filepath.Join(baseDir, filepath.FromSlash(name))
+	if !strings.HasPrefix(targetPath, filepath.Clean(baseDir)+string(os.PathSeparator)) {
+		return fmt.Errorf("invalid file path: %s", f.Name)
+	}
+
+	_ = os.RemoveAll(targetPath)
```
ItalyPaleAle: Nit: os.Remove instead, since this is not a directory?
```suggestion
	_ = os.Remove(targetPath)
```

pnpm-lock.yaml
```diff
@@ -179,10 +179,17 @@ importers:
         version: 7.0.8(@types/node@22.16.5)(jiti@2.5.1)(lightningcss@1.30.1)(terser@5.43.1)(tsx@4.20.4)(yaml@2.8.1)
 
   tests:
+    dependencies:
```
ItalyPaleAle, line 182: Why was this file updated?

kmendell: @ItalyPaleAle Cause a new dependency was added in the tests workspace.

ItalyPaleAle: duh, my bad, i had the folder collapsed and didn't see the changes

backend/internal/service/import_service.go
stonith404: Yeah we actually already normalize this to seconds only, so handling milliseconds is unnecessary. https://github.com/pocket-id/pocket-id/blob/d61fabbe0d34b2b34435b3f605882278f9ae6e6f/backend/internal/service/export_service.go#L206

backend/internal/common/env_config.go
```diff
@@ -33,31 +33,32 @@ const (
...
+	LocalIPv6Ranges       string     `env:"LOCAL_IPV6_RANGES"`
+	UiConfigDisabled      bool       `env:"UI_CONFIG_DISABLED"`
+	MetricsEnabled        bool       `env:"METRICS_ENABLED"`
+	TracingEnabled        bool       `env:"TRACING_ENABLED"`
+	LogJSON               bool       `env:"LOG_JSON"`
+	TrustProxy            bool       `env:"TRUST_PROXY"`
+	AnalyticsDisabled     bool       `env:"ANALYTICS_DISABLED"`
+	AllowDowngrade        bool       `env:"ALLOW_DOWNGRADE"`
+	InternalAppURL        string     `env:"INTERNAL_APP_URL"`
+	ForcefullyAcquireLock bool       `env:"FORCEFULLY_ACQUIRE_LOCK"`
```
ItalyPaleAle: Should this be an env var or a CLI flag?

IMHO a CLI flag would be more appropriate here because env vars are meant for things that are more "permanent", while forcefully acquire locks should only be done in a break-glass scenario.

I even wonder if there's any reason to have a forceful lock acquisiton outside of the import command. The only scenario when someone would need it if Pocket ID crashed uncleanly, and then the app restarted too quickly. In this case, by the time an admin logs in and forcefully breaks the lock, the lock would have likely expired anyways.

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,278 @@
...
+	host, err := os.Hostname()
+	if err != nil || host == "" {
+		host = "unknown-host"
+	}
+
+	return &AppLockService{
+		db:        db,
+		processID: int64(os.Getpid()),
+		hostID:    host,
+	}
```
ItalyPaleAle, line 48: in a container I believe `os.Hostname()` is just the name of the container, right? So two different containers could have the same hostname. And `os.Getpid` returns the process ID inside the container, which is isolated so I believe it should always be 1 for Pocket ID.

Suggest using a UUID as process ID, which is generated when the app starts, instead of hostname/pid

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,278 @@
...
+		HostID:    s.hostID,
+		ExpiresAt: now.Add(ttl).UnixMilli(),
+	}
+
+	raw, err := value.Marshal()
+	if err != nil {
+		return fmt.Errorf("encode lock value: %w", err)
+	}
+
+	tx := s.db.WithContext(ctx).Begin()
```
ItalyPaleAle: Very much recommend going raw SQL here instead of using transactions. Ideally we want this to be as fast as possible, so it's all done at once without multiple DB roundtrips...

See example queries here: https://chatgpt.com/share/68e79e73-0958-8011-863a-15361b9041a7

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,278 @@
...
+
+// renew tries to renew the lock, retrying up to renewRetries times (sleeping 1s between attempts).
+func (s *AppLockService) renew(ctx context.Context) error {
+	var joinedErr error
+	for attempt := 1; attempt <= renewRetries; attempt++ {
+		now := time.Now()
+		nowMillis := now.UnixMilli()
+		expiresAt := now.Add(ttl).UnixMilli()
+
+		opCtx, cancel := context.WithTimeout(ctx, 31*time.Second)
```
ItalyPaleAle: The timeout should be adjusted. Locks last 30s, and we renew them 20s before they expire, with 3 attempts. This means your timeout can't be longer than 3s, because after that the lock is lost.

stonith404: Oh, this was actually for debugging purposes, but thanks.

backend/internal/common/env_config.go
stonith404: Okay, you are right. I think we can get rid of it completely.

backend/internal/service/app_lock_service.go
stonith404: The hostname of a Docker container is the container ID, so this should be unique. I think storing the hostname and PID makes sense anyways because this helps to identify the admin where the active instance is running.

ItalyPaleAle: The PID would always be 1 inside a container.

True, the default is the container ID in Docker, but it can be overridden with the `--hostname` flag. In K8s, the default hostname is the name of the pod.

I don't think this has enough variability, it's too easy for 2 containers to have the same hostname and defeat the purpose of this conflict detection. For example, a pod with 2 replicas in K8s would have both containers having the same hostname and PID.

backend/internal/service/app_lock_service.go
stonith404: I tried to make every operation atomic but it's not possible inside the `Acquire` function because there is no way to select the `ExpiresAt` of the previous lock in the same query as we are updating the lock in Sqlite. `Acquire` only gets called once on startup so this shouldn't be a big problem.

kmendell (approved):
Overall this looks good ot me. I pulled and did some test locally over the past few days, and everytihng seems to work. Besides Itals comments which most are fixed now i think this is good to go :)

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+			return fmt.Errorf("unsupported database dialect: %s", s.db.Name())
+		}
+
+		res := s.db.WithContext(ctx).Exec(query, raw, lockKey, s.lockID, nowUnix)
+		switch {
+		case res.Error != nil:
+			lastErr = fmt.Errorf("lock renewal failed: %w", res.Error)
+		case res.RowsAffected == 0:
+			// If there isn't a lock with the same lock ID and not expired, we lost the lock
+			lastErr = ErrLockLost
```
ItalyPaleAle: This should be a fatal error, no point in retrying

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+				SET value = $1
+				WHERE key = $2
+				  AND value::json->>'lock_id' = $3
+				  AND ((value::json->>'expires_at')::bigint > $4)
+			`
+		default:
+			return fmt.Errorf("unsupported database dialect: %s", s.db.Name())
+		}
+
+		res := s.db.WithContext(ctx).Exec(query, raw, lockKey, s.lockID, nowUnix)
```
ItalyPaleAle: Should probably set a timeout on this query's context, since we have 10s to retry. So a context with a timeout of 3s should work

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+			)
+			return nil
+		}
+
+		// Wait before next attempt or cancel if context is done
+		if attempt < renewRetries {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(1 * time.Second):
```
ItalyPaleAle, line 290: We don't have much time to waste here, or we'll lose the lock!

```suggestion
			case <-time.After(100 * time.Millisecond):
```

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+	`
+	default:
+		return fmt.Errorf("unsupported database dialect: %s", s.db.Name())
+	}
+
+	res := s.db.WithContext(opCtx).Exec(query, lockKey, s.lockID)
+	if res.Error != nil {
+		return fmt.Errorf("release lock failed: %w", res.Error)
+	}
+
```
ItalyPaleAle, line 211: If `res.RowsAffected` was 0, it means we had lost the lock. May be worth logging this as a warning at least

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+		wait = time.Until(time.Unix(prevLock.ExpiresAt, 0))
+	}
+
+	slog.Info("Acquired application lock",
+		slog.Int64("process_id", s.processID),
+		slog.String("host_id", s.hostID),
+		slog.Duration("wait_before_proceeding", wait),
+	)
+
+	// If we forcefully acquired the lock and there was a previous lock that is not yet expired,
```
ItalyPaleAle: Nit: I think `Acquire` should not block. Instead, return to the caller the time it needs to wait, and let the caller block

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+		return nil
+	}
+	return json.Unmarshal([]byte(raw), lv)
+}
+
+// Acquire obtains the lock. When force is true, the lock is stolen from any existing owner.
+// If the lock is forcefully acquired, it blocks until the previous lock has expired.
+func (s *AppLockService) Acquire(ctx context.Context, force bool) error {
+	var prevLockRaw string
+	err := s.db.WithContext(ctx).Model(&model.KV{}).Where("key = ?", lockKey).Select("value").Scan(&prevLockRaw).Error
```
ItalyPaleAle: If you want this to be a separate query, it needs to be in a transaction for consistency. Or between this SELECT query and the  INSERT one later on, the owner of the lock (and the time it was acquired at) may have changed.

Alternatively, consider doing this in a single statement

backend/internal/service/app_lock_service.go
```diff
@@ -0,0 +1,279 @@
...
+	return &AppLockService{
+		db:        db,
+		processID: int64(os.Getpid()),
+		hostID:    host,
+		lockID:    uuid.NewString(),
+	}
+}
+
+type lockValue struct {
+	ProcessID int64  `json:"process_id"`
```
ItalyPaleAle, line 52: Is this necessary?

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,116 @@
...
+	defer r.Close()
+
+	db, err := bootstrap.ConnectDatabase()
+	if err != nil {
+		return err
+	}
+
+	appLockService := service.NewAppLockService(db)
+	defer func() { //nolint:contextcheck
+		if releaseErr := appLockService.Release(context.Background()); releaseErr != nil {
```
ItalyPaleAle: Should probably have a timeout on the context here

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}()
+
+	opCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	err = appLockService.Acquire(opCtx, flags.ForcefullyAcquireLock)
+	if err != nil {
+		if errors.Is(err, service.ErrLockUnavailable) {
+			return fmt.Errorf("pocket ID must be stopped before importing data; please stop the running instance or run with --forcefully-acquire-lock to terminate the other instance")
```
ItalyPaleAle: Nit:

```suggestion
			return fmt.Errorf("Pocket ID must be stopped before importing data; please stop the running instance or run with --forcefully-acquire-lock to terminate the other instance")
```

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,116 @@
...
+
+	importCmd := &cobra.Command{
+		Use:   "import",
+		Short: "Imports all data of Pocket ID from a ZIP file",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runImport(cmd.Context(), flags)
+		},
+	}
+
+	importCmd.Flags().StringVarP(&flags.Path, "path", "p", "./pocket-id-export.zip", "Path to the ZIP file to import the data from")
```
ItalyPaleAle: To avoid slashes that could be different on Windows:

```suggestion
	importCmd.Flags().StringVarP(&flags.Path, "path", "p", "pocket-id-export.zip", "Path to the ZIP file to import the data from")
```

backend/internal/service/import_service.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// Needs to be called again to re-create the schema_migrations table
+	m, err = utils.GetEmbeddedMigrateInstance(sqlDb)
+	if err != nil {
+		return fmt.Errorf("failed to get migrate instance: %w", err)
+	}
+
+	if err := m.Migrate(targetVersion); err != nil && !errors.Is(err, migrate.ErrNoChange) {
+		// Special case: If no migrations are found, it may be due to a different DB provider.
+		// In that case we just apply all migrations from the current provider.
```
ItalyPaleAle: WOuldn't this be dangerous?

backend/internal/service/import_service.go
```diff
@@ -0,0 +1,332 @@
...
+
+	//nolint:gosec // f.UncompressedSize64 is capped above
+	if _, err := io.CopyN(out, rc, int64(f.UncompressedSize64)); err != nil {
+		return fmt.Errorf("copy failed for %s: %w", f.Name, err)
+	}
+	return nil
+}
+
+// togglePostgresForeignKeyChecks enables/disables FK checks in Postgres
+func togglePostgresForeignKeyChecks(tx *gorm.DB, enable bool) error {
```
ItalyPaleAle: Nit: technically, this does a `set` not a `toggle`

backend/internal/service/import_service.go
```diff
@@ -0,0 +1,332 @@
...
+	return nil
+}
+
+// toggleSqliteForeignKeyChecks enables/disables FK checks in SQLite
+func toggleSqliteForeignKeyChecks(db *gorm.DB, enable bool) error {
+	value := "OFF"
+	if enable {
+		value = "ON"
+	}
+	return db.Exec(fmt.Sprintf("PRAGMA foreign_keys = %s;", value)).Error
```
ItalyPaleAle:
```suggestion
	return db.Exec("PRAGMA foreign_keys = " + value + ";").Error
```

backend/internal/service/export_service.go
```diff
@@ -0,0 +1,224 @@
...
+	var version uint
+	if err := s.db.Raw("SELECT version FROM schema_migrations").Row().Scan(&version); err != nil {
+		return 0, fmt.Errorf("failed to query schema version: %w", err)
+	}
+	return version, nil
+}
+
+// dumpTable selects all rows from a table and appends them to out.Tables
+func (s *ExportService) dumpTable(table string, out *DatabaseExport) error {
+	rows, err := s.db.Raw(fmt.Sprintf("SELECT * FROM %s", table)).Rows()
```
ItalyPaleAle:
```suggestion
	rows, err := s.db.Raw("SELECT * FROM " + table).Rows()
```

backend/internal/service/export_service.go
```diff
@@ -0,0 +1,224 @@
...
+			ptrs[i] = &vals[i]
+		}
+		if err := rows.Scan(ptrs...); err != nil {
+			return fmt.Errorf("failed to scan row in table %s: %w", table, err)
+		}
+
+		rowMap := make(map[string]any, len(cols))
+		for i, col := range cols {
+			sqlType := colTypes[i].DatabaseTypeName()
+			rowMap[col] = normalizeForJSON(vals[i], sqlType)
```
ItalyPaleAle: Since we know `ColumnTypes`, could we be smarter about that and actually read into `string`s and `[]byte`s? I am not a big fan of the serialization logic, especially when we check if the string is valid UTF-8... it feels brittle.

ItalyPaleAle (changes requested):
Looking very good, just some more feedback

backend/internal/service/app_lock_service.go
stonith404: Don't we have 10 seconds to renew the lock? TTL is 30 seconds and the renew interval 20 seconds. IMO loosing 3 seconds because of retries is ok.

backend/internal/service/app_lock_service.go
stonith404: Including the host and process ID in the lock metadata makes it easier for admins to diagnose lock issues. These values show up in slog logs so an admin can quickly see which instance of Pocket ID is holding the lock.

backend/internal/service/app_lock_service.go
ItalyPaleAle: The problem is that the query may not be instant. If we set a [3s timeout on the context](https://github.com/pocket-id/pocket-id/pull/998/files#diff-a07f3168d8ae71eb0b92cf94eb1c0f934c7b376bdf7053114b014f7fd52cfe66R259), the query can take up to 3s

backend/internal/cmds/export.go
```diff
@@ -0,0 +1,60 @@
...
+
+	storage, err := bootstrap.InitStorage(ctx, db)
+	if err != nil {
+		return fmt.Errorf("failed to initialize storage: %w", err)
+	}
+
+	exportService := service.NewExportService(db, storage)
+
+	var w *os.File
+	w, err = os.Create(flags.Path)
```
ItalyPaleAle: Can you add support for exporting to stdout? If `--path -` (a dash is often indicating stdout), then write to stdout (`io.Stdout`).

This is very helpful for integration with backup tools, for example. For example, I back up my Postgres database with Restic with a command similar to:

```sh
pg_dump $db \
  | gzip \
  | restic backup --verbose --stdin --stdin-filename "postgres/$db.sql.gz"
```

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,129 @@
...
+
+	importCmd := &cobra.Command{
+		Use:   "import",
+		Short: "Imports all data of Pocket ID from a ZIP file",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runImport(cmd.Context(), flags)
+		},
+	}
+
+	importCmd.Flags().StringVarP(&flags.Path, "path", "p", "pocket-id-export.zip", "Path to the ZIP file to import the data from")
```
ItalyPaleAle: Likewise, suggest adding support for reading from stdin, when `--path -`

backend/internal/cmds/import.go
```diff
@@ -0,0 +1,129 @@
...
+
+// runImport handles the high-level orchestration of the import process
+func runImport(ctx context.Context, flags importFlags) error {
+	if !flags.Yes {
+		ok, err := askForConfirmation()
+		if err != nil {
+			return fmt.Errorf("failed to get confirmation: %w", err)
+		}
+		if !ok {
+			fmt.Println("Aborted")
```
ItalyPaleAle, line 54: Should probably exit with non-0 status code?

```suggestion
			fmt.Println("Aborted")
            os.Exit(1)
```

scripts/docker/entrypoint.sh
```diff
@@ -22,17 +22,17 @@ fi
...
         existing_user=$(getent passwd "$PUID" | cut -d: -f1)
         echo "Using existing user: $existing_user"
     fi
 fi
 
-# Change ownership of the /app/data directory
+# Change ownership of the /app directory
 mkdir -p /app/data
-find /app/data \( ! -group "${PGID}" -o ! -user "${PUID}" \) -exec chown "${PUID}:${PGID}" {} +
+find /app \( ! -group "${PGID}" -o ! -user "${PUID}" \) -exec chown "${PUID}:${PGID}" {} +
```
ItalyPaleAle: Why this change? Wouldn't it also change the permissions of the binary? Which means - I'm not sure this will work with a read-only root FS

stonith404: The reason was because the export command exports the ZIP into the same folder as the binary by default, and this fails inside Docker because the `pocket-id` user can't write to the `/app` folder. 

But I think it makes more sense if we just document that you have to set the export path to `/app/data` if you are using Docker. I've reverted the change.

ItalyPaleAle: Now that you added stdin/stdout, that would be even better for docker usage

ItalyPaleAle (approved):
Phew, this was quite the effort!

stonith404:
True 😅

PR #1006, opened 2025-10-06
fix: mark any callback url as valid if they contain a wildcard

### LINKED ISSUE #1005: 🐛 Bug Report: validation for callback URLs too strict for port
### Reproduction steps

Add a callback URL such as `http://localhost:*`

### Expected behavior

The placeholder `*` should be valid for the port part of the URL.

#929 added a validation. Unfortunately the implementation simply replaces '*' with '+' and checks the URI syntax, which fails in the port part.

### Actual Behavior

Input is rejected with 'Invalid redirect URL? 

### Pocket ID Version

1.12.0

### Workaround
Use `http://localhost*`, but it offers less restriction than `http://localhost:*`

stonith404:
Fixes #1005

### DIFF
--- a/backend/internal/dto/validations.go
+++ b/backend/internal/dto/validations.go
@@ -67,14 +67,12 @@ func ValidateClientID(clientID string) bool {
 
 // ValidateCallbackURL validates callback URLs with support for wildcards
 func ValidateCallbackURL(raw string) bool {
-	if raw == "*" {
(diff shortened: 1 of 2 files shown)


kmendell (approved):
LGTM Pulled at tested earlier, this is working for me no major issues that i could encounter or find.

PR #1008, opened 2025-10-06
fix: uploading a client logo with an URL fails

CzBiX:
fix issue introduced by ad8a90c839cc79b542b60ae66c7eb9254fa5f3e4


kmendell:
We just need ot fix that lintyer error about proper formatting of the go code. If you can do that quick awesome if not ill do it in a little bit once i have time.

PR #1012, opened 2025-10-07
feat: RFC 8252 loopback interface redirection support

savely-krasovsky:
Support for special Callback URL case which is primary used by Desktop Native Applications.
https://datatracker.ietf.org/doc/html/rfc8252#section-7.3

Until recently it was possible to set `http://127.0.0.1:*` and `http://[::1]:*` (or `http://localhost:*`), now it's impossible due to the new form validation:

image

But still it presets a huge vulnerability, since `http://localhost:*` basically allows to craft authorization URL with callback URL like this: `http://localhost:password@malicious-host.com/authorization_code_stealer`.

Fortunately RFC 8252 defines this special case by specifying that ephemeral port could be completely random and authorization server should allow any arbitrary por.


kmendell:
@savely-krasovsky im marking this as draft for now

savely-krasovsky:
After this https://github.com/pocket-id/pocket-id/commit/078152d4dbb05dd027ff323f39d090ecb67927c7 should be resolved.

PR #1047, opened 2025-10-24
fix: use constant time comparisons when validating PKCE challenges

ItalyPaleAle:
This fixes a small potential timing attack


stonith404:
Interesting, I didn't know that timing attacks are possible with "normal" string comparison. Thanks!

stonith404 (approved):

PR #1054, opened 2025-10-25
fix: Prevent blinding FOUC in dark mode

mufeedali:
Pocket ID is currently one of the many self-hosted services that I use that flashbang me on launching them in a dark environment. Aside from the FOUC, dark mode is really well-supported in Pocket ID, especially as of the last release.

### DIFF
--- a/backend/frontend/frontend_included.go
+++ b/backend/frontend/frontend_included.go
@@ -32,10 +32,6 @@ func init() {
 		panic(fmt.Errorf("failed to read index.html: %w", iErr))
 	}
 
-	// Get the position of the first <script> tag
-	idx := bytes.Index(index, []byte(scriptTag))
-
-	// Create writeIndexFn, which adds the CSP tag to the script tag if needed
 	writeIndexFn = func(w io.Writer, nonce string) (err error) {
 		// If there's no nonce, write the index as-is
 		if nonce == "" {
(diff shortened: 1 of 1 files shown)


frontend/src/app.html
```diff
@@ -6,7 +6,17 @@
...
+		<link rel="apple-touch-icon" href="/img/static-logo-512.png" />
+		<script>
+			(function () {
+				const mode = localStorage.getItem('mode-watcher-mode') || 'system';
+				const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+				const isDark = mode === 'dark' || (mode === 'system' && prefersDark);
+
+				document.documentElement.classList.toggle('dark', isDark);
+				document.documentElement.style.colorScheme = isDark ? 'dark' : 'light';
+			})();
```
stonith404:
```suggestion
		const mode = localStorage.getItem('mode-watcher-mode') || 'system';
		const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
		const isDark = mode === 'dark' || (mode === 'system' && prefersDark);

		document.documentElement.classList.toggle('dark', isDark);
		document.documentElement.style.colorScheme = isDark ? 'dark' : 'light';
```

Wrapping a function around isn't necessary, right?

mufeedali: True, my bad. I've wrapped it in a try-catch instead because I'm guessing the test is failing due to localStorage.

stonith404:
Thanks for the changes! I can look into the e2e tests tomorrow.

mufeedali:
Sorry, I looked into it but couldn't get the container to build locally and my initial guess turned out to be wrong. :frowning_face:

mufeedali:
Thank you!

Issue #403, opened 2025-03-30
🐛 Bug Report: Passkey doesn't get saved on Firefox Mobile, Android 13


000yesnt:
### Reproduction steps

After adding a new passkey through Firefox Mobile on Android 13, log out and try to log back in.

### Expected behavior

Android shows me that I have a passkey available; the passkey appears on Google's Password Manager. Example video recorded on Chrome:

https://github.com/user-attachments/assets/7deb1d72-b658-41ac-9048-9272e0f31c80

### Actual Behavior

Android says I have no passkey; passkey doesn't appear on Google Password Manager

https://github.com/user-attachments/assets/8736e50d-1048-414a-977a-ad15aa081b3d

### Version and Environment

Pocket ID 0.44.0 (the demo version at time of report)
Samsung Galaxy M52 5G, Android 13, One UI 5.1
Firefox 136.0.2

Passkeys do work on Firefox. Here's a recording of me successfully using one on webauthn.io

https://github.com/user-attachments/assets/7a85e909-9ee7-4a48-bf33-d3d53d7f515d

I have to create a passkey from Chrome to be able to use it in Firefox.

### Log Output

Replicating the issue on my server:

```
pocket-id-1  | [GIN] 2025/03/30 - 07:44:17 | 200 |   21.086306ms |    192.168.1.72 | GET      "/api/webauthn/register/start"
pocket-id-1  | [GIN] 2025/03/30 - 07:44:17 | 200 |   21.140303ms |    192.168.1.72 | GET      "/api/webauthn/register/start"
pocket-id-1  | [GIN] 2025/03/30 - 07:44:20 | 200 |    26.06538ms |    192.168.1.72 | POST     "/api/webauthn/register/finish"
pocket-id-1  | [GIN] 2025/03/30 - 07:44:20 | 200 |   26.116044ms |    192.168.1.72 | POST     "/api/webauthn/register/finish"
pocket-id-1  | [GIN] 2025/03/30 - 07:44:20 | 200 |     791.412µs |    192.168.1.72 | GET      "/api/webauthn/credentials"
pocket-id-1  | [GIN] 2025/03/30 - 07:44:20 | 200 |     863.008µs |    192.168.1.72 | GET      "/api/webauthn/credentials"
[...]

PR #1055, opened 2025-10-26
fix: use credProps to save passkey on firefox android

lhoursquentin:
Fixes https://github.com/pocket-id/pocket-id/issues/403

Main issue:

When creating a new passkey from firefox android (same with firefox nightly) the key isn't actually saved in Google password manager. No error is displayed, public key is created in pocket id, but private key is lost. note: this happens without any extension installed (no bitwarden or similar can intervene).

Investigation:

As mentioned in the comments from the issue, the https://webauthn.io/ demo can save the passkey from firefox android. I started to compare the options returned for the passkey registration and saw one main difference between the two: `credProps: true` in the `extensions` object.

This suspicion was confirmed using https://www.webauthn.me/debugger where by default passkeys weren't saved using firefox android like on pocket-id, but was fixed by enabling the `credProps` extension.

This adds the relevant code to enable this extension and takes inspiration (or rather copy paste) from go-webauthn passkey example code:
https://github.com/go-webauthn/webauthn/blob/a664f002fd15d5b7393a19bb325a58df2633326b/webauthn/example_passkey_test.go#L98

(built pocket-id locally and confirmed the fix)

### DIFF
--- a/backend/internal/service/webauthn_service.go
+++ b/backend/internal/service/webauthn_service.go
@@ -81,6 +81,7 @@ func (s *WebAuthnService) BeginRegistration(ctx context.Context, userID string) 
 		&user,
 		webauthn.WithResidentKeyRequirement(protocol.ResidentKeyRequirementRequired),
 		webauthn.WithExclusions(user.WebAuthnCredentialDescriptors()),
+		webauthn.WithExtensions(map[string]any{"credProps": true}), // Required for Firefox Android to properly save the key in Google password manager
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to begin WebAuthn registration: %w", err)


stonith404:
Interesting thanks. I'm not really sure why adding this property fixes the issue because as far as I understand this just tells the client, that it should give Pocket ID some more information about the passkey. 

But you can confirm that this fixes the issue with Pocket ID, right? This PR is available as a Docker image if you want to test it on your server `ghcr.io/pocket-id/pocket-id:pr-1055`.

lhoursquentin:
> I'm not really sure why adding this property fixes the issue because as far as I understand this just tells the client, that it should give Pocket ID some more information about the passkey.

Agreed, reading the spec I don't understand either why it would have an impact, I can only guess that firefox android implementation is incorrect.

Still I was able to verify that adding this extension did solve the issue on both pocket-id & https://www.webauthn.me/debugger.

> This PR is available as a Docker image if you want to test it on your server ghcr.io/pocket-id/pocket-id:pr-1055.

I had already built one locally to confirm the fix but just double checked with this one as well, it works :) !

stonith404:
Okay great thanks. Yeah it's probably an incorrect implementation of Firefox but as adding this property doesn't have a negative impact we can easily add it.

PR #1061, opened 2025-10-29
feat: add ability to set default profile picture

### LINKED ISSUE #1018: 🚀 Feature: Allow setting a custom default profile picture
### Feature description

Being able to set an image that will be used as the default profile picture for new users.

### Pitch

I use Pocket ID as the authentication layer for my families private network and my businesses internal network. Some parts of my family aren't very tech savvy and don't tend to change their profile picture and leave it as the default, instead of updating their profile picture to match the networks brand on every new user created, I would love a way to set an image that is used as the default instead.

stonith404:
This PR adds the functionality for adding a custom default profile picture. If not set, Pocket ID will generate a profile picture with the initials of the user.

Closes #1018

### DIFF
--- a/backend/internal/controller/app_images_controller.go
+++ b/backend/internal/controller/app_images_controller.go
@@ -188,7 +188,7 @@ func (c *AppImagesController) getImage(ctx *gin.Context, name string) {
 	ctx.File(imagePath)
 }
 
-// updateBackgroundImageHandler godoc
+// updateDefaultProfilePicture godoc
 // @Summary Update default profile picture image
 // @Description Update the default profile picture image
 // @Tags Application Images
--- a/backend/internal/service/app_images_service.go
+++ b/backend/internal/service/app_images_service.go
@@ -88,6 +88,9 @@ func (s *AppImagesService) DeleteImage(imageName string) error {
 }
 
 func (s *AppImagesService) IsDefaultProfilePictureSet() bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
 	_, ok := s.extensions["default-profile-picture"]
 	return ok
 }


backend/internal/controller/app_images_controller.go
```diff
@@ -171,3 +187,41 @@ func (c *AppImagesController) getImage(ctx *gin.Context, name string) {
...
+
+	if err := c.appImagesService.UpdateImage(file, "default-profile-picture"); err != nil {
+		_ = ctx.Error(err)
+		return
+	}
+
+	ctx.Status(http.StatusNoContent)
+}
+
+// / deleteDefaultProfilePicture godoc
```
ItalyPaleAle: Nit
```suggestion
// deleteDefaultProfilePicture godoc
```

backend/internal/service/app_images_service.go
```diff
@@ -69,13 +69,36 @@ func (s *AppImagesService) UpdateImage(file *multipart.FileHeader, imageName str
...
+func (s *AppImagesService) DeleteImage(imageName string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	ext, ok := s.extensions[imageName]
+	if !ok || ext == "" {
+		return &common.ImageNotFoundError{}
+	}
+
+	imagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, ext))
```
ItalyPaleAle: nit
```suggestion
	imagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", imageName + "." + ext)
```

backend/internal/service/user_service.go
```diff
@@ -183,6 +199,14 @@ func (s *UserService) UpdateProfilePicture(userID string, file io.Reader) error
 		return err
 	}
 
+	err = s.db.Model(&model.User{}).
```
ItalyPaleAle: Do we need to use the database? Couldn't we just check if there's a file on disk?

backend/resources/migrations/postgres/20251029133312_default_profile_picture.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE users ADD COLUMN has_custom_profile_picture BOOLEAN NOT NULL DEFAULT FALSE;
```
ItalyPaleAle: Don't we need to set this to TRUE for users who have a profile picture set before the upgrade?

backend/internal/service/user_service.go
kmendell: In order to set the has custom profile picture value in the database? 

Maybe im missing something, but this makes sense to me.

ItalyPaleAle: My question is whether we need this as a column in the database. If all we need is to check if a user has a custom profile picture, can't we just look if a file exists in the filesystem?

stonith404: Thanks, I made a mistake in my thinking, it's not necessary to store it in the database we can just check the path.

stonith404:
@coderabbitai review

PR #1071, opened 2025-11-02
fix: replace %lang% placeholder in html lang

daimond113:
Replaces the `%lang%` placeholder in the `lang` attribute of the `<html>` tag, as per [Paraglide's documentation](https://inlang.com/m/gerre34r/library-inlang-paraglideJs/sveltekit#add-the-paraglidemiddleware-to-srchooksserverts).


stonith404:
Thanks for your contribution. Your solution unfortunately won't work because in production we export the web application to a SPA which gets served by the backend and therefore we can't run code on server side in Sveltekit.

daimond113:
I believe this should work now.

PR #1080, opened 2025-11-07
feat: add support for S3 storage backend

### LINKED ISSUE #497: 🚀 Feature: Store Data in S3
### Feature description

I like to run this container stateless (with an external database/postgresql) and storage somewhere else (S3).

### Pitch

Store all data (exclude sqlite) in a S3 Storage (like uploads/avatars) and the keys maybe into the database or also into S3

stonith404:
This PR adds S3 support, which was the final step needed to run Pocket ID in a serverless environment. The implementation is done with a `FileStorage` interface. Based on the configured environment variables Pocket ID chooses between the filesystem or S3 implementation. 

A nice advantage of the new file handling is that we can reduce the risk of path traversal attacks because `filesystemStorage` uses `os.OpenRoot` which only allows file system access in the defined directory which is `UPLOAD_PATH`.

Closes #497

### DIFF
--- a/tests/setup/docker-compose-s3.yml
+++ b/tests/setup/docker-compose-s3.yml
@@ -17,6 +17,7 @@ services:
     environment:
       AWS_ACCESS_KEY_ID: test
       AWS_SECRET_ACCESS_KEY: test
+      AWS_DEFAULT_REGION: us-east-1
     depends_on:
       localstack-s3:
         condition: service_healthy


stonith404:
@coderabbitai review

tests/setup/docker-compose-s3.yml
```diff
@@ -0,0 +1,46 @@
+services:
+  lldap:
+    extends:
+      file: docker-compose.yml
+      service: lldap
+
+  minio:
+    image: quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
```
ItalyPaleAle: Strongly suggest using [garage](https://garagehq.deuxfleurs.fr/)

The minio container image is [unmaintained](https://news.ycombinator.com/item?id=45665452), and the current one has security vulnerabilities.

backend/internal/storage/storage.go
```diff
@@ -0,0 +1,56 @@
...
+
+type FileStorage interface {
+	Save(ctx context.Context, relativePath string, data io.Reader) error
+	Open(ctx context.Context, relativePath string) (io.ReadCloser, int64, error)
+	Delete(ctx context.Context, relativePath string) error
+	DeleteAll(ctx context.Context, prefix string) error
+	List(ctx context.Context, prefix string) ([]ObjectInfo, error)
+}
+
+type Config struct {
```
ItalyPaleAle: Thoughts on having a separate Config for S3 and filesystem? So if in the future new backends are added (Database, Azure Blob...), this doesn't become overloaded

FWIW I think a database provider would be the logical next step, since it's a lot simpler than S3 if someone doesn't already have a place for S3

backend/internal/storage/s3.go
```diff
@@ -0,0 +1,155 @@
...
+				Size: aws.ToInt64(obj.Size),
+			})
+		}
+	}
+	return objects, nil
+}
+
+func (s *s3Storage) buildObjectKey(p string) string {
+	p = filepath.Clean(p)
+	p = filepath.ToSlash(p)
```
ItalyPaleAle, line 162: I think this needs to use `path` and not `filepath`? Because `filepath` on Windows uses `\`

backend/internal/storage/s3.go
```diff
@@ -0,0 +1,155 @@
...
+	"io/fs"
+	"path/filepath"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awscfg "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/aws/smithy-go"
```
ItalyPaleAle, line 17: Seems that we should not be using this package, see docs

backend/internal/utils/image/profile_picture.go
```diff
@@ -19,25 +19,21 @@ import (
...
-	}()
-
-	return pr, nil
+	var buf bytes.Buffer
+	err = imaging.Encode(&buf, img, imaging.PNG)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode image: %w", err)
+	}
+
+	return bytes.NewReader(buf.Bytes()), nil
```
ItalyPaleAle, line 36: Can you maintain the code that uses `io.Pipe`, so it avoids buffering in memory?

backend/internal/job/file_cleanup_job.go
```diff
@@ -44,29 +44,24 @@ func (j *FileCleanupJobs) clearUnusedDefaultProfilePictures(ctx context.Context)
...
 	}
 
-	defaultPicturesDir := common.EnvConfig.UploadPath + "/profile-pictures/defaults"
-	if _, err := os.Stat(defaultPicturesDir); os.IsNotExist(err) {
-		return nil
-	}
-
-	files, err := os.ReadDir(defaultPicturesDir)
+	defaultPicturesDir := path.Join("profile-pictures", "defaults")
+	files, err := j.fileStorage.List(ctx, defaultPicturesDir)
```
ItalyPaleAle, line 55: Would this also clean up old temporary files?

The filesystem storage does "write then copy", but if the process crashes half-way, there would be leftover files on disk.

backend/internal/service/app_images_service.go
```diff
@@ -51,15 +61,20 @@ func (s *AppImagesService) UpdateImage(file *multipart.FileHeader, imageName str
...
 
-	if err := utils.SaveFile(file, imagePath); err != nil {
+	if err := s.storage.Save(context.Background(), imagePath, fileReader); err != nil {
 		return err
 	}
 
 	if currentExt != "" && currentExt != fileType {
-		oldImagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, currentExt))
-		if err := os.Remove(oldImagePath); err != nil && !os.IsNotExist(err) {
+		oldImagePath := path.Join("application-images", fmt.Sprintf("%s.%s", imageName, currentExt))
```
ItalyPaleAle:
```suggestion
		oldImagePath := path.Join("application-images", imageName + "." + currentExt)
```

backend/internal/service/app_images_service.go
```diff
@@ -51,15 +61,20 @@ func (s *AppImagesService) UpdateImage(file *multipart.FileHeader, imageName str
...
-	imagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", fmt.Sprintf("%s.%s", imageName, fileType))
+	imagePath := path.Join("application-images", imageName+"."+fileType)
+	fileReader, err := file.Open()
+	if err != nil {
+		return err
+	}
+	defer fileReader.Close()
 
-	if err := utils.SaveFile(file, imagePath); err != nil {
+	if err := s.storage.Save(context.Background(), imagePath, fileReader); err != nil {
```
ItalyPaleAle: Shouldn't this have a context passed by the caller?

backend/internal/service/app_images_service.go
```diff
@@ -78,8 +93,8 @@ func (s *AppImagesService) DeleteImage(imageName string) error {
 		return &common.ImageNotFoundError{}
 	}
 
-	imagePath := filepath.Join(common.EnvConfig.UploadPath, "application-images", imageName+"."+ext)
-	if err := os.Remove(imagePath); err != nil && !os.IsNotExist(err) {
+	imagePath := path.Join("application-images", imageName+"."+ext)
+	if err := s.storage.Delete(context.Background(), imagePath); err != nil {
```
ItalyPaleAle: Here too, context?

backend/internal/service/oidc_service.go
```diff
@@ -884,34 +887,41 @@ func (s *OidcService) CreateClientSecret(ctx context.Context, clientID string) (
...
 	default:
-		return "", "", errors.New("image not found")
+		return nil, 0, "", errors.New("image not found")
+	}
+
+	mimeType := utils.GetImageMimeType(ext)
+	if mimeType == "" {
+		return nil, 0, "", fmt.Errorf("unsupported image type '%s'", ext)
+	}
+	key := path.Join("oidc-client-images", fmt.Sprintf("%s%s.%s", client.ID, suffix, ext))
```
ItalyPaleAle:
```suggestion
	key := path.Join("oidc-client-images", client.ID + suffix + "." + ext)
```

backend/internal/service/user_service.go
```diff
@@ -173,15 +171,8 @@ func (s *UserService) UpdateProfilePicture(userID string, file io.Reader) error
...
-	profilePictureDir := common.EnvConfig.UploadPath + "/profile-pictures"
-	err = os.MkdirAll(profilePictureDir, os.ModePerm)
-	if err != nil {
-		return err
-	}
-
-	// Create the profile picture file
-	err = utils.SaveFileStream(profilePicture, profilePictureDir+"/"+userID+".png")
+	profilePicturePath := path.Join("profile-pictures", userID+".png")
+	err = s.fileStorage.Save(context.Background(), profilePicturePath, profilePicture)
```
ItalyPaleAle: Context?

backend/internal/service/user_service.go
```diff
@@ -682,20 +671,10 @@ func (s *UserService) ResetProfilePicture(userID string) error {
...
-	// Check if file exists and delete it
-	if _, err := os.Stat(profilePicturePath); err == nil {
-		if err := os.Remove(profilePicturePath); err != nil {
-			return fmt.Errorf("failed to delete profile picture: %w", err)
-		}
-	} else if !os.IsNotExist(err) {
-		// If any error other than "file not exists"
-		return fmt.Errorf("failed to check if profile picture exists: %w", err)
+	profilePicturePath := path.Join("profile-pictures", userID+".png")
+	if err := s.fileStorage.Delete(context.Background(), profilePicturePath); err != nil {
```
ItalyPaleAle: Context?

backend/internal/common/env_config.go
```diff
@@ -72,30 +80,16 @@ func init() {
...
-		TracingEnabled:     false,
-		TrustProxy:         false,
-		AnalyticsDisabled:  false,
-		AllowDowngrade:     false,
-		InternalAppURL:     "",
+		AppEnv:        "production",
+		LogLevel:      "info",
+		DbProvider:    "sqlite",
+		FileBackend:   "fs",
+		KeysPath:      "data/keys",
```
ItalyPaleAle, line 87: Should keys be stored on S3 too? (**IF they are encrypted**)

ItalyPaleAle:
Pretty good job!

> final step needed to run Pocket ID in a serverless environment. 

Don't fully agree with the **final step** part, there's a few more things that won't work (like cron jobs)

backend/internal/storage/s3.go
stonith404: [`ToSlash`](https://pkg.go.dev/path/filepath#ToSlash) only exists in `filepath` and replaces all path separators with a `/`. I think this makes path normalization safe because if we would call `storage.Save(..., filepath.Join("some", "path"))` on Windows all `\` will be replaced with `/`.

backend/internal/storage/s3.go
stonith404: You mean because of [this warning](https://github.com/aws/smithy-go?tab=readme-ov-file#no_entry_sign-do-not-use-the-code-generators-in-this-repository)? We are not using the code generators, just `smithy.APIError` to handle errors, which is fine in my opinion.

backend/internal/utils/image/profile_picture.go
stonith404: S3 expects a seekable reader which isn't the case with `io.Pipe`. If we directly pass the stream of `io.Pipe` to `storage.Save` we get the following error: 

```
unseekable stream is not supported without TLS and trailing checksum
```

backend/internal/common/env_config.go
stonith404: Currently we require users to set `KEYS_STORAGE=database` if they use S3. I suggest that in the next major release, we should make it mandatory for users to store the keys in the database. I don't see any advantage to storing them on disk, or do you disagree?

backend/internal/job/file_cleanup_job.go
stonith404: I have added a new job that clears all temporary files that are older than 10 minutes in ec0cc49ce229e7d4b35987c124e0bbc4f41f454b.

tests/setup/docker-compose-s3.yml
stonith404: We are just using Minio for the e2e tests. The Docker setup for Garage seems pretty complicated just for a simple e2e test, I'm not sure if that's worth it.

ItalyPaleAle: Uhm true, their setup is not easy. But I also don't think pinning to un-maintained software is a good idea either, also because people shouldn't really use minio anymore (and given how things are going over there, I wouldn't be surprised if they removed the Docker images entirely)

backend/internal/storage/s3.go
ItalyPaleAle: Got it, I see

backend/internal/utils/image/profile_picture.go
ItalyPaleAle: Ugh, ok...

backend/internal/common/env_config.go
ItalyPaleAle: > I suggest that in the next major release, we should make it mandatory for users to store the keys in the database.

Yes!

> I don't see any advantage to storing them on disk

The only "downside" is that users will be required to set an encryption key for them, so an extra env var. Not a big downside IMHO.

ItalyPaleAle (approved):
LGTM with the additional comments above, non-blocking

kmendell (approved):
I did some basic test with it based on the e2e tests, all seems to work with no major issues :)

PR #1081, opened 2025-11-08
feat: light/dark/system mode switcher

kmendell:


kmendell:
Whoops forgot the translations keys....

ItalyPaleAle:
I am personally not a fan of client-side switching, because this will lead to "flashes of light mode" (a variant of "flashes of unstyled content").

I found apps are much better when dark mode follows the system only and uses the `dark:` variant for Tailwind...

PR #1086, opened 2025-11-10
chore: update example .env to use database for keys storage

ItalyPaleAle:
Given recent conversations on how the database will be the preferred (or sole) method in a future release, start by updating the example env


kmendell (approved):

stonith404:
I'm closing this in favor of #1088. I think it makes sense to make the next release breaking because there are a few issues in the backlog that require breaking changes as well.

PR #1088, opened 2025-11-11
feat!: drop support for storing JWK on the filesystem

stonith404:
This PR drops the support for storing the JWK on the filesystem, the JWK private key will now be stored in the database. There isn't really an advantage of storing them on the disk and keeping it supported adds unnecessary complicated logic to the import/export feature in #998.

This change removes the `KEYS_PATH` and `KEYS_STORAGE` environment variables. 

There are two breaking changes:
- The `ENCRYPTION_KEY` environment variable is now mandatory and must be at least 16 bytes long.
- The JWK private key stored on the disk will be lost and therefore all tokens will be invalidated. All users need to sign in again. Additionally, if another algorithm than `RS256` was used, the key needs to be regenerated again with `pocket-id key-rotate  -a <algo>`

Note: This PR gets merged into the `breaking/v2` branch where all breaking changes will get collected.

### DIFF
--- a/backend/cmd/main.go
+++ b/backend/cmd/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+	"os"
 	_ "time/tzdata"
 
 	"github.com/pocket-id/pocket-id/backend/internal/cmds"
@@ -13,7 +15,8 @@ import (
 
 func main() {
 	if err := common.ValidateEnvConfig(&common.EnvConfig); err != nil {
-		panic(err)
+		fmt.Fprintf(os.Stderr, "config error: %v\n", err)
+		os.Exit(1)
 	}
 	cmds.Execute()
 }
--- a/tests/setup/docker-compose-s3.yml
+++ b/tests/setup/docker-compose-s3.yml
@@ -36,7 +36,6 @@ services:
       S3_SECRET_ACCESS_KEY: test
       S3_FORCE_PATH_STYLE: true
       KEYS_STORAGE: database
-      ENCRYPTION_KEY: test
     depends_on:
       create-bucket:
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -1,5 +1,11 @@
 # See the documentation for more information: https://pocket-id.org/docs/configuration/environment-variables
+
+# These variables must be configured for your deployment:
 APP_URL=https://your-pocket-id-domain.com
+# Can be generated with `openssl rand -base64 32`
+ENCRYPTION_KEY=
```
ItalyPaleAle: Thoughts on adding an example of using ENCRYPTION_KEY_FILE like in https://github.com/pocket-id/pocket-id/pull/1086/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR13-R14 ? Files are a bit safer for things like encryption keys

kmendell: As long as our documentation is updated, i dont think we need to bloat the examples in the repo with every possible solution just the most common one.

ItalyPaleAle: My point is, the most common one **should be** the key file. People tend not to read docs, showing it here makes it much more visible

stonith404: I don't really get why it's more secure to use `ENCRYPTION_KEY_FILE` instead of `ENCRYPTION_KEY`, would you mind elaborating on that? If the advantage is really is minor I would still prefer to use `ENCRYPTION_KEY` as default because it makes the setup significantly easier.

ItalyPaleAle: You can read more on this issue when we implemented `_FILE` for all secrets: https://github.com/pocket-id/pocket-id/issues/685#issuecomment-3010523710

PR #1089, opened 2025-11-11
fix!: rename `LDAP_ATTRIBUTE_ADMIN_GROUP` env variable to `LDAP_ADMIN_GROUP_NAME`

### LINKED ISSUE #619: 🐛 Bug Report: LDAP_ATTRIBUTE_ADMIN_GROUP is misnamed
### Reproduction steps

This might be more of a deficiency than a bug, but I didn't have that option to select.

The configuration item LDAP_ATTRIBUTE_ADMIN_GROUP is not an attribute name, it's the name of a group. It should be called "LDAP_ADMIN_GROUP" or something like that.

### Expected behavior

Only attributes should begin with "LDAP_ATTRIBUTE_".

### Actual Behavior

A non-attribute begins with "LDAP_ATTRIBUTE_".

### Version and Environment

v1.2.0

### Log Output

_No response_

stonith404:
Fixes #619.

Breaking changes:
- `LDAP_ATTRIBUTE_ADMIN_GROUP` renamed to `LDAP_ADMIN_GROUP_NAME`.

### DIFF
--- a/backend/internal/dto/app_config_dto.go
+++ b/backend/internal/dto/app_config_dto.go
@@ -47,7 +47,7 @@ type AppConfigUpdateDto struct {
 	LdapAttributeGroupMember                   string `json:"ldapAttributeGroupMember"`
 	LdapAttributeGroupUniqueIdentifier         string `json:"ldapAttributeGroupUniqueIdentifier"`
 	LdapAttributeGroupName                     string `json:"ldapAttributeGroupName"`
-	LdapAttributeAdminGroup                    string `json:"ldapAttributeAdminGroup"`
+	LdapAdminGroupName                         string `json:"ldapAdminGroupName"`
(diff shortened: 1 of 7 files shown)


kmendell (approved):

AlexTMjugador:
Now that v2 has been released and in circulation for a while, maybe the documentation at https://pocket-id.org/docs/configuration/environment-variables should be updated as well? It still refers to the previous environment variable name.

kmendell:
> Now that v2 has been released and in circulation for a while, maybe the documentation at [pocket-id.org/docs/configuration/environment-variables](https://pocket-id.org/docs/configuration/environment-variables) should be updated as well? It still refers to the previous environment variable name.

Updated here :) https://github.com/pocket-id/website/commit/fd1ada38f7c474749401d76e9f9c3720f31a00a6

PR #1091, opened 2025-11-11
feat: add database storage backend

ItalyPaleAle:
Follow-up from #1080 that implemented S3 as storage

For small deployments, it makes sense to keep all data inside the database, which simplifies backing up of the data

### DIFF
--- a/backend/internal/service/oidc_service.go
+++ b/backend/internal/service/oidc_service.go
@@ -1974,13 +1974,9 @@ func httpClientWithCheckRedirect(source *http.Client, checkRedirect func(req *ht
 	}
 
 	// Create a new client that clones the transport
-	client := &http.Client{}
-	sourceTransport, ok := http.DefaultTransport.(*http.Transport)
-	if !ok {
-		// Indicates a development-time error
-		panic("Source transport is not of type *http.Transport")
+	client := &http.Client{
+		Transport: source.Transport,
 	}
-	client.Transport = sourceTransport.Clone()
 
 	// Assign the CheckRedirect function
 	client.CheckRedirect = checkRedirect
--- a/backend/internal/storage/database.go
+++ b/backend/internal/storage/database.go
@@ -125,7 +125,7 @@ func (s *databaseStorage) DeleteAll(ctx context.Context, prefix string) error {
 	}
 
 	query := s.db.WithContext(ctx)
-	query = addPathPrefixClause(s.db.Dialector.Name(), query, prefix)
+	query = addPathPrefixClause(s.db.Name(), query, prefix)
 	result := query.Delete(&model.Storage{})
 	if result.Error != nil {
 		return fmt.Errorf("failed to delete files with prefix '%s' from database: %w", prefix, result.Error)
@@ -145,7 +145,7 @@ func (s *databaseStorage) List(ctx context.Context, prefix string) ([]ObjectInfo
 		if !strings.HasSuffix(prefix, "/") {
 			prefix += "/"
 		}
-		query = addPathPrefixClause(s.db.Dialector.Name(), query, prefix)
+		query = addPathPrefixClause(s.db.Name(), query, prefix)
 	}
 
 	result := query.
@@ -184,7 +184,7 @@ func (s *databaseStorage) Walk(ctx context.Context, root string, fn func(ObjectI
 		if !strings.HasSuffix(root, "/") {
 			root += "/"
 		}
-		query = addPathPrefixClause(s.db.Dialector.Name(), query, root)
+		query = addPathPrefixClause(s.db.Name(), query, root)
 	}
 
 	result := query.
--- a/backend/internal/utils/ip_util_test.go
+++ b/backend/internal/utils/ip_util_test.go
@@ -1,8 +1,14 @@
 package utils
 
 import (
+	"context"
 	"net"
+	"net/url"
 	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/pocket-id/pocket-id/backend/internal/common"
 )
@@ -20,9 +26,8 @@ func TestIsLocalhostIP(t *testing.T) {
 
 	for _, tt := range tests {
 		ip := net.ParseIP(tt.ip)
-		if got := IsLocalhostIP(ip); got != tt.expected {
-			t.Errorf("IsLocalhostIP(%s) = %v, want %v", tt.ip, got, tt.expected)
-		}
+		got := IsLocalhostIP(ip)
+		assert.Equal(t, tt.expected, got)
 	}
 }
 
@@ -40,9 +45,8 @@ func TestIsPrivateLanIP(t *testing.T) {
 
 	for _, tt := range tests {
 		ip := net.ParseIP(tt.ip)
-		if got := IsPrivateLanIP(ip); got != tt.expected {
-			t.Errorf("IsPrivateLanIP(%s) = %v, want %v", tt.ip, got, tt.expected)
-		}
+		got := IsPrivateLanIP(ip)
+		assert.Equal(t, tt.expected, got)
 	}
 }
 
@@ -59,9 +63,9 @@ func TestIsTailscaleIP(t *testing.T) {
 
 	for _, tt := range tests {
 		ip := net.ParseIP(tt.ip)
-		if got := IsTailscaleIP(ip); got != tt.expected {
-			t.Errorf("IsTailscaleIP(%s) = %v, want %v", tt.ip, got, tt.expected)
-		}
+
+		got := IsTailscaleIP(ip)
+		assert.Equal(t, tt.expected, got)
 	}
 }
 
@@ -86,16 +90,17 @@ func TestIsLocalIPv6(t *testing.T) {
 
 	for _, tt := range tests {
 		ip := net.ParseIP(tt.ip)
-		if got := IsLocalIPv6(ip); got != tt.expected {
-			t.Errorf("IsLocalIPv6(%s) = %v, want %v", tt.ip, got, tt.expected)
-		}
+		got := IsLocalIPv6(ip)
+		assert.Equal(t, tt.expected, got)
 	}
 }
 
 func TestIsPrivateIP(t *testing.T) {
 	// Save and restore env config
 	origRanges := common.EnvConfig.LocalIPv6Ranges
-	defer func() { common.EnvConfig.LocalIPv6Ranges = origRanges }()
+	t.Cleanup(func() {
+		common.EnvConfig.LocalIPv6Ranges = origRanges
+	})
 
 	common.EnvConfig.LocalIPv6Ranges = "fd00::/8"
 	localIPv6Ranges = nil // reset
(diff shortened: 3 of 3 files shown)


backend/internal/model/storage.go
```diff
@@ -0,0 +1,17 @@
+package model
+
+import (
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+)
+
+type Storage struct {
+	Path      string            `gorm:"primaryKey;not null"`
+	Data      []byte            `gorm:"not null"`
```
stonith404: `gorm:"not null"` can be omitted since it only has an effect if we create the database schema with Gorm, but we are doing it manually with the migration.

backend/internal/service/ldap_service.go
```diff
@@ -101,9 +103,11 @@ func (s *LdapService) SyncAll(ctx context.Context) error {
 }
 
 //nolint:gocognit
-func (s *LdapService) SyncGroups(ctx context.Context, tx *gorm.DB, client *ldap.Conn) error {
+func (s *LdapService) SyncGroups(ctx context.Context, client *ldap.Conn) error {
 	dbConfig := s.appConfigService.GetDbConfig()
 
+	tx := utils.TransactionFromContext(ctx)
```
stonith404: I really like the solution to set the transaction on the context, I think we can use this everywhere in the future. The problem I'm seeing is that if you call `SyncGroups` without setting the transaction on the context it will panic.

My suggestion would be to pass `s.db` as a second parameter to `utils.TransactionFromContext` and if there is no transaction in the context, we just create a new one inside the `utils.TransactionFromContext` function.

backend/internal/service/ldap_service.go
```diff
@@ -68,25 +68,27 @@ func (s *LdapService) createClient() (*ldap.Conn, error) {
...
 	// Setup LDAP connection
 	client, err := s.createClient()
 	if err != nil {
 		return fmt.Errorf("failed to create LDAP client: %w", err)
 	}
 	defer client.Close()
 
-	err = s.SyncUsers(ctx, tx, client)
+	// Start a transaction
+	tx := s.db.Begin()
```
stonith404, line 89: What about creating an additional util that combines `s.db.Begin` with `utils.ContextWithTransaction`? I think this would reduce the risk of forgetting to call `utils.ContextWithTransaction`.

```go
func BeginTransactionWithContext(ctx context.Context, db *gorm.DB) (*gorm.DB, context.Context) {
	tx := db.Begin()
	return tx, ContextWithTransaction(ctx, tx)
}

tx, ctx := utils.BeginTransactionWithContext(ctx, s.db)
```

or maybe even pass the context as a pointer?

```go
func BeginTransactionWithContext(ctx *context.Context, db *gorm.DB) (*gorm.DB) {
	tx := db.Begin()
	*ctx = ContextWithTransaction(*ctx, tx)
	return tx
}

tx := utils.BeginTransactionWithContext(&ctx, s.db)
```

backend/internal/service/user_service.go
```diff
@@ -183,11 +183,14 @@ func (s *UserService) UpdateProfilePicture(ctx context.Context, userID string, f
 
 func (s *UserService) DeleteUser(ctx context.Context, userID string, allowLdapDelete bool) error {
 	return s.db.Transaction(func(tx *gorm.DB) error {
```
stonith404: If we would create a new transaction inside `utils.TransactionFromContext`, like described in my previous comment, we can get rid off the `deleteUserInternal` function and move its content directly into `DeleteUser`.

There might be some more "internal" functions that we can remove because of the same reason.

backend/internal/storage/database.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (s *databaseStorage) Type() string {
+	return TypeDatabase
+}
+
+func (s *databaseStorage) Save(ctx context.Context, relativePath string, data io.Reader) error {
+	// Get the database from the context, in case there's an active transaction
+	db := utils.TransactionFromContext(ctx)
+	if db == nil {
```
stonith404: This check would be unnecessary if we create the transaction inside `utils.TransactionFromContext`.

backend/internal/storage/database.go
```diff
@@ -0,0 +1,259 @@
...
+	prefix = filepath.ToSlash(filepath.Clean(prefix))
+
+	// Get the database from the context, in case there's an active transaction
+	db := utils.TransactionFromContext(ctx)
+	if db == nil {
+		db = s.db
+	}
+
+	// If empty prefix, delete all
+	if prefix == "" || prefix == "/" || prefix == "." {
```
stonith404: A util function might be cleaner for this because we are using this logic in three places inside the `storage` package.

tests/setup/docker-compose.yml
```diff
@@ -14,6 +14,9 @@ services:
       - '1411:1411'
     environment:
       - APP_ENV=test
+      - DB_PROVIDER=${DB_PROVIDER}
```
stonith404: I would prefer to keep the environment variables hard-coded in the the related Docker Compose files, because this makes it easier to set the environments up locally.

If they are hard-coded I can for example just run `docker compose -f docker-compose-postgres.yml up -d`.

.github/workflows/e2e-tests.yml
```diff
@@ -78,21 +88,18 @@ jobs:
...
-
       - name: Pull and save PostgreSQL image
         if: matrix.db == 'postgres' && steps.postgres-cache.outputs.cache-hit != 'true'
         run: |
           docker pull postgres:17
           docker save postgres:17 > /tmp/postgres-image.tar
-
       - name: Load PostgreSQL image
-        if: matrix.db == 'postgres' && steps.postgres-cache.outputs.cache-hit == 'true'
+        if: matrix.db == 'postgres' && steps.postgres-cache.outputs.cache-hit != 'true'
```
stonith404: Is this intended?

tests/setup/docker-compose.yml
ItalyPaleAle: I understand. The alternative would be creating 5 different docker-compose files, would you be ok with that?

.github/workflows/e2e-tests.yml
ItalyPaleAle: Whoops

ItalyPaleAle:
@stonith404 thanks for the review!

Regarding passing the transaction in the context... I am actually thinking of reverting it. I don't like it too much, and it seems that others agree it's [anti-pattern in Go](https://www.reddit.com/r/golang/comments/1awp5af/is_passing_database_transactions_as_via_context/). (I know it's common in other stacks, _I'm looking at you, Entity Framework_, but I agree that it's opaque while in Go we strive for clarity whenever posible)

Some ...context on why I did that. There are a couple of places where we are writing images to the storage while we have a database transaction active. The most notable example (not only) is while syncing with LDAP.

The above is a problem with SQLite, because while there's a transaction open, the entire database is locked. So we cannot perform any other query unless they re-use the same transaction. This made the tests fail, among others.

Thinking more about it, I am thinking of removing passing the transaction in the context. Not just because it's anti-pattern, but also and most importantly because while we hold a transaction, we are blocking the database (in the case of SQLite, entirely). Fetching images and writing them to disk can take a long time, and while that's happening the rest of Pocket ID isn't responding at all.

So I'm thinking instead of making it so we do not write to storage while in a transaction. In the case of LDAP, images will be added to storage in a step _after_ the users/groups have been synced.

What do you think?

stonith404:
@ItalyPaleAle Okay yeah it seems like no one recommends this solution 😅 I don't really understand why this is a anti pattern but the concern about blocking the database for too long is valid.

> So I'm thinking instead of making it so we do not write to storage while in a transaction. In the case of LDAP, images will be added to storage in a step after the users/groups have been synced.

Yes, I think this would be a good solution.

tests/setup/docker-compose.yml
stonith404: We could keep the following environment variables in `tests/setup/docker-compose-postgres.yml`:
```
      - DB_PROVIDER=postgres
      - DB_CONNECTION_STRING=postgres://postgres:postgres@postgres:5432/pocket-id
```

and just set `FILE_BACKEND=${FILE_BACKEND}` on the `docker-compose.yml`. This way we can keep only 3 Docker Compose files while still being able to use them independently without needing to set environment variables locally.

ItalyPaleAle: Done!

ItalyPaleAle:
> > So I'm thinking instead of making it so we do not write to storage while in a transaction. In the case of LDAP, images will be added to storage in a step after the users/groups have been synced.
> 
> Yes, I think this would be a good solution.

Ok should be done now. PTAL!

backend/internal/model/storage.go
```diff
@@ -0,0 +1,17 @@
+package model
+
+import (
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+)
+
+type Storage struct {
+	Path      string `gorm:"primaryKey;not null"`
```
stonith404:
```suggestion
	Path      string `gorm:"primaryKey"`
```

backend/internal/model/storage.go
```diff
@@ -0,0 +1,17 @@
...
+	Data      []byte
+	Size      int64
+	ModTime   datatype.DateTime
+	CreatedAt datatype.DateTime
+}
+
+func (Storage) TableName() string {
+	return "storage"
+}
```
stonith404, line 17:
```suggestion
```

The table name is already implicitly "storage."

stonith404 (approved):
I have two small suggestions, but everything else looks good to me, thanks!

stonith404:
Sorry my bad, my statement about the automatic table name was wrong. Gorm uses the plural form of the struct as table name. In this case defining it manually makes sense. I have reverted the commit.

ItalyPaleAle:
@stonith404 I found a couple more bugs and fixed them as I was working on this PR.

1. When creating a client and uploading both a light and dark logo, there was a race condition so sometimes one logo was not saved. This is because we were doing this by first loading the row, and then updating it. While we were using a transaction, that still caused a race condition on Postgres (with SQLite a transaction locks the entire database, but with Postgres we need to perform a "SELECT FOR UPDATE" or the row won't be locked and two queries can read it at the same time). _(PS: If you wonder why I'm obsessed with doing queries atomically whenever possible, this is one of the reasons - besides perf :) )_ Change is here: https://github.com/pocket-id/pocket-id/pull/1091/commits/e74474df70e70dc86e37580d57635ddc7c05a5a7#diff-25f504aed465ee11d0481b80d78561cabe0dff6d344ab071732760841bac5db8R2059-R2065
   - I'm flagging this because I'm sure there are other instances where this happens in the codebase
3. A second bug was that when clients are deleted, images were not deleted from the storage

ItalyPaleAle:
@coderabbitai review

backend/internal/storage/database.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	// Ensure prefix ends with / for proper prefix matching
+	if !strings.HasSuffix(prefix, "/") {
+		prefix += "/"
+	}
+
+	query := s.db.WithContext(ctx)
+	query = addPathPrefixClause(s.db.Name(), query, prefix)
+	result := query.Delete(&model.Storage{})
```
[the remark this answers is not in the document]
ItalyPaleAle: This is incorrect and causes linter failures

backend/internal/service/oidc_service.go
```diff
@@ -2005,7 +2024,7 @@
 	req.Header.Set("User-Agent", "pocket-id/oidc-logo-fetcher")
 	req.Header.Set("Accept", "image/*")
 
-	resp, err := s.httpClient.Do(req)
+	resp, err := client.Do(req)
```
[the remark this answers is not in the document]
ItalyPaleAle: That is precisely the point...

PR #1094, opened 2025-11-12
feat: add MySQL support to the application

jeboehm:
- Updated CI/CD workflow to include MySQL in the database matrix for end-to-end tests.
- Added MySQL image caching and loading steps in the workflow.
- Integrated MySQL database support including migration handling and environment configuration.
- Created initial MySQL migration files for database schema.
- Updated environment configuration tests to validate MySQL settings.

### DIFF
--- a/backend/internal/service/webauthn_service.go
+++ b/backend/internal/service/webauthn_service.go
@@ -120,15 +120,23 @@ func (s *WebAuthnService) VerifyRegistration(ctx context.Context, sessionID, use
 		tx.Rollback()
 	}()
 
-	// Load & delete the session row
+	// Load the session row first (MySQL doesn't support RETURNING in DELETE)
 	var storedSession model.WebauthnSession
 	err := tx.
 		WithContext(ctx).
-		Clauses(clause.Returning{}).
-		Delete(&storedSession, "id = ?", sessionID).
+		First(&storedSession, "id = ?", sessionID).
 		Error
 	if err != nil {
 		return model.WebauthnCredential{}, fmt.Errorf("failed to load WebAuthn session: %w", err)
+	}
+
+	// Delete the session after loading
+	err = tx.
+		WithContext(ctx).
+		Delete(&model.WebauthnSession{}, "id = ?", sessionID).
+		Error
+	if err != nil {
+		return model.WebauthnCredential{}, fmt.Errorf("failed to delete WebAuthn session: %w", err)
 	}
 
 	session := webauthn.SessionData{
@@ -224,15 +232,23 @@ func (s *WebAuthnService) VerifyLogin(ctx context.Context, sessionID string, cre
 		tx.Rollback()
 	}()
 
-	// Load & delete the session row
+	// Load the session row first (MySQL doesn't support RETURNING in DELETE)
 	var storedSession model.WebauthnSession
 	err := tx.
 		WithContext(ctx).
-		Clauses(clause.Returning{}).
-		Delete(&storedSession, "id = ?", sessionID).
+		First(&storedSession, "id = ?", sessionID).
 		Error
 	if err != nil {
 		return model.User{}, "", fmt.Errorf("failed to load WebAuthn session: %w", err)
+	}
+
+	// Delete the session after loading
+	err = tx.
+		WithContext(ctx).
+		Delete(&model.WebauthnSession{}, "id = ?", sessionID).
+		Error
+	if err != nil {
+		return model.User{}, "", fmt.Errorf("failed to delete WebAuthn session: %w", err)
 	}
 
 	session := webauthn.SessionData{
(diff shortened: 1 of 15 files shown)


backend/internal/service/webauthn_service.go
```diff
@@ -120,17 +120,25 @@ func (s *WebAuthnService) VerifyRegistration(ctx context.Context, sessionID, use
...
 	}()
 
-	// Load & delete the session row
+	// Load the session row first (MySQL doesn't support RETURNING in DELETE)
 	var storedSession model.WebauthnSession
 	err := tx.
 		WithContext(ctx).
-		Clauses(clause.Returning{}).
-		Delete(&storedSession, "id = ?", sessionID).
+		First(&storedSession, "id = ?", sessionID).
```
jeboehm, line 127: Do we need to READ LOCK the row?

backend/internal/service/webauthn_service.go
```diff
@@ -224,17 +232,25 @@ func (s *WebAuthnService) VerifyLogin(ctx context.Context, sessionID string, cre
...
```
jeboehm, line 239: Do we need to explicitly READ LOCK the row?

backend/internal/service/webauthn_service.go
ItalyPaleAle: Not on SQLite (a transaction blocks the entire database).

Good q on whether this is needed on PG/My

backend/internal/service/webauthn_service.go
```diff
@@ -120,17 +120,25 @@ func (s *WebAuthnService) VerifyRegistration(ctx context.Context, sessionID, use
...
 		Error
 	if err != nil {
 		return model.WebauthnCredential{}, fmt.Errorf("failed to load WebAuthn session: %w", err)
 	}
 
+	// Delete the session after loading
+	err = tx.
```
ItalyPaleAle, line 134: Can you do this in 2 queries just for MySQL? PG and SQLite support this as a single statement, so there's one less roundtrip

backend/internal/service/webauthn_service.go
ItalyPaleAle: Ditto as above

backend/internal/service/webauthn_service.go
```diff
@@ -224,17 +232,25 @@ func (s *WebAuthnService) VerifyLogin(ctx context.Context, sessionID string, cre
...
-		Clauses(clause.Returning{}).
-		Delete(&storedSession, "id = ?", sessionID).
+		First(&storedSession, "id = ?", sessionID).
 		Error
 	if err != nil {
 		return model.User{}, "", fmt.Errorf("failed to load WebAuthn session: %w", err)
 	}
 
+	// Delete the session after loading
+	err = tx.
```
ItalyPaleAle, line 246: Ditto as above

backend/internal/service/webauthn_service.go
```diff
@@ -390,17 +406,26 @@ func (s *WebAuthnService) CreateReauthenticationTokenWithWebauthn(ctx context.Co
...
-		Delete(&storedSession, "id = ? AND expires_at > ?", sessionID, datatype.DateTime(time.Now())).
+		Where("id = ? AND expires_at > ?", sessionID, datatype.DateTime(time.Now())).
+		First(&storedSession).
 		Error
 	if err != nil {
 		return "", fmt.Errorf("failed to load WebAuthn session: %w", err)
 	}
 
+	// Delete the session after loading
+	err = tx.
```
ItalyPaleAle, line 421: Ditto as above

ItalyPaleAle (commented):
Some review comments.

However, more generically, can you elaborate on the need for MySQL? For larger changes like these, usually the standard approach is to open an issue first and gauge the interest in the community...

ItalyPaleAle:
Also, one more method that needs to be updated is here: https://vscode.dev/github/pocket-id/pocket-id/blob/main/backend/internal/service/app_config_service.go#L110 Not sure how that would look like with MySQL, but we need to lock the table entirely

backend/internal/service/webauthn_service.go
jeboehm: The goal is to ensure the session is only used once, right?

jeboehm:
Thank you @ItalyPaleAle 
I‘ll come back to this in the next few days

stonith404:
Thanks for your contribution @jeboehm, I really appreciate it.

I agree with @ItalyPaleAle though that it would be helpful to open a feature request first so we can find out whether MySQL support is something that a broader group of users would benefit from. Adding support for another database engine isn’t just the initial implementation, it also adds ongoing maintenance and testing overhead due to database specific differences.

If only a small number of users would actually use MySQL, the long term cost may outweigh the benefit. I hope you can understand our hesitation here, and I’d be glad to continue the discussion in a feature request.

jeboehm:
I understand your position in general. However, I’d like to mention that the PR is pretty much finished.

Every RDBMS has its strengths and weaknesses. I personally tend to choose MySQL quite often, which is why I thought Pocket-ID could broaden its support a bit.

I’m happy to adjust the PR to fully meet the project’s requirements., just hmu!

ItalyPaleAle:
I do like seeing contributions like this too! I think it's very valuable work.

While I can't speak for @stonith404 the reason why I'm a bit hesitant myself is that the initial implementation is only the smaller part of the work, and the added maintenance effort that will be on the project (and project's maintainers) will be growing over time. We will need to test each change against one more database, and while the CI helps, some bugs will only be caught with manual tests (or in production...).

We've had multiple instances of bugs that only showed up with one database, e.g. just on Postgres or just on SQLite. As an example, in a [PR merged earlier today](https://github.com/pocket-id/pocket-id/pull/1091#issuecomment-3537224642) I fixed a race condition that only appeared on Postgres, and not on SQLite, due to how transactions work differently between the two. The bug was not introduced in the PR, but it was causing test failures in that PR specifically (not an usual thing for race conditions to show up when something unrelated changes).

I just agree that we will need to make sure that there's at least a few more people who would need support for MySQL, so we can be more confident that the maintenance effort will be worth it in the future.

> However, I’d like to mention that the PR is pretty much finished.

Actually have some bad news there, since #1091 added some more migrations :(

le-lenn:
Adding support for various DBs clearly violates pocket-ids vision in my opinion:

"Pocket ID is a **simple** OIDC provider that allows users to authenticate with their passkeys to your services."

Maintaining various DB backends raises the long-term maintenance **a lot**. I also wager that the benefits of different RDBMS are negligible for almost all installations.

I raised a discussion about pocket-ids long term vision/mission/roadmap. This PR shows me that more information on what pocket-id is supposed to become helps the community to contribute in a more focused and efficient manner.

Linking the discussion here:  
https://github.com/pocket-id/pocket-id/discussions/1170

stonith404:
I'm closing this because there wasn't really a demand by the community to add support for MySQL. Thanks anyways for your contribution, I appreciate it.

PR #1099, opened 2025-11-16
feat: adding/removing passkeys creates an entry in audit logs

### LINKED ISSUE #1095: 🚀 Feature: Create passkey fires an audit log event
### Feature description

When a user adds a new passkey to their account an audit event should be created.

### Pitch

Admins can quickly verify whether a passkey was legitimately added
Users can review their own security activity

ItalyPaleAle:
Fixes #1095

### DIFF
--- a/backend/internal/service/webauthn_service.go
+++ b/backend/internal/service/webauthn_service.go
@@ -300,7 +300,7 @@ func (s *WebAuthnService) DeleteCredential(ctx context.Context, userID string, c
 	}()
 
 	credential := &model.WebauthnCredential{}
-	err := s.db.
+	err := tx.
 		WithContext(ctx).
 		Clauses(clause.Returning{}).
 		Delete(credential, "id = ? AND user_id = ?", credentialID, userID).


backend/internal/service/webauthn_service.go
```diff
@@ -288,16 +293,30 @@ func (s *WebAuthnService) ListCredentials(ctx context.Context, userID string) ([
...
 
-func (s *WebAuthnService) DeleteCredential(ctx context.Context, userID, credentialID string) error {
+func (s *WebAuthnService) DeleteCredential(ctx context.Context, userID string, credentialID string, ipAddress string, userAgent string) error {
+	tx := s.db.Begin()
+	defer func() {
+		tx.Rollback()
+	}()
+
+	credential := &model.WebauthnCredential{}
 	err := s.db.
```
stonith404:
```suggestion
	err := tx.
```
You probably forgot to use the transaction here, in Sqlite this fails because the database is locked. Does it really make sense to use a transaction here though? Do we want to fail the deletion of the passkey when inserting in the audit log fails?

stonith404 (changes requested):

backend/internal/service/webauthn_service.go
ItalyPaleAle: Oh good catch, thank you!

I want to say yes? Because if people expect the log for security reasons, it missing would be a problem.

stonith404: Okay yeah fair.

stonith404 (approved):

PR #1108, opened 2025-11-20
Updates to import/export

ItalyPaleAle:
PR against #998

1. Reads the column types from the database to correctly scan and insert data into the database. Does not use heuristics such as checking if data "looks like a UTF-8 string".
2. Handles NULL-able columns with "double pointers"
3. Does not require disabling Foreign Key constraints. This is done by hard coding a list of 3 tables that need to be migrated first, in order, as they are the target of FK relationships. The problem with disabling FKs is that on Postgres it requires super-user permissions.
3. Fixes an error indicating that locks cannot be released after an import, since the kv table was dropped and all locks were destroyed
4. Fixes importing into an empty database when there's no KV table

Tested that this also allows switching between Postgres and SQLite

Note: I have not updated the E2E tests

### DIFF
--- a/backend/resources/files_test.go
+++ b/backend/resources/files_test.go
@@ -1,0 +1,72 @@
+package resources
+
+import (
+	"embed"
+	"slices"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This test is meant to enforce that for every new migration added, a file with the same migration number exists for all supported databases
+// This is necessary to ensure import/export works correctly
+// Note: if a migration is not needed for a database, ensure there's a file with an empty (no-op) migration (e.g. even just a comment)
+func TestMigrationsMatchingVersions(t *testing.T) {
+	// We can ignore migrations with version below 20251115000000
+	const ignoreBefore = 20251115000000
+
+	// Scan postgres migrations
+	postgresMigrations := scanMigrations(t, FS, "migrations/postgres", ignoreBefore)
+
+	// Scan sqlite migrations
+	sqliteMigrations := scanMigrations(t, FS, "migrations/sqlite", ignoreBefore)
+
+	// Sort both lists for consistent comparison
+	slices.Sort(postgresMigrations)
+	slices.Sort(sqliteMigrations)
+
+	// Compare the lists
+	assert.EqualValues(t, postgresMigrations, sqliteMigrations, "Migration versions must match between Postgres and SQLite")
+}
+
+// scanMigrations scans a directory for migration files and returns a list of versions
+func scanMigrations(t *testing.T, fs embed.FS, dir string, ignoreBefore int64) []int64 {
+	t.Helper()
+
+	entries, err := fs.ReadDir(dir)
+	require.NoErrorf(t, err, "Failed to read directory '%s'", dir)
+
+	// Divide by 2 because of up and down files
+	versions := make([]int64, 0, len(entries)/2)
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+
+		filename := entry.Name()
+
+		// Only consider .up.sql files
+		if !strings.HasSuffix(filename, ".up.sql") {
+			continue
+		}
+
+		// Extract version from filename (format: <version>_<anything>.up.sql)
+		versionString, _, ok := strings.Cut(filename, "_")
+		require.Truef(t, ok, "Migration file has unexpected format: %s", filename)
+
+		version, err := strconv.ParseInt(versionString, 10, 64)
+		require.NoErrorf(t, err, "Failed to parse version from filename '%s'", filename)
+
+		// Exclude migrations with version below ignoreBefore
+		if version < ignoreBefore {
+			continue
+		}
+
+		versions = append(versions, version)
+	}
+
+	return versions
+}
--- a/backend/resources/migrations/postgres/20251117141000_export_normalization.down.sql
+++ b/backend/resources/migrations/postgres/20251117141000_export_normalization.down.sql
@@ -1,0 +1,1 @@
+-- No-op in Postgres
--- a/backend/resources/migrations/postgres/20251117141000_export_normalization.up.sql
+++ b/backend/resources/migrations/postgres/20251117141000_export_normalization.up.sql
@@ -1,0 +1,1 @@
+-- No-op in Postgres


backend/internal/service/export_service.go
```diff
@@ -135,6 +117,56 @@ func (s *ExportService) dumpTable(table string, out *DatabaseExport) error {
...
+				res[i] = utils.Ptr(x)
+			}
+		case "blob", "bytea", "jsonb":
+			// Treat jsonb columns as binary too
+			var x []byte
+			if types[col].Nullable {
+				res[i] = utils.Ptr(utils.Ptr(x))
+			} else {
+				res[i] = utils.Ptr(x)
+			}
```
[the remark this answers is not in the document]
ItalyPaleAle: This is done automatically

backend/internal/service/import_service.go
```diff
@@ -154,109 +166,48 @@ func (s *ImportService) resetSchema(targetVersion uint, exportDbProvider string)
...
-			}
-			defer func() {
-				_ = setPostgresForeignKeyChecks(tx, true)
-			}()
+		// Iterate through all tables
+		// Some tables need to be processed in order
+		tables := make([]string, 0, len(dbData.Tables))
+		for _, t := range dbData.TableOrder {
+			// These tables need to be done first, in order
+			tables = append(tables, t)
```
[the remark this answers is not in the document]
ItalyPaleAle: This is not a problem here

backend/internal/service/import_service.go
```diff
@@ -154,109 +166,48 @@ func (s *ImportService) resetSchema(targetVersion uint, exportDbProvider string)
...
-		if strings.HasPrefix(err.Error(), "no migration found") && exportDbProvider != string(common.EnvConfig.DbProvider) {
-			if err := m.Up(); err != nil && !errors.Is(err, migrate.ErrNoChange) {
-				return fmt.Errorf("migration failed: %w", err)
-			}
-			return nil
-		}
+	err = m.Migrate(targetVersion)
+	if err != nil {
 		return fmt.Errorf("migration failed: %w", err)
 	}
```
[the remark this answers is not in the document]
ItalyPaleAle: How would that happen if we drop the entire database?

backend/internal/service/export_service.go
```diff
@@ -96,37 +81,34 @@ func (s *ExportService) schemaVersion() (uint, error) {
...
+		if err != nil {
 			return fmt.Errorf("failed to scan row in table %s: %w", table, err)
 		}
 
 		rowMap := make(map[string]any, len(cols))
 		for i, col := range cols {
-			sqlType := colTypes[i].DatabaseTypeName()
-			rowMap[col] = normalizeForJSON(vals[i], sqlType)
+			rowMap[col] = vals[i]
 		}
```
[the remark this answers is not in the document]
ItalyPaleAle: This is not a problem here

backend/internal/service/export_service.go
```diff
@@ -135,6 +117,56 @@ func (s *ExportService) dumpTable(table string, out *DatabaseExport) error {
...
+			} else {
+				res[i] = utils.Ptr(x)
+			}
+		case "timestamp", "timestamptz", "timestamp with time zone", "datetime":
+			var x datatype.DateTime
+			if types[col].Nullable {
+				res[i] = utils.Ptr(utils.Ptr(x))
+			} else {
+				res[i] = utils.Ptr(x)
+			}
```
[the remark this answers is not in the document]
ItalyPaleAle: DateTime implements MarshalJSON

backend/internal/utils/db_util.go
```diff
@@ -0,0 +1,116 @@
...
+			return nil, err
+		}
+		for _, table := range tables {
+			var cols []struct {
+				Name    string
+				Type    string
+				Notnull bool
+			}
+			err := db.
+				Raw(`PRAGMA table_info("` + table + `");`).
```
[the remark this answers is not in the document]
ItalyPaleAle: This is not a problem here

backend/internal/cmds/root.go
```diff
@@ -17,8 +17,11 @@ var rootCmd = &cobra.Command{
 	Long:         "By default, this command starts the pocket-id server.",
 	SilenceUsage: true,
 	Run: func(cmd *cobra.Command, args []string) {
+		// Get a context that is canceled when the application is stopping
+		ctx := signals.SignalContext(cmd.Context())
```
stonith404: Couldn't we create this context in the `Execute` function instead? This way all CLI commands will get cancelled on keyboard interrupt.
https://github.com/pocket-id/pocket-id/blob/6b5f08ca4a7f3c821a04486c05033cf90cf355ad/backend/internal/cmds/root.go#L33

backend/internal/service/import_service.go
```diff
@@ -58,11 +61,13 @@ func (s *ImportService) ImportFromZip(ctx context.Context, r *zip.Reader) error
...
 // ImportDatabase only imports the database data from the given DatabaseExport struct.
 func (s *ImportService) ImportDatabase(dbData DatabaseExport) error {
-	if err := s.resetSchema(dbData.Version, dbData.Provider); err != nil {
+	err := s.resetSchema(dbData.Version, dbData.Provider)
+	if err != nil {
 		return err
 	}
 
-	if err := s.insertData(dbData); err != nil {
+	err = s.insertData(dbData)
```
stonith404, line 69: Do you hate if statements with an initializer 😅? I usually prefer it if the initializer is short.

backend/internal/service/import_service.go
```diff
@@ -154,109 +166,48 @@ func (s *ImportService) resetSchema(targetVersion uint, exportDbProvider string)
...
-		// Special case: If no migrations are found, it may be due to a different DB provider.
-		// In that case we just apply all migrations from the current provider.
-		if strings.HasPrefix(err.Error(), "no migration found") && exportDbProvider != string(common.EnvConfig.DbProvider) {
-			if err := m.Up(); err != nil && !errors.Is(err, migrate.ErrNoChange) {
-				return fmt.Errorf("migration failed: %w", err)
-			}
-			return nil
-		}
+	err = m.Migrate(targetVersion)
+	if err != nil {
```
stonith404, line 170: I know the "special case" logic is not really great, but we still have to keep it. There might be releases where we only create a migration for a specific database provider. 

If we for example create a migration with the timestamp `2` for Postgres and Sqlite doesn't have this migration, the last known migration of Sqlite has the timestamp `1`. If the user exports from Postgres and imports it into Sqlite, the migration will fail because the migration with timestamp `2` can't be found.

stonith404 (changes requested):
Thanks for the changes! Please see my comments.

backend/internal/cmds/root.go
ItalyPaleAle: Yes, I think you're right. I moved it for a reason but it was an incorrect one

backend/internal/service/import_service.go
ItalyPaleAle: There were many instances where the assignment had a very long call (with like chained ORM calls, or multi-line things), and I thought that was very hard to read.

I feel that inlined assignments when you're doing business logic may be [less common in Go](https://www.reddit.com/r/golang/comments/ywvx2f/idiomatic_use_of_assignment_in_conditional/). There is also one small difference where:

```go
if err := s.insertData(dbData); err != nil {
```

Allocates a new `err` that is only scoped within the `if` block

While:

```go
err = s.insertData(dbData)
if err != nil {
```

Uses for `err` the same memory from the previous allocation.

AFAIK these are all allocated on the stack, not heap, so at least they don't cause more work on the GC... But they do use a tiny bit more memory.

backend/internal/service/import_service.go
ItalyPaleAle: Could we enforce that we always have migrations for all databases, even if they're just no-ops? I can add a unit test to enforce that if you prefer.

What could be a problem is:

- SQLite has 1, 2, 3, 4
- Postgres has 1, 3, 4

Create an export with schema version 2 on SQlite and import into Postgres. The latest version of PocketID has schema version 4. So then the import happen with schema 4...

stonith404: Okay yeah, that's probably the best solution. A migration for a single database provider is rare anyway.

stonith404 (approved):

PR #1113, opened 2025-11-26
V2 Breaking Changes

stonith404:
- Migrations need to be renamed before merging.
- Should be rebased instead of squash merged into main.


ItalyPaleAle:
@stonith404 I have re-ordered the migrations so the v2 ones come after those that have been released already

PR #1114, opened 2025-11-26
feat: remove DbProvider env variable and calculate it dynamically

kmendell:
Removes the `DbProvider` env variable, and calulates it dyanamically based on the connection string. 

DbProvider constants were left, as well as the value in the EnvConfigSchema as its used in alot of places, as keeping that seemed easier than making a helper but im not in love with it.. and open to changing it.

### DIFF
--- a/backend/internal/common/env_config.go
+++ b/backend/internal/common/env_config.go
@@ -131,12 +131,13 @@ func ValidateEnvConfig(config *EnvConfigSchema) error {
 		return errors.New("ENCRYPTION_KEY must be at least 16 bytes long")
 	}
 
-	if config.DbConnectionString == "" {
+	switch {
+	case config.DbConnectionString == "":
 		config.DbProvider = DbProviderSqlite
 		config.DbConnectionString = defaultSqliteConnString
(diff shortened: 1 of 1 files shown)


stonith404 (approved):
LGTM, thanks!

backend/internal/common/env_config.go
```diff
@@ -38,25 +38,25 @@ const (
...
-	TracingEnabled     bool       `env:"TRACING_ENABLED"`
-	TrustProxy         bool       `env:"TRUST_PROXY"`
-	AnalyticsDisabled  bool       `env:"ANALYTICS_DISABLED"`
-	AllowDowngrade     bool       `env:"ALLOW_DOWNGRADE"`
-	InternalAppURL     string     `env:"INTERNAL_APP_URL"`
+	AppEnv             AppEnv `env:"APP_ENV" options:"toLower"`
+	LogLevel           string `env:"LOG_LEVEL" options:"toLower"`
+	LogJSON            bool   `env:"LOG_JSON"`
+	AppURL             string `env:"APP_URL" options:"toLower,trimTrailingSlash"`
+	DbProvider         DbProvider
```
ItalyPaleAle, line 45: I think this is fine but I'd go one step further and completely remove this variable?

We can determine the DB provider in the code using `db.Name()`

Issue #742, opened 2025-07-11
🚀 Feature: User custom claims with LDAP properties


bluewalk:
### Feature description

* Usage of LDAP properties in custom claims
* Make these also configurable on LDAP level instead of user level (user claims template)

### Pitch

As you can create custom LDAP properties, like `quota` for Nextcloud or `mailbox_template` for Mailcow it would be awesome if you can use these properties when creating custom claims.

eg:
| Key | Value |
|--|--|
| quota | {{ldap.quota}} |

Furthermore it would make sense to have a user custom claims template in the LDAP configuration so you don't have to configure these custom claims for every user in LDAP.

tilwegener:
+1

PR #1117, opened 2025-11-27
fix/doc: Update launch.json and Contributing doc with minor fixes for first time setup of dev environment

mnestor:
When getting setup for the first time to try and implement #742 there were a couple of minor things I had to adjust to get up and running.

fix: Set correct path of .env file to backend dir instead of backend/cmd
fix: Set tag=exclude_frontend when working with backend in debug mode to match go run
fix: Set working directory to backend/ when working in backend mode in debug mode instead of backend/cmd to match go run
doc: Added note that may need to create a `data` directory in backend/ so that the sqlite database will be created

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,8 +65,6 @@ The backend is built with [Gin](https://gin-gonic.com) and written in Go. To set
 2. Copy the `.env.development-example` file to `.env` and edit the variables as needed
 3. Start the backend with `go run -tags exclude_frontend ./cmd`
 
-* Note: May need to create `data` folder in `backend` folder
-
 ### Frontend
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -65,6 +65,8 @@ The backend is built with [Gin](https://gin-gonic.com) and written in Go. To set
 2. Copy the `.env.development-example` file to `.env` and edit the variables as needed
 3. Start the backend with `go run -tags exclude_frontend ./cmd`
 
+* Note: May need to create `data` folder in `backend` folder
+
```
stonith404:
```suggestion
```

This was actually a bug. With cfc9e464d983b051e7ed4da1620fae61dc73cff2 the directory now should be created automatically.

stonith404 (approved):
Thanks!

PR #1121, opened 2025-11-28
feat: ldap custom attributes

mnestor:
- Feature for #742 
- Adds field ldapExtraAttributes to app configuration table
- Adds field is_ldap to custom_claims table
- Add configuration section to ldap for defining what attributes to sync to what keys in pocket-id along with denoting if the value is multivalue
- Adds extra section to User/Group pages showing the ldap synced attribute values
- Returns the ldap attributes with other custom claims when authenticating


mnestor:
There are a few things still missing I think. I implemented this as adding more custom-claims instead of putting them in a separate table to minimize the code changes. 

- Have claims set in pocket-id override ldap
- For group claims I'm a bit baffled on duplication, maybe nothing needs to be done there
- I did not look at removing of users/groups when disabling ldap at all

I'm still fairly new to svelte so I expect to have made big mistakes there and haven't gotten into the testing part yet at all.

Looking forward to feedback, I'm sure there will be plenty.

kmendell:
I'll try to take a look at this tonight for you.

kmendell:
First off my bad... Got very busy at the end of the year and completley spaced on this... I looked over it overall at a glance the logic seems okay to me, though there are synatx issues, but i have not dove deep enough, i think a starting point is to fix the E2e test and linters issues and merge the latest changes in. Once that is comeplete we can provide a better review of this. 

Once that is ready mark this PR as Read for review :)

stonith404:
Thanks for your contribution, I'm closing this in favor of #1486, which implements this as well.

bluewalk:
@stonith404 Can these custom fields also be derived from ldap attributes which was the original request in https://github.com/pocket-id/pocket-id/issues/742? I cannot seem to find this in https://github.com/pocket-id/pocket-id/pull/1486

stonith404:
@bluewalk I'm not sure if I understand you wrong, but if you e.g add a custom field called "nickname" to Pocket ID, it will automatically try to get this attribute from the LDAP server during the sync.

bluewalk:
> @bluewalk I'm not sure if I understand you wrong, but if you e.g add a custom field called "nickname" to Pocket ID, it will automatically try to get this attribute from the LDAP server during the sync

Ah perfect! It appears that I missed the following part in the PR, apologies for the confusion.
> If a custom field was added, Pocket ID automatically tries to get the custom field by its key from the LDAP server.

PR #1144, opened 2025-12-10
fix: restrict email one time sign in token to same browser

### LINKED ISSUE #1141: 🚀 Feature: Email Login Codes should be usable only on the device that requested them
### Feature description

For security reasons, email-based login codes should only be usable on the same device / browser that requested them.

Practically speaking, when users request an email login code for themselves, a cookie is set in the browser. The login code should be redeemable only if the cookie is available in the same browser.

### Pitch

Other apps enforce similar requirements, as this prevents certain kinds of social engineering attacks.

stonith404:
Fixes #1141


stonith404:
@ItalyPaleAle Good catch, thanks.

PR #1155, opened 2025-12-17
feat: add ability define user groups for sign up tokens

### LINKED ISSUE #1084: 🚀 Feature: Add users to Group(s) on signup via signup token
### Feature description

Let the admin select one or maybe even multiple groups when creating a Signup Token. When users signup via this Token they are automatically to these configured groups.

### Pitch

I have an app that I am planning to make accessible to my extended family but the app in question should only visible to a one User Group so that nobody besides my family can have access. This feature would spare me the headache of adding every family member to the group by hand after they sign up and users wouldn't have to wait for me to do so.

stonith404:
Closes #1084


kmendell:
Overall looks good, I haven't been able to fully test it yet, I will do that tomorrow and provide more comments if I find anything

PR #1158, opened 2025-12-19
feat: allow audit log retention to be controlled by env setting

jenic:
I am a paranoid person and want 1yr+ audit logs where possible. For your consideration, this change removes the hard-coded 90 day retention on audit logs within the database and allows it to be configured using `AUDIT_LOG_RETENTION_DAYS` environment variable, defaulting to 90 days.

I also made an attempt at changing the messages to not refer to the hard-coded 90 day value but I only know English so fair warning: the other languages are done with gAI (ChatGPT)


.env.example
```diff
@@ -15,4 +15,5 @@ ENCRYPTION_KEY=
 TRUST_PROXY=false
 MAXMIND_LICENSE_KEY=
 PUID=1000
-PGID=1000
\ No newline at end of file
+PGID=1000
+AUDIT_LOG_RETENTION_DAYS=90
```
ItalyPaleAle: We save the example env just for those things that are required - please revert the change to this file

frontend/messages/en.json
```diff
@@ -95,7 +95,7 @@
 	"settings": "Settings",
 	"update_pocket_id": "Update Pocket ID",
 	"powered_by": "Powered by",
-	"see_your_account_activities_from_the_last_3_months": "See your account activities from the last 3 months.",
+	"see_your_account_activities_from_the_last_3_months": "See your account activities within the configured retention period.",
```
ItalyPaleAle: What I'd recommend is to change the identifier from `see_your_account_activities_from_the_last_3_months` to
```suggestion
	"see_your_recent_account_activities": "See your recent account activities",
```

Because the identifier is changed now, you will not need to translate this to all languages. Contributors will eventually submit a translation for each language!

jenic: I can revert all the language translations and focus on the English one only, not a problem. I don't really agree with just saying "recent activities" though as those who are not familiar or comfortable looking at the code will not clearly see how long the audit logs are stored for. I would have preferred to include the configuration within the message, but Go is the only language I am really comfortable with here.

.env.example
jenic: No problem, I'll revert this.

frontend/messages/en.json
ItalyPaleAle: Feel free to adjust the language. My main advice was to make sure that the ID of the string needs to change too, so it will automatically invalidate all translations. You can populate the updated string in English only for now

ItalyPaleAle: PS: please don't force push commits, it makes reviewing the PR harder since GitHub's "Changes since your last review" won't work. We squash all PRs anyways, so no need to keep it all to a single commit

frontend/messages/en.json
```diff
@@ -328,7 +328,7 @@
 	"all_clients": "All Clients",
 	"all_locations": "All Locations",
 	"global_audit_log": "Global Audit Log",
-	"see_all_account_activities_from_the_last_3_months": "See all user activity for the last 3 months.",
+	"see_all_recent_account_activities": "See all recent user activity. (Default 90 days)",
```
stonith404:
```suggestion
	"see_all_recent_account_activities": "View the account activities of all users during the set retention period.",
```

frontend/messages/en.json
```diff
@@ -95,7 +95,7 @@
 	"settings": "Settings",
 	"update_pocket_id": "Update Pocket ID",
 	"powered_by": "Powered by",
-	"see_your_account_activities_from_the_last_3_months": "See your account activities from the last 3 months.",
+	"see_your_recent_account_activities": "See your recent account activities. (Default 90 days)",
```
stonith404:
```suggestion
	"see_your_recent_account_activities": "View your account activities during the set retention period.",
```
IMO this would fit better.

frontend/messages/en.json
jenic: Apologies for the force push, I wrongly assumed that would make it easier. I've pushed a new commit that should address this change.

PR #1161, opened 2025-12-22
fix!: make wildcard matching in callback URLs more stricter

stonith404:


frontend/messages/en.json
```diff
@@ -350,8 +349,8 @@
 	"login_code_email_success": "The login code has been sent to the user.",
 	"send_email": "Send Email",
 	"show_code": "Show Code",
-	"callback_url_description": "URL(s) provided by your client. Will be automatically added if left blank. Wildcards (*) are supported, but best avoided for better security.",
-	"logout_callback_url_description": "URL(s) provided by your client for logout. Wildcards (*) are supported, but best avoided for better security.",
+	"callback_url_description": "URL(s) provided by your client. Will be automatically added if left blank. <link href='https://pocket-id.org/docs/advanced/callback-url-wildcards'>Wildcards</link> are supported.",
```
ItalyPaleAle, line 352: Shouldn't we leave "but best avoided for better security"? This still holds true

stonith404: Sorry I forgot to provide the link to the new docs page: https://preview.pocket-id-website.pages.dev/docs/advanced/callback-url-wildcards

I have removed this statement because it's mentioned in the docs that exact URLs are preferred. Don't you think that would be enough?

ItalyPaleAle: Up to you! I thought it wouldn't hurt to keep it since _repetita iuvant_ (esp when about security). But I am fine with whatever you decide.

Issue #1037, opened 2025-10-18
🚀 Feature: Add 'Allowed OIDC Clients' to Group edit screen


mitchplze:
### Feature description

From within an app currently, you can set the **Allowed User Groups** from the app-edit page.

The FR is to add the opposite direction, so that **Allowed OIDC Clients** is a pane when editing a group.

### Pitch

Being able to edit the specific _apps_ that a group can access, from the group-edit screen, would be very convenient.

This functionality already exists for Users/Groups on both sides, as well.

stonith404:
Implemented in #1164.

PR #1164, opened 2025-12-23
feat: restrict oidc clients by user groups per default

### LINKED ISSUE #1052: 🚀 Feature: set default client group as none
### Feature description

feature request to change the new client logic to set no default access instead of the current unrestricted.

### Pitch

Having default unrestricted access is not great practice imo, as forgetting to set allowed groups allows all users to access that client which is in many cases unwanted.

The tradeoff is that a default group needs to be set or else no one can access the client.

An alternative could be to create a new group for each user and set that 'group' as the default user.

stonith404:
This PR changes the default behavior for OIDC client access restrictions.

Previously an OIDC client was implicitly allowed for all user groups if no user groups were selected. This implicit logic has been removed. By default newly created OIDC clients now allow no user groups.

To preserve the previous behavior in an explicit way, an “Unrestricted” option has been added. When enabled, the OIDC client becomes accessible to all user groups, matching how earlier versions behaved when no user group was selected.

Screenshot 2025-12-23 at 16 59 07

Additionally, it is now possible to configure allowed OIDC clients directly from the user group detail page. This behaves the same as configuring allowed user groups on the OIDC client side, both approaches modify the same underlying access rules.

Screenshot 2025-12-23 at 17 04 26

Closes #1052 and #1037

### DIFF
--- a/frontend/src/routes/settings/admin/oidc-clients/[id]/+page.svelte
+++ b/frontend/src/routes/settings/admin/oidc-clients/[id]/+page.svelte
@@ -83,8 +83,14 @@
 	async function enableGroupRestriction() {
 		client.isGroupRestricted = true;
 		await oidcService
-			.updateClient(client.id, client)
-			.then(() => toast.success(m.user_groups_restriction_updated_successfully()))
+			.updateClient(client.id, {
+				...client,
+				isGroupRestricted: true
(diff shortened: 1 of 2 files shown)


stonith404:
@coderabbitai review

PR #1209, opened 2026-01-06
feat: add CLI command for encryption key rotation

### LINKED ISSUE #1199: 🚀 Feature: support rotating encryption key
### Feature description

I would like the ability to rotate the encryption key for pocket-id. Something like a command that could be run via ./pocket-id to support old / new key to support this or some method for the import/export command that was recently released.

### Pitch

This could help for mistakes in a docker-compose setup (missing variable) or in the event the key is compromised it needs to be rotated. 

I verified with import/export but was unable to get it to work successfully with an existing passkey authentication after performing the import/export after the container was started with the fresh key.

stonith404:
Closes #1199


stonith404:
@coderabbitai review

PR #1214, opened 2026-01-07
feat: add option to renew API key

### LINKED ISSUE #1207: 🚀 Feature: API key renewal
### Feature description

Renew a expired API key without having to register another one.

### Pitch

My API key for Terraform just expired a few days ago, and I noticed that there's no 'renew' button for the expired key.

Image

stonith404:
Adds an option to renew API keys:
Screenshot 2026-01-07 at 16 08 28

Closes #1207

### DIFF
--- a/tests/resources/export/database.json
+++ b/tests/resources/export/database.json
@@ -5,9 +5,12 @@
 	"tables": {
 		"api_keys": [
 			{
+				"created_at": "2025-12-21T19:12:03Z",
 				"description": null,
 				"expiration_email_sent": false,
+				"expires_at": "2025-12-25T12:00:00Z",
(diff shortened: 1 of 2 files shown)


stonith404:
@coderabbitai review

frontend/src/routes/settings/admin/api-keys/renew-api-key-modal.svelte
kmendell, line 1: Thinking out loud would it make sense to somehow combind the dialogs so its only one component? We add a prop for a "mode" if mode is new then its shows the new form, if the mode is renew then it shows the renew form ?

stonith404: I mean it's a good idea but I think in this case it would make the component a bit too confusing, because the logic inside the script tag is different between these two forms.

PR #1215, opened 2026-01-07
feat: make home page URL configurable

### LINKED ISSUE #1157: 🚀 Feature: Make "My apps" the landing page instead of "My Account", and/or make the landing page configurable.
### Feature description

In my opinion, it would be better if “My Apps” were the landing page instead of “My Account,” and it would be even better, of course, if this could be configured for users who are of a different opinion.

### Pitch

Users usually need to launch apps much more often than they need to change their profile picture or add or remove passkeys. Furthermore, The "My Apps" section of Pocket ID could then be used as the start page in browsers.

stonith404:
Adds an option to select the home page URL:
Screenshot 2026-01-07 at 17 29 16

Closes #1157


stonith404:
@coderabbitai review

PR #1229, opened 2026-01-11
feat: add static api key env variable

### LINKED ISSUE #1166: 🚀 Feature: Improvements for declarative installations
### Feature description

* Allow specifying admin API key as env var.
* Allow specifying custom client secrets instead of requesting them.
* ... Something else?

### Pitch

I like the idea of Pocket ID being simple, but I struggle with setting it up in my NixOS installation. A common pattern in NixOS is to generate and encrypt secrets with something like age and store them in the same repo as the configuration. On service start, secrets are decrypted and fed into environment variables or as file paths.

For example, I use this pattern with Authelia and Immich. I can store a secret once and refer to it from both configurations, as they can pick up secrets from config files or environment variables. This also works well with tools like agenix-rekey, which makes it easy to regenerate all secrets with one command and then deploy.

Although Pocket ID's API can be used in scripts to manage entities declaratively, it's hard to use right now. My current method is ugly: on service start, a script requests a one-time access token via the pocket-id CLI, parses it from the human-readable text, and then exchanges it for a cookie to make API calls. This is fragile because the CLI output may change, and it requires creating an admin user first.

Several related issues exist (e.g., #83, #586), but they don't fully explain the use case for declarative setups.

stonith404:
Adds support for the `STATIC_API_KEY` environment variable. This variable can be useful if you need API access before the first user has been created for example with declarative installations.

Closes #1166

### DIFF
--- a/backend/internal/service/api_key_service.go
+++ b/backend/internal/service/api_key_service.go
@@ -26,9 +26,11 @@ type ApiKeyService struct {
 func NewApiKeyService(ctx context.Context, db *gorm.DB, emailService *EmailService) (*ApiKeyService, error) {
 	s := &ApiKeyService{db: db, emailService: emailService}
 
-	_, err := s.initStaticApiKeyUser(ctx, common.EnvConfig.StaticApiKey)
-	if err != nil {
-		return nil, err
+	if common.EnvConfig.StaticApiKey == "" {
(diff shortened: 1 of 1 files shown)


stonith404:
@coderabbitai review

PR #1254, opened 2026-01-18
feat: add VERSION_CHECK_DISABLED environment variable

dihmandrake:
Allows disabling the automatic version check in the backend. This is useful for air-gapped environments where external requests to the version endpoint fail or time out. The frontend has been updated to handle disabled checks gracefully.

Doc changes under: https://github.com/pocket-id/website/pull/208


kmendell (approved):
Thanks!

PR #1288, opened 2026-01-31
feat: allow first name and display name to be optional

taoso:
In some simple scenario, it seems too cumbersome to require the First/Last/Display names.

So I propose to make all these fields optional.

BTW:

~~Please give advice on when to use the `emptyToUndefined()`~~

Make this fields optional is a little hard for me. Instead, I just let them accept empty string.

Thanks.


MarkusThielker:
I don't know how to feel about making every field optional, except email and username, but the maintainers will have to decide if this is merged. As far as I can see this is not related to any open feature request?

But what I really don't get:
1. why did you change the required minimal length of the username inside the UserCreateDto?
2. why did you reorder the form fields?

taoso:
Sorry for unclarifying @MarkusThielker @kmendell 

> I don't know how to feel about making every field optional, except email and username, but the maintainers will have to decide if this is merged. As far as I can see this is not related to any open feature request?
> 
> But what I really don't get:

> 1. why did you change the required minimal length of the username inside the UserCreateDto?

The current minimum length of username is 2, which disallow username with only one char.

If there is no limit in the standard, maybe it's a good convenient to relax this restriction.

> 2. why did you reorder the form fields?

Just to make the required fields in the top area and the optional ones in the following area, which, in my personal opinion, will be more convenient for scenarios that first/display names ares optional.

kmendell:
I also not sure how I feel about this, partially the reason I haven't fully reviewed it, along with Elias being gone for a bit. However how many user names really are 1 character? This could cause conflicts in the database potentially and users could have the same username.

MarkusThielker:
> However how many user names really are 1 character? This could cause conflicts in the database potentially and users could have the same username.

@kmendell 
While I agree with you, I can not see how this could lead to database conflicts. Could you elaborate on that?

@taoso 
Even if the length of the username was changed, this is not yet applied to the frontend. The username Zod schema has not been adjusted.

kmendell:
Someone could use the api and not hit the Zod schema

taoso:
Hi @kmendell , I have fixed the e2e test. Could you approve the workflows?

taoso:
@kmendell @MarkusThielker 
> However how many user names really are 1 character? 

 As a infrastructure, pocket id should not limit the min length of user name to 2. It's seems better to  be decide by the system administrator.

> This could cause conflicts in the database potentially and users could have the same username.

Allowing one char username has no relation to the conflicts of database. Please offer example for this case. Thanks.

MarkusThielker:
> As a infrastructure, pocket id should not limit the min length of user name to 2. It's seems better to be decide by the system administrator.

Agreed. 
My question was intended to point out that this change has nothing to do with the PR title or description. Given that this change, which is reasonable, is not described in the PR, it should not be part of this PR.
And on top of this, the changes you made regarding this, were incomplete at the time since the usernameSchema had not been adjusted (done by now in [d7e7659](https://github.com/pocket-id/pocket-id/pull/1288/commits/d7e76598e94b818e92a92666eef0c8ca8ff816a0))

taoso:
> > As a infrastructure, pocket id should not limit the min length of user name to 2. It's seems better to be decide by the system administrator.
> 
> Agreed. My question was intended to point out that this change has nothing to do with the PR title or description. Given that this change, which is reasonable, is not described in the PR, it should not be part of this PR. And on top of this, the changes you made regarding this, were incomplete at the time since the usernameSchema had not been adjusted (done by now in [d7e7659](https://github.com/pocket-id/pocket-id/pull/1288/commits/d7e76598e94b818e92a92666eef0c8ca8ff816a0))

Sorry for un-clarification. The related changes have been reverted.

@MarkusThielker @kmendell

kmendell:
LGTM, Thanks

PR #1290, opened 2026-01-31
feat: allow clearing background image

taoso:


kmendell:
Just curious Why would one want to delete the background? It would be a black area on the logic screen then?

taoso:
> Just curious Why would one want to delete the background? It would be a black area on the logic screen then?

Just make pocket-id as lightweight as possible.

After the background deleted, the login form will be centered in the page.

Maybe its core to show some image next to the form. However, it's kind of burden to choose suitable picture for self-hosted instance.

So I propose to make it possible to clean the image.

clemone210:
thanks for this!
I really wanted this since a long time.
My previous "workaround" was to use a transparent image.
This is much appreciated.

Issue #1297, opened 2026-02-05
🚀 Feature: Create a new API endpoint to get current version


Na0mir:
### Feature description

I'd like to use [release argus](https://release-argus.io/) to check the deployed version of Pocket-ID via API.
Currently there is only the endpoint /api/version/latest. Is that possible to create an /api/version/current ?

### Pitch

This will allow the dashboard to show the latest version (from github) and the deployed version (from my installed server).

crivchri:
Hi,

Same for me, I would really like the installed version of pocket-id (self hosted) to be returned by the API. This would allow to detect whether a new version is available, and use that information with the Argus tool (https://release-argus.io/).

I would suggest creating a `/api/version` endpoint to be added that can only be accessed from a user with an admin role.

Thank you in advance and kind regards!

PR #1310, opened 2026-02-10
feat: current version api endpoint

kmendell:
Closes: https://github.com/pocket-id/pocket-id/issues/1297

### DIFF
--- a/backend/internal/controller/version_controller.go
+++ b/backend/internal/controller/version_controller.go
@@ -6,15 +6,16 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/pocket-id/pocket-id/backend/internal/common"
+	"github.com/pocket-id/pocket-id/backend/internal/middleware"
 	"github.com/pocket-id/pocket-id/backend/internal/service"
 	"github.com/pocket-id/pocket-id/backend/internal/utils"
 )
 
 // NewVersionController registers version-related routes.
-func NewVersionController(group *gin.RouterGroup, versionService *service.VersionService) {
+func NewVersionController(group *gin.RouterGroup, authMiddleware *middleware.AuthMiddleware, versionService *service.VersionService) {
 	vc := &VersionController{versionService: versionService}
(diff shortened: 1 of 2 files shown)


kmendell:
@coderabbitai review

backend/internal/controller/version_controller.go
```diff
@@ -38,3 +40,17 @@ func (vc *VersionController) getLatestVersionHandler(c *gin.Context) {
...
 }
+
+// getCurrentVersionHandler godoc
+// @Summary Get current deployed version of Pocket ID
+// @Tags Version
+// @Produce json
+// @Success 200 {object} map[string]string "Current version information"
+// @Router /api/version/current [get]
+func (vc *VersionController) getCurrentVersionHandler(c *gin.Context) {
+	utils.SetCacheControlHeader(c, 5*time.Minute, 15*time.Minute)
```
ItalyPaleAle: I would remove this, since it's a dynamic route. If the user re-deploys Pocket ID with an updated version, callers should see that immediately.

backend/internal/controller/version_controller.go
```diff
@@ -13,6 +14,7 @@ import (
 func NewVersionController(group *gin.RouterGroup, versionService *service.VersionService) {
 	vc := &VersionController{versionService: versionService}
 	group.GET("/version/latest", vc.getLatestVersionHandler)
+	group.GET("/version/current", vc.getCurrentVersionHandler)
```
ItalyPaleAle: Should this route require authentication? Since it can be used to expose the version of the app, which _could_ be used by attackers to check if it's a version vulnerable to some attacks.

ItalyPaleAle: See: https://security.stackexchange.com/questions/170352/is-it-safe-to-display-version-information-on-a-public-webpage-of-your-web-app

kmendell: Oops I meant too but forgot :o I'll add it

PR #1321, opened 2026-02-18
fix: token endpoint must not accept params as query string args

ItalyPaleAle:
Discovered this accidentally. Per RFC-6749, the /token endpoint must not accept parameters sent in the query string, which must instead all be set in the body (client credentials can be in the Authorization header). This is for security reasons, since query string args can be included in server logs or logged by proxies.

However, Gin's default behavior for c.ShouldBind is to _also_ include all paramters sent in the URL.

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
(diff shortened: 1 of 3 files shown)


kmendell (approved):
LGTM

PR #1322, opened 2026-02-18
feat: add JWT ID for generated tokens

### LINKED ISSUE #1320: 🚀 Feature: Include JWT ID for each generated token
### Feature description

Include a JWT ID at the `jti` claim in access tokens and ID tokens.  This is described in [RFC7519 4.1.7](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7).

### Pitch

While the `jti` claim is optional per the JSON Web Token RFC, it allows for some applications to block tokens that have been revoked, such as during a "sign out" process.

This may be generated using the same UUID generation code we use elsewhere.

In my example, this is expected by Ceph to support "signing out".

imnotjames:
adds the JWT ID as the `jti` claim on access tokens, oauth tokens, and identity tokens

fixes #1320

### DIFF
--- a/backend/internal/service/jwt_service.go
+++ b/backend/internal/service/jwt_service.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/lestrrat-go/jwx/v3/jwa"
 	"github.com/lestrrat-go/jwx/v3/jwk"
 	"github.com/lestrrat-go/jwx/v3/jwt"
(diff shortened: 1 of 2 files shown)


imnotjames:
I ran the tests in my repo and they seemed to pass.

Open to suggestions on any extra tests that should be written.  I had started to write a test to validate that the JWT ID changes but that seemed to be testing the uuid implementation.

I also had been looking at using v6 or v7 UUIDs but thought that's up to y'all.  For my use cases it's not as important.

imnotjames:
Thanks!

PR #1328, opened 2026-02-22
chore(deps): upgrade to node 24 and go 1.26.0

kmendell:


kmendell:
@stonith404 @ItalyPaleAle go 1.26 rejects malformed urls without brackets now as they wouldn't be a valid url otherwise, I replace the test with a proper compliant ipv6 url wildcard test, I don't think it should be an issue but if you think it may be or maybe im misunderstanding something too. 

Im not too sharp with ipv6 stuff so hence the confusion form my end. 


image

backend/internal/bootstrap/observability_boostrap.go
```diff
@@ -119,10 +119,10 @@ func initOtelLogging(ctx context.Context, resource *resource.Resource) error {
 	globallog.SetLoggerProvider(provider)
 
 	// Wrap the handler in a "fanout" one
```
stonith404:
```suggestion
```

backend/internal/dto/scim_dto.go
```diff
@@ -61,14 +61,14 @@ type ScimResourceData struct {
 	ID         string           `json:"id,omitempty"`
 	ExternalID string           `json:"externalId,omitempty"`
 	Schemas    []string         `json:"schemas"`
-	Meta       ScimResourceMeta `json:"meta,omitempty"`
```
stonith404, line 64: I think we should keep `omitempty`'s here, because there are cases where `ScimResourceMeta` isn't populated.

backend/internal/dto/scim_dto.go
```diff
@@ -61,14 +61,14 @@ type ScimResourceData struct {
...
+	Meta       ScimResourceMeta `json:"meta"`
 }
 
 type ScimResourceMeta struct {
 	Location     string    `json:"location,omitempty"`
 	ResourceType string    `json:"resourceType,omitempty"`
-	Created      time.Time `json:"created,omitempty"`
-	LastModified time.Time `json:"lastModified,omitempty"`
```
stonith404, line 71: `omitempty` should be kept here as well, because `LastModified` can be `NULL`.

backend/internal/dto/scim_dto.go
kmendell: I'll add back that was go fix and I didn't notice it

kmendell:
@stonith404 Fixed, let me know if there more did a glance over again and all the rest should be good.

PR #1329, opened 2026-02-22
ci/cd: migrate github actions runners to depot  runners

kmendell:


kmendell:
@stonith404 If you want we can start with just the runner configurations that way all runs on depot then we can work on building the conatiners it self with depot, i tried the e2e tests one, and it will have to be a big refactor for those since depot doesnt support loading image like we do now, we would just have to build directly when playwright runs i tried to do it, but got frustrated as everything except the sqlite/fs tests were "failing" lol

kmendell:
@ItalyPaleAle From a older conversation -- The attestaions should still be pushed to github for all of the images, i cant test this on the Pr, but they should according to what i read as depot uses githubs OIDC token ust provides the self hosted runner basically.

stonith404 (changes requested):
As discussed, I would only migrate the following actions to Depot:
- `.github/workflows/release.yml`
- `.github/workflows/build-next.yml`

There is unfortunately no way to make the logs of the Depot actions public and this is not ideal for actions that run as a check.

kmendell:
@stonith404 This specific PR just updates the Runners to be "self hosted" or in this case depots runners, the whole Logs conversation was for Depot CI itself, Do you wnat me to add depot ci for those too in addition to all of this?

stonith404 (approved):
@kmendell Oh I misunderstood something, thanks for clarifying. I though that the the Depot CI is the same as Depot GitHub runners, but that's not the case.

ItalyPaleAle:
Folks it's up to you eventually but I am not super comfortable with this.

This project is small enough that CI doesn't take too long (I have worked on projects where CI took 1+ hours). I am unsure if the added complexity is worth it...

kmendell:
Im not really sure what you mean by complexity, all of the stuff with depot is drop in, a lot of the other changes were just to get the builds working with the e2e tests correctly, the speed is just one of the things depot provides due to its better caching system than githubs and the infrastructure it runs on.  -- that being said depot ci is another conversation entirely

ItalyPaleAle:
I'm referring to the complexity of having a separate CI infrastructure in which GH Actions triggers depot, but then we finish in GHA for example for code signing.

What level of speed improvements are we seeing with depot?

PR #1332, opened 2026-02-23
fix: improve wildcard matching by using `go-urlpattern`

stonith404:
A vulnerability has been reported, where we had a bug in our callback URL wildcard matching. It's really hard to maintain this logic because there are so many edge cases. Because of that I decided to use the library [`go-urlpattern`](https://pkg.go.dev/github.com/dunglas/go-urlpattern) library. 

A few tweaks to wrapper have been made:
1. As this library uses the [WHATWG URL Pattern Living Standard](https://urlpattern.spec.whatwg.org/) wildcards are handled differently. Because of that our logic replaces any globstars (**) with a single asterix (\*) and all single asterix (\*) with a url slug (`:param`)
2. The query parameter validation is done manually, because the library doesn't validate query params as strictly as needed.

Thanks to @ByamB4 which has reported the vulnerability.


backend/internal/utils/callback_url_util.go
```diff
@@ -64,143 +87,135 @@ func matchCallbackURL(pattern string, inputCallbackURL string) (matches bool, er
...
-			i++
+		} else {
+			result.WriteByte(patternPath[i])
 		}
 	}
+	patternPath = result.String()
 
-	regexPattern.WriteString("$")
+	// Replace globstar wildcards (**) with a single *
+	re := regexp.MustCompile(`\*{2}`)
```
ItalyPaleAle: Optimization for a future PR:

1. Consider moving this regexp to a static variable in the package. This way, it's compiled once rather than on each request
2. Maybe even better, rather than using a regex, see if this can be done by iterating through each rune. the algorithm would be something like

```go
func collapseDoubleAsterisks(s string) string {
	// Fast path: if no consecutive asterisks, return original string
	if !strings.Contains(s, "**") {
		return s
	}

	b := strings.Builder{}
	b.Grow(len(s))

	for i := 0; i < len(s); i++ {
		// If we detect "**", write one "*" and skip the next character
		if s[i] == '*' && i+1 < len(s) && s[i+1] == '*' {
			b.WriteByte('*')
			// Skip second '*'
			i++
			continue
		}

		b.WriteByte(s[i])
	}

	return b.String()
}
```

backend/internal/utils/callback_url_util.go
```diff
@@ -64,143 +87,135 @@ func matchCallbackURL(pattern string, inputCallbackURL string) (matches bool, er
...
-		rest := s[i+3:]
+		rest := url[i+3:]
 		if j := strings.IndexRune(rest, '/'); j >= 0 {
 			pathStart = i + 3 + j
 		}
 	} else {
 		// Otherwise, first slash is path start
-		pathStart = strings.IndexRune(s, '/')
+		pathStart = strings.IndexRune(url, '/')
 	}
```
ItalyPaleAle: Nit: `bytes.IndexByte` is a lot faster than `strings.IndexRune` when you're not dealing with Unicode chars

backend/internal/utils/callback_url_util.go
```diff
@@ -64,143 +87,135 @@ func matchCallbackURL(pattern string, inputCallbackURL string) (matches bool, er
...
-		parts = strings.FieldsFunc(s, split)
 		path = ""
+		base = url
+	}
+
+	return base, path
+}
+
+func extractQueryParams(rawUrl string) (base string, query url.Values, err error) {
+	if i := strings.Index(rawUrl, "?"); i >= 0 {
```
ItalyPaleAle: bytes.IndexByte here too ideally

Issue #936, opened 2025-09-11
🚀 Feature: Security notifications to telegram discord webhook etc.


lordraiden:
### Feature description

A feature to generate notifications based on some scenarios

- user: add delete modify
- tokens: add delete
- admin: login
- permissions: new admin
- app/integration: add delete

### Pitch

For security reasons a way to be notified about relevant changes in the configuration.

kmendell:
Pocket ID is meant to be soley a OIDC provider, Features like this would bloat the UI with configuration options in order to configure them, and Pocket ID biggest goal is to keep everything simple and to the point. Features like this cabn easily be built around the api for your individual use case.

lordraiden:
> Pocket ID is meant to be soley a OIDC provider, Features like this would bloat the UI with configuration options in order to configure them, and Pocket ID biggest goal is to keep everything simple and to the point. Features like this cabn easily be built around the api for your individual use case.

Configuration options? it requires 2 parameters to send notifications to telegram. and the 4 check boxes to select what you want to send...

cdanis:
I came here to file a FR for similar functionality.

I understand we don't want to bloat the UI, have to take on maintenance burdens for many different notification services, etc.

But can we instead added support for just **one** simple outgoing webhook implementation, and then leave it to users or the community to hook that up to actual push notification services?

POSTing the JSON payload of an audit-log event would be simple and sufficient -- exactly what you can poll from the API, only pushed.

(And in the case of common self-hosted push servers like ntfy.sh, this would already be sufficient, along with some basic [configuration](https://docs.ntfy.sh/publish/#custom-templates) on the ntfy side.)

Alfredooe:
I'd quite like to see this re-opened if possible @kmendell , happy to work on a PR. I don't think this would add excessive complexity, You could just assume Slack's webhook format which would also integrate with Discord / RocketChat / Mattermost.

SinTan1729:
Support for webhooks will be extremely useful. I'm currently relaying the emails through an smtp to webhook app, but direct support will be nice. I'm willing to help with the implementation if needed.

SinTan1729:
It's great to see a PR for this. In the meantime, I'm using this.
https://github.com/SinTan1729/mailhook
It's just a fork of an existing project with some modifications so that it works with NTFY.

PR #1341, opened 2026-02-25
feat: Add event webhook notifications

Alfredooe:
Hello, Thought I'd have a crack at this issue https://github.com/pocket-id/pocket-id/issues/936 

I understand that this was originally decided as out of scope due to increase maintenance burden of handling multiple notification services. Decided I'd have a go at implementing this as simple as possible. This seems relatively low complexity and a decent QoL feature.

Adding a new webhook notification feature to application configuration allowing pocketid to send Discord, Slack, Generic webhooks when selected audit events occur. 

- Adding webhookUrl and WebhookEvents to appconfig 
- Adding webhookService to handle sending webhooks to target service, formatting depending on slack / discord url 
- Adding webhooks collapsible to application config panel 
- Adding meh playwright tests might need improvement

Open to any feedback / changes to help this get implemented, would like to see this functionality. I haven't written playwright tests before, so these are unapologetically vibecoded, will likely need to be improved.

> PwANbASsQq
> Discord_6xLhXVjkXB
>


backend/internal/service/audit_log_service.go
```diff
@@ -61,6 +63,24 @@ func (s *AuditLogService) Create(ctx context.Context, event model.AuditLogEvent,
 		return model.AuditLog{}, false
 	}
 
+	// Dispatch webhook notification asynchronously
+	go func() {
```
ItalyPaleAle, line 67: Should this be sent also in CreateNewSignInWithEmail?

backend/internal/service/audit_log_service.go
```diff
@@ -61,6 +63,24 @@ func (s *AuditLogService) Create(ctx context.Context, event model.AuditLogEvent,
...
+		innerCtx := context.WithoutCancel(ctx)
+		// Fetch user details if user ID is present
+		if auditLog.UserID != "" && auditLog.Username == "" {
+			var user model.User
+			if err := s.db.WithContext(innerCtx).Select("username", "email").First(&user, "id = ?", auditLog.UserID).Error; err == nil {
```
ItalyPaleAle, line 72: Ideally inside a closure we use different names for variables to avoid shadowing, so something like `rErr`

backend/internal/service/audit_log_service.go
```diff
@@ -61,6 +63,24 @@ func (s *AuditLogService) Create(ctx context.Context, event model.AuditLogEvent,
 		return model.AuditLog{}, false
 	}
 
+	// Dispatch webhook notification asynchronously
+	go func() {
+		innerCtx := context.WithoutCancel(ctx)
```
ItalyPaleAle, line 68: Because this runs in a goroutine, we need to use a background context.

The reason is that the context passed to `Create` is derived from the HTTP request, and it's canceled when the request ends. So this context in the goroutine will be canceled mid-flight.

Also note tracing propagation. See what we do below:

```go
		// We use a background context here as this is running in a goroutine
		//nolint:contextcheck
		go func() {
			span := trace.SpanFromContext(ctx)
			innerCtx := trace.ContextWithSpan(context.Background(), span)
```

backend/internal/service/audit_log_service.go
```diff
@@ -61,6 +63,24 @@ func (s *AuditLogService) Create(ctx context.Context, event model.AuditLogEvent,
 		return model.AuditLog{}, false
 	}
 
+	// Dispatch webhook notification asynchronously
```
ItalyPaleAle, line 66: We should avoid running this goroutine entirely if webhooks aren't configured for the event. Otherweise, we're running SQL queries unnecessarily.

Suggestion: add in WebhookService a method `IsEnabledForEvent(event string) bool`. Invoke that before spawning the goroutine, and if it returns false, don't spawn it.

backend/internal/service/webhook_service.go
```diff
@@ -0,0 +1,209 @@
...
+	Timestamp string              `json:"timestamp"`
+}
+
+type webhookEmbedField struct {
+	Name   string `json:"name"`
+	Value  string `json:"value"`
+	Inline bool   `json:"inline"`
+}
+
+type webhookPayload struct {
```
ItalyPaleAle, line 43: I think the kind of webhook should be configurable, rather than sending data twice? So for each webhook endpoint we should configure if it's slack-compatible or discord-compatible. There may be other types in the future.

backend/internal/service/webhook_service.go
```diff
@@ -0,0 +1,209 @@
...
+					{Name: "Device", Value: valueOrDash(auditLog.UserAgent), Inline: true},
+				},
+				Timestamp: auditLog.CreatedAt.UTC().Format(time.RFC3339),
+			},
+		},
+	}
+
+	// Add any extra data fields from the audit log.
+	for k, v := range auditLog.Data {
+		payload.Embeds[0].Fields = append(payload.Embeds[0].Fields, webhookEmbedField{
```
ItalyPaleAle, line 89: Please pre-allocate `payload.Embeds[0].Fields` with the desired capacity:

```go
    fields := make([]webhookEmbedField, 0, 4 + len(auditLog.Data))
    fields = append(fields, 
        {Name: "User", Value: valueOrDash(auditLog.Username), Inline: true},
	    {Name: "IP Address", Value: valueOrDash(ipAddress), Inline: true},
		{Name: "Location", Value: valueOrDash(location), Inline: true},
		{Name: "Device", Value: valueOrDash(auditLog.UserAgent), Inline: true},
    )
	payload := webhookPayload{
		Embeds: []webhookEmbed{
			{
				Title: formatEventTitle(string(auditLog.Event)),
				Color: 5814783, // A pleasant blue/purple color
				Fields: fields,
				Timestamp: auditLog.CreatedAt.UTC().Format(time.RFC3339),
			},
		},
	}
```

This way we avoid re-allocations

backend/internal/service/webhook_service.go
```diff
@@ -0,0 +1,209 @@
...
+		if strings.TrimSpace(allowed) == event {
+			return true
+		}
+	}
+
+	return false
+}
+
+// formatEventTitle converts an event constant like "SIGN_IN" to a title like "Sign In"
+func formatEventTitle(event string) string {
```
ItalyPaleAle, line 181: Can you move this to the utils package?

Even better, since the constants are a finite amount, we should have a function with a `switch` case that converts them, without doing it at runtime.

backend/internal/service/webhook_service.go
```diff
@@ -0,0 +1,209 @@
...
+// isEventAllowed checks if the given event is in the configured event filter.
+// If no filter is configured (empty string), all events are allowed.
+func (s *WebhookService) isEventAllowed(event string) bool {
+	cfg := s.appConfigService.GetDbConfig()
+	eventsFilter := cfg.WebhookEvents.Value
+	if eventsFilter == "" {
+		return true
+	}
+
+	for _, allowed := range strings.Split(eventsFilter, ",") {
```
ItalyPaleAle, line 171:
```suggestion
	for allowed := range strings.SplitSeq(eventsFilter, ",") {
```

ItalyPaleAle (changes requested):
I'll let the maintainers decide if they want this, but left a review meanwhile.

The thing is, webhooks are actually a lot more complex than one would think, and if we get them, we need to them more reliably :)

Two things:

1. Slack and Discord have rate-limits. By personal experience, they are very low, and they respond with 429's in that case, also including a Retry-After header.
2. Network failures happen a lot more than you would think, we need to be resilient to those.

To do webhooks properly and reliably:

1. We need to batch webhooks and send more than one message at a time if they are happening in a short burst. For example, batch up to 10 messages or 1s (whichever comes first), and then send a single webhook. This is very helpful to avoid rate-limits.
2. We need to manage 429's appropraitely, and retry the request after the specific amount of time
3. We need to handle network failures and retry in case of network errors or when you get a 5xx error. This can be done with the `github.com/cenkalti/backoff/v5` library which this package already uses.

My suggestion:

1. `WebhookService` internally should maintain a queue. Calling `SendEvent` performs a synchronous operation of adding messages to the queue (so you can also call `SendEvent` without a separate goroutine)
2. When the queue has X messages (e.g. 10) or after Y time (e.g. 1s) since the first `SendEvent`, then you send the actual webhook. This allows batching.
3. The code that sends the webhook needs to use the backoff library to retry in case of network failures or 5xx status codes.
4. In case of 429 responses, we need to wait for the requested time and retry after that

Alfredooe:
Thanks for the review @ItalyPaleAle , Yeah it's quite basic to demonstrate this. I understand previously this wasn't wanted because of added complexity, If this has changed and maintainers are interested in this happy to pick up the fixes and batching / queueing / retry logic. Will wait to see what they think.

kmendell:
Im not saying we will go with this, as im still not sure if its justfied i want to ask elias, but would it be better to use something like https://pkg.go.dev/github.com/nicholas-fedor/shoutrrr, which handles the webhook stuff?

Alfredooe:
That'd probably be a good idea wasn't aware this existed, Looks like Watchtower uses it decently proven.

Alfredooe:
Any interest in this one? Happy for it to be closed off if it's out of scope

PR #1342, opened 2026-02-26
fix: federated client credentials not working if sub ≠ client_id

### LINKED ISSUE #902: 🐛 Bug Report: Federated client credentials RFC 7523 inconsistency

### LINKED ISSUE #1333: 🐛 Bug Report: Federated Credentials: Invalid client assertion when subject is not client ID
### Reproduction steps

1. Configure an OIDC client
2. Configure [Federated Client Credentials](https://pocket-id.org/docs/guides/oidc-client-authentication)
3. Put any string (other than the Client ID) as the Subject
4. Have your issuer issue a token, with `sub` equal to the configured subject in pocket-id.
5. Try do an OAuth token exchange with `grant_type=authorization_code` as per the docs.
6. Observe the error message:
```
{"error":"Invalid client assertion"}
```

### Expected behavior

Pocket ID should consider the token valid if the subject is the string configured as the subject under Federated Client Credentials config.  

### Actual Behavior

Before validating the token, the program extracts the subject from the token, assumes it is the client ID, and then looks up the client from the database with the token sub. 

https://github.com/pocket-id/pocket-id/blob/ae269371da95cd55f94ad8862822a114fe9d166d/backend/internal/service/oidc_service.go#L1800

Of course, if the `sub` is not the client ID, then the lookup will fail, thus resulting in the final error message, "Invalid client assertion"

### Pocket ID Version

2.2.0

### Database

SQLite

### OS and Environment

Kubernetes. Deployment manifest with image `ghcr.io/pocket-id/pocket-id:v2.2.0`

### Log Output

```
Feb 23 18:06:42 ERR record not found app=pocket-id version=2.2.0 error="record not found" query="SELECT * FROM `oidc_clients` WHERE id = \"project_path:my-gitlab-group/my-repo:ref_type:branch:ref:my-branch-name\" ORDER BY `oidc_clients`.`id` LIMIT 1" duration=809.287µs rows=0 file=github.com/orandin/slog-gorm@v1.4.0/logger.go:163
Feb 23 18:06:42 WRN Request with errors: Error #01: invalid client assertion
app=pocket-id version=2.2.0 status=400 method=POST path=/api/oidc/token query="" route=/api/oidc/token ip=10.42.2.1 latency=1.857896ms referer="" user_agent=curl/8.17.0 body_size=36
```

For sake of interest, I'm trying exchange a [Gitlab pipeline token](https://docs.gitlab.com/ci/secrets/id_token_authentication/) with Pocket ID.

ItalyPaleAle:
Fixes a regression that broke federated client credentials when the "sub" is not the same as the client ID.

See [this comment](https://github.com/pocket-id/pocket-id/issues/902#issuecomment-3961460340) with implementation notes and decisions around framing this in terms of identity federation, which is on purpose not compliant with RFC 7532 (however, it's possible to configure this to be fully RFC-7532 compliant)

- Fixes #1333
- Closes #902
- Closes #991

### DIFF
--- a/backend/internal/common/errors.go
+++ b/backend/internal/common/errors.go
@@ -20,7 +20,7 @@ type AlreadyInUseError struct {
 func (e *AlreadyInUseError) Error() string {
 	return e.Property + " is already in use"
 }
-func (e *AlreadyInUseError) HttpStatusCode() int { return http.StatusUnauthorized }
+func (e *AlreadyInUseError) HttpStatusCode() int { return http.StatusBadRequest }
 
 func (e *AlreadyInUseError) Is(target error) bool {
 	// Ignore the field property when checking if an error is of the type AlreadyInUseError
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -664,7 +664,7 @@ func (oc *OidcController) updateAllowedUserGroupsHandler(c *gin.Context) {
 }
 
 func (oc *OidcController) deviceAuthorizationHandler(c *gin.Context) {
-	// Per RFC-6749, parameters passed to the /token endpoint MUST be passed in the body of the request
+	// Per RFC 8628 (OAuth 2.0 Device Authorization Grant), parameters for the device authorization request MUST be sent in the body of the POST request
 	// Gin's "ShouldBind" by default reads from the query string too, so we need to reset all query string args before invoking ShouldBind
 	c.Request.URL.RawQuery = ""
 
--- a/backend/internal/service/oidc_service.go
+++ b/backend/internal/service/oidc_service.go
@@ -1768,7 +1768,7 @@ func (s *OidcService) verifyClientAssertionFromFederatedIdentities(ctx context.C
 	// (Note: we don't use jwt.WithIssuer() because that would be redundant)
 	_, err = jwt.Parse(assertion,
 		jwt.WithValidate(true),
-		jwt.WithVerify(true),
+
 		jwt.WithAcceptableSkew(clockSkew),
 		jwt.WithKeySet(jwks, jws.WithInferAlgorithmFromKey(true), jws.WithUseDefault(true)),
 		jwt.WithAudience(audience),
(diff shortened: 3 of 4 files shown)


ItalyPaleAle:
@copilot can you please update the E2E tests, which are currently failing, so they work with the updated code and rules? 🙏

backend/internal/dto/oidc_dto.go
```diff
@@ -98,7 +98,8 @@ type OidcCreateTokensDto struct {
 }
 
 type OidcIntrospectDto struct {
-	Token string `form:"token" binding:"required"`
+	Token    string `form:"token" binding:"required"`
+	ClientID string `form:"client_id"`
```
kmendell, line 102: Should the client ID be required as well? This wouldnt work without a client id would it?

kmendell (approved):
Besides that one comment, It looks good to me :)

backend/internal/dto/oidc_dto.go
ItalyPaleAle: Most people calling this endpoint would pass the client ID (and secret) as a Basic auth header. This is just an option for when using federation

PR #1344, opened 2026-02-26
perf: frontend performance optimizations

ItalyPaleAle:
Introduces two optimizations to improve the performance of serving the frontend application, which were measured using Lighthouse reports.

1. Serves compiled assets by SvelteKit with a much longer `Cache-Control` and the `immutable` tag. Because these compiled assets have a version number at the end, they are immutable even when Pocket ID is updated.
2. Enables serving compiled assets that are pre-compressed with gzip or brotli. This is done by adding an extra step in the frontend compilation that generates pre-built br and gz files, which are then served to clients when available.
   - This reduces the total downloaded size of the sign in page by approx 300KB (with a fresh cache)


kmendell:
I thought vite automatically compressed the built files? I could be entirley wrong though.

kmendell:
We could make a servcie worker to do this too vs having the backend do it, that was its client side not server side?

ItalyPaleAle:
> I thought vite automatically compressed the built files? I could be entirley wrong though.

Sort of, vite minifies JS and CSS but that's different from transport-level encoding. Basically with [Content-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Encoding), the data sent over the wire is compressed with gzip or brotli, and because JS and CSS are text-based formats, they compress well. Often that is something done at the server or proxy layer (e.g. CDN), but if people don't have a CDN in front of Pocket ID they don't get that.

> We could make a servcie worker to do this too vs having the backend do it, that was its client side not server side?

That would be a nice thing to add too. I don't think they are mutually exclusive.

PS: I have discovered these optimizations by running a Lighthouse report with Chrome. It was suggesting to increase the cache lifetime for immutable assets and to enable compression.

stonith404 (approved):
Thanks!

ItalyPaleAle:
Tests are failing in main (see #1332) and are unrealted from this PR

PR #1360, opened 2026-03-04
feat: add support for response_mode=form_post

### LINKED ISSUE #1357: 🚀 Feature: add support for response_mode=form_post
### Reproduction steps

Hi, when a request for authentication is sent to the server with response_mode=form_post parameter, when the user authenticates a get redirect is done instead of a post.

I dont know if this is defined in RFC's, but our backend app is expecting a POST request instead of a get redirect.

### Expected behavior

It should respect the app responsemode

### Actual Behavior

It always does a get redirect, not a form_post

### Pocket ID Version

Latest, 2.3.0

### Database

SQLite

### OS and Environment

Ubuntu ARM

### Log Output

Not relevant

Johnwulp:
Adding support for response_mode form_post. 

- Added fallback logic to check query parameters when context value isn't set
- CSP Headers are adjusted to allow the browser to submit the form to the client's redirect URI

Closes #1357


backend/internal/middleware/csp_middleware.go
```diff
@@ -23,17 +23,39 @@ func GetCSPNonce(c *gin.Context) string {
...
+				formAction = "'self' " + redirectURI
+			}
+		}
+
 		csp := "default-src 'self'; " +
 			"base-uri 'self'; " +
 			"object-src 'none'; " +
 			"frame-ancestors 'none'; " +
-			"form-action 'self'; " +
+			"form-action " + formAction + "; " +
```
ItalyPaleAle: Nit: since "self" is always included, leave it here, and then just use `formAction` to add _additional_ targets

backend/internal/dto/oidc_dto.go
```diff
@@ -71,6 +71,7 @@ type AuthorizeOidcClientRequestDto struct {
 	CodeChallenge         string `json:"codeChallenge"`
 	CodeChallengeMethod   string `json:"codeChallengeMethod"`
 	ReauthenticationToken string `json:"reauthenticationToken"`
+	ResponseMode          string `json:"responseMode"`
```
ItalyPaleAle: Should probably add a validator to ensure that only the modes we support are used?

ItalyPaleAle (changes requested):
Thank you! Overall LGTM, but can you please add some E2E tests to validate this? It's a new login flow so we must make sure we have tests

kmendell:
I agree with @ItalyPaleAle suggestions above and the e2e tests as well.

Johnwulp:
I simpliefied the CSP middeware to always include self, added validation for responsemode.

For the tests, im not familiar with hwo to set that up correctly. The current suggestion is highly AI inspired. Please tak a close look at this part.

backend/internal/dto/validations.go
```diff
@@ -68,3 +74,15 @@ func ValidateCallbackURL(raw string) bool {
...
+// ValidateResponseMode validates response_mode parameter
+// If responseMode is present, it must be "form_post"
+// Empty responseMode is allowed (will use default behavior)
+func ValidateResponseMode(responseMode string) bool {
+	// Empty responseMode is allowed (field not provided, use default)
+	if responseMode == "" {
+		return true
+	}
+	// If present, it must be form_post
+	return responseMode == "form_post"
```
ItalyPaleAle: We should allow 'code' and 'fragment' too, no?

```suggestion
	switch responseMode {
    case "query", "fragment", "form_post":
        return true
    case "":
	    // Empty responseMode is allowed (field not provided, use default)
		return true
    default:
        // Unsupported value
        return false
	}
```

backend/frontend/frontend_included.go
```diff
@@ -91,6 +91,15 @@ func RegisterFrontend(router *gin.Engine, rateLimitMiddleware gin.HandlerFunc) e
 		}
 
 		if path == "index.html" {
+			// Check if this is an OAuth2 authorization request with response_mode=form_post
+			// In that case, we need to allow form submissions to the redirect_uri
+			responseMode := c.Query("response_mode")
+			redirectURI := c.Query("redirect_uri")
```
ItalyPaleAle: Upon further review, I think we need to validate the passed `redirect_uri` against the allowlist for the client.

Otherwise, this allows passing arbitrary values on all pages that inject allowed form actions in the CSP

ItalyPaleAle (changes requested):
Sorry, a few more changes upon another review.

backend/frontend/frontend_included.go
Johnwulp: That is possible, but requires the frontend to inject the context oidc service I suppose. I will try to write a function for that.

backend/internal/dto/validations.go
Johnwulp: I think you mean query and fragment. These are the supported response_mode options. The code is a response_type, not to confuse with the mode the code is delivered.

**form_post**
As i understand from https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html the response_mode=form_post should execute a post

**query**
This is already implemented as the default, so this is done when no response_mode parameter is set. We could enhance it when the response_mode=query that the default ouput is given.

**fragment**
As i understand it, the server redirects with parameters in the URL fragment (#). I don't have a use-case for this,

Example would be: 
GET /callback#access_token=2YotnFZFEjr1zCsicMWpAA&state=xyz HTTP/1.1

backend/frontend/frontend_included.go
Johnwulp: Please double check this validation logic.

backend/internal/dto/validations.go
ItalyPaleAle: Yes, sorry I meant "query" and "fragment".

backend/frontend/frontend_included.go
```diff
@@ -54,7 +57,28 @@ func init() {
...
+// validateRedirectURI validates that the redirect_uri is in the client's allowed callback URLs
+func validateRedirectURI(ctx any, oidcService *service.OidcService, clientID, redirectURI string) (bool, error) {
+	client, err := oidcService.GetClient(ctx.(context.Context), clientID)
+	if err != nil {
+		return false, err
+	}
+
+	// If the client has no callback URLs configured, reject the redirect URI
+	if len(client.CallbackURLs) == 0 {
+		return false, fmt.Errorf("client has no callback URLs configured")
```
ItalyPaleAle: Nit

```suggestion
		return false, errors.New("client has no callback URLs configured")
```

ItalyPaleAle (commented):
redirect_uri handling LGTM now

There's a few open comments left, but the biggest one would be that we need some testing for this, ideally E2E, at least for the auth flow. It's too in the critical path to not have some validation...

backend/internal/dto/validations.go
Johnwulp: I did some more reading on this topic. The fragment response type delivers the tokens directly to the end-user’s browser. The `#` character ensures that the tokens are never sent to the server. This is called the implicit flow. As far as I can see, this project doesn’t support it.

I will extend the current changes to include the query parameter as well and produce the already present output.

backend/frontend/frontend_included.go
Johnwulp: Changed this, Added "errors" to the imports as well

backend/internal/middleware/csp_middleware.go
```diff
@@ -23,17 +23,31 @@ func GetCSPNonce(c *gin.Context) string {
...
-		// Generate a random base64 nonce for this request
 		nonce := generateNonce()
 		c.Set("csp_nonce", nonce)
 
+		// Determine if there is an EXTRA target beyond 'self'
+		var extraAction string
+		if v, ok := c.Get("csp_allowed_form_action"); ok {
+			extraAction, _ = v.(string)
+		} else if c.Query("response_mode") == "form_post" {
+			extraAction = c.Query("redirect_uri")
```
ItalyPaleAle: This should be removed, right? Because it does allow injecting a redirect URI. Instead it should just come from the "csp_allowed_forma_action" in the context?

Johnwulp: Yes, indeed. I removed the else condition, so the CSP form-action can only be extended via values set via SetAllowedFormAction, which is called by handlers after validating the redirect URI against the client's registered allowlist.

Johnwulp:
Hi, are there still things that need to be changed?

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -116,7 +110,16 @@ func registerGlobalMiddleware(r *gin.Engine) {
 	r.Use(middleware.NewErrorHandlerMiddleware().Add())
 }
 
-func registerRoutes(r *gin.Engine, db *gorm.DB, svc *services) {
+func registerRoutes(r *gin.Engine, db *gorm.DB, svc *services) error {
+	frontendRateLimitMiddleware := middleware.NewRateLimitMiddleware().Add(rate.Every(100*time.Millisecond), 300)
```
ItalyPaleAle: I don't think we need a rate-limiter for the middleware, since it's just serving static assets and it's not a protected resource. Please revert

ItalyPaleAle: I see the reason for the rate limiter is because now there can be a DB query if there's a form_post... Can we limit the rate-limiter middleware to that scenario, and not rate limit everything else?

PR #1362, opened 2026-03-05
fix: various fixes in background jobs

ItalyPaleAle:
This was originally an attempt for #1356, but it doesn't seem like it's addressing the problem (whose cause is still unclear). However, it still offers a number of important fixes and improvements.

Includes:

- Fixed API key expiration emails not working (template failed to be compiled)
- Ensure all cleanup jobs are executed with some jitter, so they don't all run at the same time
- Improve logging in app lock service while renewing locks
- Added missing DB indexes for tables that have daily cleanups
- Cleanup jobs are automatically retried in case of transient failure, instead of crashing the application

### DIFF
--- a/tests/resources/export/database.json
+++ b/tests/resources/export/database.json
@@ -53,7 +53,7 @@
 				"user_id": "f4b89dc2-62fb-46bf-9f5f-c34f4eafe93e"
 			},
 			{
-				"client_id": "7c21a609-96b5-4011-9900-272b8d31a9d1",
+				"client_id": "c48232ff-ff65-45ed-ae96-7afa8a9b443b",
 				"code": "federated",
 				"code_challenge": null,
 				"code_challenge_method_sha256": null,


kmendell:
I'll take a deeper look tomorrow

backend/internal/job/scheduler.go
ItalyPaleAle: @copilot In RegisterJob, refactor the signature to be:

```go
type RegisterJobOpts struct {
	RunImmediately bool
	ExtraOptions []gocron.JobOption
	BackOff backoff.BackOff
}

func (s *Scheduler) RegisterJob(ctx context.Context, name string, def gocron.JobDefinition, job func(ctx context.Context) error, opts RegisterJobOpts)
```

Note the addition of BackOff, which is optional. If `BackOff != nil`, then inside RegisterJob, wrap `job` into a function that uses `github.com/cenkalti/backoff.RetryNotify` with the provided BackOff object to automatically retry in case of transient errors.

Use the new BackOff option for the DB cleanup jobs, so that a query failure will be retried and not cause a crash immediately

ItalyPaleAle:
This doesn't seem like the solution to #1356, but I think it still includes some very useful fixes

backend/internal/job/api_key_expiry_job.go
```diff
@@ -37,12 +37,16 @@ func (j *ApiKeyEmailJobs) checkAndNotifyExpiringApiKeys(ctx context.Context) err
 	}
 
 	for _, key := range apiKeys {
-		if key.User.Email == nil {
+		if key.User.Email == nil || key.User.ID == "" {
```
stonith404: Is this additional condition really necessary? If the email is set the ID is set as well in every case.

backend/internal/job/db_cleanup_job.go
```diff
@@ -7,28 +7,28 @@ import (
...
 	return errors.Join(
-		s.RegisterJob(ctx, "ClearWebauthnSessions", def, jobs.clearWebauthnSessions, true),
-		s.RegisterJob(ctx, "ClearOneTimeAccessTokens", def, jobs.clearOneTimeAccessTokens, true),
-		s.RegisterJob(ctx, "ClearSignupTokens", def, jobs.clearSignupTokens, true),
-		s.RegisterJob(ctx, "ClearEmailVerificationTokens", def, jobs.clearEmailVerificationTokens, true),
-		s.RegisterJob(ctx, "ClearOidcAuthorizationCodes", def, jobs.clearOidcAuthorizationCodes, true),
-		s.RegisterJob(ctx, "ClearOidcRefreshTokens", def, jobs.clearOidcRefreshTokens, true),
-		s.RegisterJob(ctx, "ClearReauthenticationTokens", def, jobs.clearReauthenticationTokens, true),
-		s.RegisterJob(ctx, "ClearAuditLogs", def, jobs.clearAuditLogs, true),
+		s.RegisterJob(ctx, "ClearWebauthnSessions", jobDefWithJitter(24*time.Hour), jobs.clearWebauthnSessions, service.RegisterJobOpts{RunImmediately: true, BackOff: backoff.NewExponentialBackOff()}),
```
stonith404: In my opinion the default configuration of backoff makes too many retries. Wouldn't a multiplier of `2` make more sense? We would still have 8 retries in a minute.

backend/internal/job/api_key_expiry_job.go
ItalyPaleAle: I don't think it is, but the old code was retrieving the user again inside the called method, and then was checking for `key.User.ID` too. So I kept it here. But it's likely redundant.

backend/internal/job/db_cleanup_job.go
ItalyPaleAle: Changed, this should repeat 4 times I think

stonith404:
@ItalyPaleAle The e2e tests are failing on the main branch, this is not related to this PR. I'm looking into it.

PR #1370, opened 2026-03-08
fix: better error messages when there's another instance of Pocket ID running

ItalyPaleAle:
See https://github.com/pocket-id/pocket-id/issues/1274#issuecomment-4015888747


ItalyPaleAle:
CC: @akosiaris thanks for the suggestion :)

stonith404 (approved):
Thanks 👍

PR #1372, opened 2026-03-08
feat: add TRUSTED_PLATFORM environment variable for gin

choyri:
The current `TRUST_PROXY` environment variable can only control gin's `SetTrustedProxies` logic. However, gin's ClientIP logic is quite complex and cannot obtain the accurate client IP in some scenarios.

This PR is similar to #265, but it solves the problem using a more convenient variable, [see here](https://github.com/gin-gonic/gin/blob/3e44fdc4d1636a2b1599c6688a76e13216a413dd/context.go#L977) .

### DIFF
--- a/backend/internal/bootstrap/router_bootstrap.go
+++ b/backend/internal/bootstrap/router_bootstrap.go
@@ -49,6 +49,10 @@ func initRouter(db *gorm.DB, svc *services) (utils.Service, error) {
 		_ = r.SetTrustedProxies(nil)
 	}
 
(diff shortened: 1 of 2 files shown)


kmendell:
Would you mind add some info on how to use this in the website documentation as well? As it looks like Gin only suports Cloudlfare, Fly.io, and Google, so any other providers or platform wouldnt be used

choyri:
@kmendell I have added some usage instructions: pocket-id/website#234

Issue #1269, opened 2026-01-24
🐛 Bug Report: (LDAP) Checking membership for admin group is inconsistent with populating group memberships


chipschipschips:
### Reproduction steps

This sets up an LDAP account `testadmin` and LDAP group `pocketid.admin`. `pocketid.admin` will be the PocketID admin group.

- Create `testadmin` user account in LDAP
  - DO NOT add a `memberOf` attribute for the `pocketid.admin` group
- Create the `pocketid.admin` group in LDAP
  - DO add a `member` attribute for the `testadmin` user
- Configure PocketID to sync from LDAP
  - Set `LDAP_ADMIN_GROUP_NAME` to `pocketid.admin`
- Sync from LDAP

Full repro is available [here](https://github.com/chipschipschips/repro-pocketid-isadmin-ldap-bug).

### Expected behavior

- The `testadmin` user should have admin privileges in PocketID
  - The UI should show the admin pages
  - The `is_admin` field in the `users` table should be set to `true`

### Actual Behavior

- The `testadmin` user does not have admin privileges in PocketID
  - The UI does not show any admin pages
  - The `is_admin` is `false` in the `users` table in the database
- The `testadmin` user does show as a member of the `pocketid.admin` group in the `membership` table in the database
  - (and in the UI if you can log in with an actual admin user; eg: created before enabling LDAP)

### Pocket ID Version

v2.2.0

### Database

SQLite and Postgres 18.

### OS and Environment

Docker compose on Docker Desktop for macOS, served using Traefik.

### Log Output

Logs don't show anything interesting.

```
pocketid-1  | Jan 24 06:05:58 INF Pocket ID is starting app=pocket-id version=2.2.0
pocketid-1  | Jan 24 06:05:58 INF Connected to database app=pocket-id version=2.2.0 provider=sqlite
pocketid-1  | Jan 24 06:05:58 WRN MAXMIND_LICENSE_KEY environment variable is empty: the GeoLite2 City database won't be updated app=pocket-id version=2.2.0
pocketid-1  | Jan 24 06:05:58 INF Acquired application lock app=pocket-id version=2.2.0 process_id=1 host_id=1eea36e92906
pocketid-1  | Jan 24 06:05:58 INF Starting job scheduler app=pocket-id version=2.2.0
pocketid-1  | Jan 24 06:05:58 INF Server listening app=pocket-id version=2.2.0 addr=0.0.0.0:1411
[...]

PR #1374, opened 2026-03-08
fix: derive LDAP admin access from group membership

kmendell:
Fixes: https://github.com/pocket-id/pocket-id/issues/1269

Instead of applying Admin memebership from querying ldap it derives it from the group membership snapshot instead.


kmendell:
@coderabbitai review

Issue #1273, opened 2026-01-26
🐛 Bug Report: Custom Claims with same key overrides previous defined values


etho201:
### Reproduction steps

If a user belongs to multiple groups, and you assign a custom claim to each group, the key-value pairs seem to conflict when the same key (but different value) is defined in more than one group.

For example:
I have Pocket-ID configured with both OCIS and OpenCloud. OpenCloud is a fork of OCIS so the way it is configured with OIDC is nearly identical. The difference is you have a group called `ocisAdmin`, and a group called `opencloudAdmin`. For the `ocisAdmin` group you are to add a custom claim key-value pair called `roles:ocisAdmin`, and for the `opencloudAdmin` group you are to add a custom claim key-value pair called `roles:opencloudAdmin`.

After doing this, if you click on `OIDC Data Preview" you will only see one of the values (probably whichever one was added most recently).

The only way around this is to instead define the custom claims at the user level instead, and it needs to be defined in a very specific way. So if I want my admin user to have a key called `roles` with two values: `ocisAdmin` and `opencloudAdmin`, I need to define it within an array list. The key is `roles`, the value needs to be defined as `["ocisAdmin", "opencloudAdmin"]`.

I'm happy to have found a workaround, but I think it would be far more elegant if Pocket ID could simply combine any key-value pairs that have the same key into an array list, rather than wiping out the previously defined key-value pair.

### Expected behavior

I expected that if I define a key-value pair in a group, if the key is the same as another key defined in a different group, the user profile assigned to those groups would have both values for that key.

### Actual Behavior

I define a key-value pair in a group, if the key is the same as another key defined in a different group, the user profile assigned to those groups only has the latest value for that key (as verified in the OIDC Data Preview).

### Pocket ID Version

v2.2.0

### Database

SQLite

### OS and Environment

Docker on Armbian 25.11.2 jammy

kiaraly:
Would a setting per claim work that changes the behavior? You could have something like aggregate that would take all the different values and process them as a list and first/last that sorts them first and then picks the respective option.

That would also make things like setting a quota over different groups work and then selecting the highest option.

Issue #1361, opened 2026-03-04
🚀 Feature: Per-Client Username Claim Override


barolinecrewer:
### Problem

When integrating Pocket ID with services that map OIDC identity to local user accounts by username (e.g. Unraid, which maps `preferred_username` to a local system account), there is currently no way to send a different username claim value to a specific client without changing your global Pocket ID username.

This creates a conflict when your Pocket ID username doesn't match the expected account name on the target service. For example, a user named `john` in Pocket ID who needs to authenticate as `root` on Unraid. The only workarounds are:
- Changing your global Pocket ID username to match the target service (affects all other clients)
- Creating a dummy local account on the target service to match the Pocket ID username

### Proposed Solution

Add a per-client claim override configuration that allows administrators to define a static value or an alternative user attribute to use for specific claims when issuing tokens to a given client.

For example, on a per-client basis:
- Override `preferred_username` with a static value (e.g. `root`)
- Or map `preferred_username` to a different user attribute (e.g. an custom attribute field)

### Use Case

A user authenticates as `john` across all Pocket ID-connected apps, but a specific client (Unraid) requires the `preferred_username` claim to be `root` to match its local admin account. A per-client override would allow this mapping without any global username change.

### Alternatives Considered

- Changing the global Pocket ID username: works but breaks the username for all other clients
- Creating a matching local account on the target service: works but requires managing redundant accounts on every affected service

### Additional Context

Some other identity providers support this via claim mapping or transformation rules scoped to individual clients/applications. Examples include Authentik (property mappings per provider) and Keycloak (protocol mappers per client).

Silther:
I have the same issue with unraid.
But it would be useful for more apps too.
When self hosting I try to separate my main account from my admin account but with some apps that's not necessary or even possible.
In Authentik it was possible to overwrite oidc claims (or have custom scopes), with those you could do things like:
```
if user.email == admin@domain.com
  email = main@domain.com
else
  email = user.email
```

Issue #1278, opened 2026-01-27
🚀 Feature: Allow multiple email addresses binding


Lucas04-nhr:
### Feature description

I wonder if it's possible to bind multiple email addresses to one user. During the authentication process, the user can choose which email address would be provided.

### Pitch

I want to use different emails to log into different apps, such as Cloudflare Zero Trust. However, I don't want to register multiple user accounts.

Issue #807, opened 2025-08-02
🚀 Feature: Claim field override or custom per service mapping


Ulrar:
### Feature description

Some basic templating support in the Custom Claims section of the user, to override fields such as email with something like `$serviceName`. There may be a better way of doing this, that's just the first thing that came to mind.

### Pitch

I realize this is a bit of a niche use case, but I use a different email address per service. Currently PocketID just sends whatever it has as the user's email to the apps in the claim, which mean all of these apps are using the PocketID specific email address of my user, instead of their own.

But I'm sure there's other use cases for templated custom mapping fields.

ItalyPaleAle:
Would #781 be a way for you to implement this?

Ulrar:
@ItalyPaleAle there's already a way to add custom fields from the UI, but as far as I know it only takes plain text. I suppose fetching these from http may work but then I'd have to make and maintain an api for that, it's not ideal

ItalyPaleAle:
Indeed, having to maintain an API is the price you pay.

The benefits are that it's a lot more flexible (you can interact with external DBs, for example), and it doesn't require implementing a new DSL, which would result in a lot of things asked for the future.

> An alternative I considered for #781 was suggesting the inclusion of a built-in engine, for example using [wazero](https://github.com/tetratelabs/wazero), but that comes with other complexities

h3nnes:
Stumbled across this issue when I tried setting up Mailcow with Pocket-ID (via [generic-OIDC](https://docs.mailcow.email/manual-guides/mailcow-UI/u_e-mailcow_ui-generic-oidc/)). The email address configured in my Pocket-ID profile does not match the mail ccount address in mailcow, so I was looking for a way to override the email claim to the mail address of my Mailcow mail account.

With a bit of help through non-biological intelligence, I've created a PoC with claim remapping capability. There are three ways to override claims for an individual app:

1. By a user field claim value (like first name, last name, etc.)
2. By a custom claim defined for the user
3. By a custom static value

The feature is located in the advanced configuration section inside an OIDC-client. [Docker image available here for testing](https://github.com/h3nnes/pocket-id/pkgs/container/pocket-id/995139918?tag=next).

For my usecase to override the email claim, it works like a charm.

<img width="2113" height="640" alt="Image" src="https://github.com/user-attachments/assets/fe1fef4f-412c-4804-b930-d551a1294290" />

Ulrar:
That looks interesting, are you going to open a PR for it ?

[...]

PR #1409, opened 2026-04-03
feat: add custom claims mapping for OIDC clients

ghoudmon:
## Summary

Add per-client custom claims mapping to allow administrators to configure standard claims (like "email") to return values from different sources.

## Motivation

Some client requires email as preferred_username (#807),  others don't support email as preferred username, others requires static value (#1361) or you could want to communicate a secondary email  (#1278). 

 ## Technical details

This PR is derivated from the [branch](https://github.com/h3nnes/pocket-id/tree/custom-claims-remapping) of  @h3nnes described in #807.

### Backend changes

* Add `OidcClientClaimMapping` model types 
* Add `OidcClientClaimMappingDto` DTO and update `OidcClientDto` and `OidcClientUpdateDto` to add it as attribute
* update `oid_service` to :
  * `validateClaimMappings`: validate the source value regarding source type and reserved claims
  * implement claim mapping logic in `applyClaimMappings`
* Add SQL migration scripts

### Frontend changes

* Add TypeScript types for claim remappings in `oidc/type.ts`
* Create `claim-mapping-list-input.svelte` and `claim-mapping-input.svelte` component
* Integrate remapping UI into OIDC client form
* Add form validation and error handling
* Add `en` and `fr translations

## Test Results

* Unit test add in `oidc_service_test.go`
* Mapping availabled in token preview

image


GameTec-live:
Hey, randomly happened to stumble upon this PR, a few questions / a bit of feedback:

- This is when creating a OIDC Client and defines overwrite rules for that client, right?
- How does this interact with custom claims on users or groups? Which one takes priority?
- Could it maybe make sense to merge both Source/Value into a tree like thing? (See example below but with "User Attribute -> Displayname", etc, just an idea)
{360E471A-6B09-4121-911E-671D602ECB2E}

ghoudmon:
@GameTec-live 

- Yes, the mapping is defined during OIDC client creation/update, and it overwrites claims only for that client.
- The priority between group and user doesn't change: the user custom claims overwrite the group ones. 
- The mapping is applied during token generation after the computation of custom claims for one user. If a mapping has the same key than a custom claim, it overwrites the custom claim.
- About the UI suggestion, I wanted to keep it homogeneous with the rest of the application. I didn’t see a tree-structured field as suggested.

GameTec-live:
@ghoudmon 
Thanks, that clears it up! (I guess that would have been explained in the docs too?)

The frontend is using shadcnui tho and that picture is from the shadcnui documentation / examples. So itd fit in with the design system, but idk. Was just an idea as that may be a bit neater than two separate fields.

kiaraly:
> The priority between group and user doesn't change: the user custom claims overwrite the group ones.

What happens if two groups have the same custom claim? Does it just overwrite the value with the last group? Do you think the mapping mechanism would be a good place to customize that behavior and fix https://github.com/pocket-id/pocket-id/issues/1273?

ghoudmon:
> > The priority between group and user doesn't change: the user custom claims overwrite the group ones.
> 
> What happens if two groups have the same custom claim? Does it just overwrite the value with the last group? Do you think the mapping mechanism would be a good place to customize that behavior and fix #1273?

In my opinion, we should keep the PRs to the smallest scope possible. The issue mentioned is also a bit unclear to me in terms of how to resolve it, and it involved other parts of the code. In fatc, the claim mapping customization takes place after all custom claims  (user and groups) have been merged.

kiaraly:
I didn't mean to add it to this PR, I agree with you that PRs should stay as small as possible. I was more interested if this would be a good place to build on, but if this takes place after claims have been merged then it's not possible.

jacek-dargiel:
Hey there 👋

Just wanted to bump this up because it's a great feature I'd love to use, but it looks like it's stuck for no apparent reason.

Thanks for the great work.

Leseratte10:
I would be interested in having that feature as well, to be able to use it with applications where multiple different Pocket ID users should be able to log into a shared user account in the application.

PR #1422, opened 2026-04-11
fix(ldap): resolve posixGroup memberUid as bare usernames (#1408)

gucong3000:
## Summary

Fixes LDAP group sync when groups use **`posixGroup`** and the configured member attribute is **`memberUid`**. Those entries store **plain uid strings** (e.g. `alice`, `bob`), not member DNs like `groupOfNames` / `groupOfUniqueNames`.

`resolveGroupMemberUsername` previously fell through to a base-object LDAP search using the raw value as a DN, which fails for bare usernames and left groups empty or incomplete.

## Changes

- After DN cache lookup and RDN username extraction, if the member value **does not parse as a DN** (`ldap.ParseDN`), treat it as the **username** directly (Unicode NFC normalized), matching OpenLDAP / RFC2307-style `memberUid` semantics.
- Add regression test `TestLdapServiceSyncAllMapsPosixGroupMemberUid` covering `(objectClass=posixGroup)` + `memberUid` with multiple values (unknown uids are still skipped when no matching user exists).

## Testing

- `go test ./internal/service/ -run Ldap -count=1`

Closes #1408

### DIFF
--- a/backend/internal/service/ldap_service.go
+++ b/backend/internal/service/ldap_service.go
@@ -391,6 +391,12 @@ func (s *LdapService) resolveGroupMemberUsername(ctx context.Context, client lda
 		return norm.NFC.String(username)
 	}
 
+	// posixGroup (and similar) stores bare usernames in memberUid, not DNs. Treat any value
+	// that is not a valid DN as the username directly — see https://github.com/pocket-id/pocket-id/issues/1408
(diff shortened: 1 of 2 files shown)


stonith404 (approved):
LGTM, thanks!

jadev-projects:
Thank you for promptly taking care of the issue.

PR #1425, opened 2026-04-12
chore(translations): update translations via Crowdin

stonith404:


ItalyPaleAle (approved):

frontend/messages/cs.json
```diff
@@ -106,6 +106,7 @@
 	"ip_address": "IP adresa",
 	"device": "Zařízení",
 	"client": "Klient",
+	"actor": "Herec",
```
jose-d, line 109: hey, czech speaker here. this must be some glitch. Indeed in czech language "actor" is "herec". But in this context it is nonsense.

PR #1429, opened 2026-04-15
feat: add TLS support for HTTP/2 server

IngmarStein:
Add configuration options for serving over HTTPS with HTTP/2:
- TLS_CERT_FILE / TLS_KEY_FILE: Certificate and private key file paths

When TLS cert/key are provided, the server uses crypto/tls with HTTP/2. Otherwise, it falls back to h2c (unencrypted HTTP/2).

Implements RFC 7540 HTTP/2 protocol negotiation over TLS with ALPN.

Minimum TLS version is 1.3


kmendell:
I requested @ItalyPaleAle to look as im a bit busy with some family stuff at the moment, and elias is still in the military till may i belive so he has limited time, I glanced at it and it looked okay, but nowhere near a full review of it.

ItalyPaleAle:
@IngmarStein thanks for this PR.

For some context, I think we recommend users to expose Pocket ID behidn a reverse proxy, hence why we didn't add TLS support natively.

However, I think adding it can still be helpful.

My biggest feedback is around the implementation. Currently, it reads the certificate when the server starts, but then it doesn't support any hot reloading. Given that certs' validity is getting shorter and shorter, and that many people would want to use something like Let's Encrypt, IMHO hot reload would be helpful.

If it helps, this is code I've used in other projects: https://github.com/ItalyPaleAle/traefik-forward-auth/blob/main/pkg/server/tls-certs.go

IngmarStein:
@ItalyPaleAle thanks for the quick feedback and for acknowledging the usefulness of TLS even in internal networks. While less important behind a reverse proxy, it's still nice to always have traffic encrypted in transit, even in a LAN.

I agree that hot-reloading is a very nice property given the ever-decreasing lifetime. Many projects I know leave that up to the environment, i.e. require a service restart in an acme.sh deploy script. How about we follow Traefik's approach in a follow-up PR? The configuration surface would remain stable.

ItalyPaleAle:
@IngmarStein sorry, to be clear, my comment wasn't about LAN... just that whether it's LAN or not, we recommend having a reverse proxy regardless, like Traefik, Caddy, Nginx, etc

I would prefer waiting and doing this with hot reloading right away unles @kmendell thinks otherwise

IngmarStein:
@ItalyPaleAle oh, I got you about the reverse proxy! My remark about encryption was especially about the connection reverse proxy <-> pocket-id.

I'll add hot reloading then in one go.

kmendell:
I agree, there should be some sort of hot reloading as like @ItalyPaleAle said, most people would probably want to use lets encrypt or another ACME solution potentially.

IngmarStein:
Alright, added.

backend/internal/common/env_config.go
```diff
@@ -70,6 +71,10 @@ type EnvConfigSchema struct {
 	UnixSocketMode  string `env:"UNIX_SOCKET_MODE"`
 	LocalIPv6Ranges string `env:"LOCAL_IPV6_RANGES"`
 
+	TLSMinVersion string `env:"TLS_MIN_VERSION"`
```
ItalyPaleAle: Do we even need TLS 1.2 support? Among clients, TLS 1.3 is baseline, with broader support than WebAuthn. https://caniuse.com/?search=tls+1.3

Can we just ship 1.3 only and ignore 1.2? (And if someone really, absolutely needed 1.2, they can use a proxy)

backend/internal/common/env_config.go
```diff
@@ -211,6 +217,33 @@ func ValidateEnvConfig(config *EnvConfigSchema) error {
...
+	// Validate TLS config
+	switch {
+	case config.TLSCertFile != "" && config.TLSKeyFile == "":
+		return errors.New("TLS_KEY_FILE must be set when TLS_CERT_FILE is set")
+	case config.TLSCertFile == "" && config.TLSKeyFile != "":
+		return errors.New("TLS_CERT_FILE must be set when TLS_KEY_FILE is set")
+	}
+
+	if config.TLSCertFile != "" && config.TLSKeyFile != "" {
+		if _, err := os.Stat(config.TLSCertFile); err != nil {
```
ItalyPaleAle, line 226: Please use `utils.FileExists`

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -224,3 +289,87 @@ func initLogger(r *gin.Engine) {
...
+			return
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+			// Only reload on write/rename events
+			if event.Has(fsnotify.Write | fsnotify.Rename) {
+				slog.Info("TLS certificate file changed, reloading", slog.String("path", event.Name))
+
+				if err := p.reloadCertificate(); err != nil {
```
ItalyPaleAle: Important: batch these so you consider all events within, say, 1s, as the same event.

In many cases people won't just write the cert or the key, but both. We should make sure to load them atomically together, or there may be a brief time where the server is using a cert/key pair that don't match

ItalyPaleAle (changes requested):

ItalyPaleAle:
Pretty good, but please see my comments!

backend/internal/common/env_config.go
IngmarStein: That's quite a progressive stance, I like it! Done.

backend/internal/common/env_config.go
IngmarStein: I used the functionally equivalent `os.Stat` to avoid a circular import:

- common imports utils (for FileExists)
- utils already imports common (for EnvConfig)

Ok for you or should we split out `FileExists`?

backend/internal/bootstrap/router_bootstrap.go
IngmarStein: Good point. Added the debounce logic.

backend/internal/common/env_config.go
ItalyPaleAle: Oh I didn't realize the circular dependency. I'm ok leaving  os.Stat here then

ItalyPaleAle (approved):

IngmarStein:
@kmendell are the snyk checks actually failing? I don't have access to the results.

stonith404 (approved):

PR #1433, opened 2026-04-18
feat: add auth method claim (`amr`) to tokens

### LINKED ISSUE #861: 🚀 Feature: store auth method (passkey, e-mail, code) in claim
### Feature description

Optionally add the method of authentication (software passkey, hardware token, by e-mail or code) to the access token claims. This way the app can check for how (securely) the user logged in and optionally prevent access to most secured areas.

An easier, but less flexible way would be to add a Boolean to the oidc client "Require passkey login" or something, denying access to the app when the user does not login with a passkey.

### Pitch

Passkeys are great. But not all my users are comfortable with them yet. So many use e-mail auth as method. For most apps this is fine for us, but for some apps (or areas of apps) I would like to require the user to use a passkey. Even nicer would be to be able to check if a physical key has been used.

stonith404:
This PR adds the `amr` claim to the access tokens and ID token. If the user has authenticated with a login code its value is `["otp"]` and if authenticated with a passkey it's `["phr"]`.

Closes #861 and a security vulnerability which will get disclosed in the next release.

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -99,6 +99,7 @@ func (oc *OidcController) authorizeHandler(c *gin.Context) {
 		c.Request.Context(),
 		input,
 		c.GetString("userID"),
+		c.GetString("authenticationMethod"),
 		c.ClientIP(),
 		c.Request.UserAgent(),
 	)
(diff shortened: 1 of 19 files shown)


stonith404:
@coderabbitai review

ItalyPaleAle (approved):

kmendell (approved):
LGTM as well

PR #1445, opened 2026-04-21
fix: improve keyboard navigation and screen-reader labels

### LINKED ISSUE #1444: 🐛 Bug Report: Keyboard focus is lost when activating a link in the settings sidebar
### Reproduction steps

1. Open Pocket ID in a browser and sign in as an administrator.
2. Using only the keyboard, press Tab repeatedly to move focus through the page.
3. Continue pressing Tab until focus is on the "Administration" button in the settings sidebar and press Enter to expand it.
4. Continue pressing Tab until focus lands on a sub-item such as "OIDC Clients".
5. Press Enter to activate the link.

### Expected behavior

After navigating to the new page, keyboard focus should remain on the activated sidebar link so that pressing Tab continues from where the user was in the sidebar. The sidebar stays visible across navigations, so losing focus is disorienting for keyboard-only users.

### Actual Behavior

After pressing Enter, focus is reset to the document body as part of the client-side navigation. The sidebar is still visible but no longer focused, so the next Tab press starts from the top of the page (through the header logo, theme switcher and avatar) before the user can reach the sidebar again. This is effectively a WCAG 2.4.3 "Focus Order" regression for in-layout navigation.

While in there I also noticed a few related accessibility gaps that I would like to bundle into the same pull request if that is acceptable: the sidebar navigation and the header account-menu trigger have no accessible name, the email and code inputs on the alternative login pages are only labelled by their placeholder, and table column headers lack `scope` / `aria-sort`.

### Pocket ID Version

v2.6.0 (reproduced on current `main`)

### Database

Not relevant to this bug — it is purely client-side keyboard focus behaviour. Reproduced with the default SQLite build.

### OS and Environment

Reproduced on macOS with recent Chrome and Firefox. Not environment specific.

### Log Output

_Not applicable — client-side only._

bjoernch:
## Summary

Addresses a regression where activating a settings sidebar link with Enter resets keyboard focus to the document body, forcing keyboard-only users to tab through the entire header again before they can reach the sidebar. While working on that I also fixed a handful of related accessibility gaps I ran into in the same area.

Closes #1444.

## Changes

- **Preserve focus on sidebar navigation.** The settings sidebar lives in the `/settings` layout and stays in the DOM across client-side navigations, so resetting focus to `<body>` (SvelteKit's default) is unnecessary and disorienting. Adding `data-sveltekit-keepfocus` on the `<nav>` keeps focus on the activated link. SvelteKit's built-in live-region announcer still fires, so assistive technology is still told about the navigation.
- **Give landmarks accessible names.** The settings sidebar `<nav>` now carries `aria-label={m.settings()}` so multiple navigation landmarks are distinguishable. The header account dropdown trigger (which only contains the user's avatar image) now carries `aria-label={m.my_account()}`.
- **Label the alternative login inputs.** The email and code inputs on `/login/alternative/email` and `/login/alternative/code` were previously labelled only by a placeholder, which disappears on focus. They now also expose an `aria-label`. The placeholder is kept for visual consistency with the existing login design.
- **Mark column headers as columns.** Default `scope="col"` on the shared `<th>` component.
- **Announce the current sort direction.** Sortable columns in the advanced table now set `aria-sort` to `ascending`, `descending` or `none`. Non-sortable columns set no attribute at all.

## Tests

Added `tests/specs/navigation.spec.ts` with two Playwright specs:

- The settings sidebar is reachable as a navigation landmark named "Settings".
- Pressing Enter on a sidebar link keeps focus on that link after navigation.

## How to try it out

1. Sign in.
2. Using only the keyboard, Tab to any sidebar link (for example "Audit Log") and press Enter.
3. The page navigates and the next Tab press moves to the next sidebar entry rather than restarting from the header.

No visual changes.

### DIFF
--- a/frontend/src/lib/components/header/header-avatar.svelte
+++ b/frontend/src/lib/components/header/header-avatar.svelte
@@ -17,7 +17,7 @@
 </script>
 
 <DropdownMenu.Root>
-	<DropdownMenu.Trigger
+	<DropdownMenu.Trigger aria-label={m.my_account()}
 		><Avatar.Root class="size-9">
 			<Avatar.Image src={cachedProfilePicture.getUrl($userStore!.id)} />
 		</Avatar.Root></DropdownMenu.Trigger
--- a/frontend/src/lib/components/sidebar.svelte
+++ b/frontend/src/lib/components/sidebar.svelte
@@ -77,7 +77,11 @@
 	const delayUpdateLink = () => `${layout().total * ROW_STAGGER}ms`;
 </script>
 
-<nav class="text-muted-foreground grid gap-2 text-sm">
+<nav
+	class="text-muted-foreground grid gap-2 text-sm"
+	aria-label={m.settings()}
+	data-sveltekit-keepfocus
+>
 	{#each items as item, i}
 		{#if item.children?.length}
 			{@const id = groupId(item, i)}
--- a/frontend/src/lib/components/table/advanced-table.svelte
+++ b/frontend/src/lib/components/table/advanced-table.svelte
@@ -223,7 +223,16 @@
 						{/if}
 
 						{#each visibleColumns as column}
-							<Table.Head class={cn(column.sortable && 'p-0')}>
+							<Table.Head
+								class={cn(column.sortable && 'p-0')}
+								aria-sort={column.sortable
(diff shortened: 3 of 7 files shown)


kmendell (approved):
LGTM, Thanks!

PR #1449, opened 2026-04-22
feat: add emergency recovery codes

bjoernch:
## Summary

Adds emergency recovery codes so that users can sign in when they have lost access to every passkey on their account. Administrators can toggle the feature globally; disabling it purges every existing code in the same transaction as the configuration change.

### Account settings

- New **Recovery Codes** card with *Generate* / *Regenerate* / *Revoke* actions. Shows `X of Y codes remaining`.
- After generation, codes appear once in a dialog with:
  - copy-to-clipboard
  - plain-text download
  - minimalist single-page PDF "emergency kit" download

The PDF intentionally omits username, email, instance URL and generation date. A lost copy is not traceable back to an instance or account. It is produced by a small hand-rolled generator (`frontend/src/lib/utils/recovery-code-pdf.ts`) to avoid pulling a large PDF dependency for one printable page.

### Sign-in flow

- The alternative sign-in chooser gains a **Use a recovery code** entry next to the existing login-code / email options.
- When recovery codes are the only alternative configured, the login screen links straight to `/login/alternative/recovery-code` instead of the chooser.
- Recovery-code redemption is rate limited to **5 requests per 10 seconds** via the existing rate-limit middleware.
- Entered codes are normalised (dashes/spaces stripped, lowercased) and hashed with SHA-256 before comparison, so codes work regardless of how the user typed them.
- The matching row is selected with `SELECT ... FOR UPDATE` inside a transaction so a code cannot be redeemed twice under a race.

### Application configuration

- New **Enable recovery codes** switch in the General card. When turned off, every `recovery_codes` row is deleted in the same transaction as the config update and the recovery-code service refuses every subsequent generate/status/revoke/redeem call.

### Data model

- New `recovery_codes` table (`id`, `user_id`, `code_hash`, `used_at`, `created_at`) with a `user_id` index. Migrations provided for both SQLite and Postgres.
- New audit-log events: `RECOVERY_CODES_GENERATED`, `RECOVERY_CODES_REVOKED`, `RECOVERY_CODE_SIGN_IN`.
- New JWT authentication method `AuthenticationMethodRecoveryCode` so downstream policies can distinguish a recovery-code login from a passkey login if they want to.

### Translation

All new user-visible strings live in `frontend/messages/en.json` and are consumed via Paraglide `m.*()`. Crowdin will pick them up via the existing `crowdin.yml`. The PDF receives every string as a parameter — there is no English embedded in the generator itself.

## Test plan

Backend

- [x] `go test -tags exclude_frontend ./...` — all 14 packages green
- [x] `TestRecoveryCodeService_*` covers: batch generate, regenerate invalidates previous batch, single-use redeem, separator-insensitive redeem, case-insensitive redeem, redeem rejected when feature is disabled, garbage code, revoke-all, revoke rejected when disabled, status rejected when disabled
- [x] `prettier --check` passes on every touched frontend file

End-to-end (Playwright, `tests/specs/recovery-codes.spec.ts`, sqlite + filesystem locally; CI runs the full matrix)

- [x] Generate a batch from account settings, sign in with a code
- [x] Reusing the same code is rejected
- [x] Revoke from account settings empties the card
- [x] Code format matches `xxxx-xxxx-xxxx-xxxx`

Manual

- [x] Generated PDF contains no username, email, instance URL or date
- [x] Generated PDF has no `/Info` dictionary (no Author/Creator/Producer metadata)
- [x] Admin-disabling the feature wipes existing codes and hides all UI
- [x] Passkey sign-in still works; email one-time access still works

## Notes for reviewers

- The recovery code batch size (10) and code length (16 unambiguous alphanumeric chars, ~94 bits of entropy) are defined in `backend/internal/service/recovery_code_service.go`.
- The PDF generator uses only the three built-in PDF fonts (Helvetica, Helvetica-Bold, Courier-Bold) so nothing needs to be embedded.
- `+page.ts` falls back to `{ total: 0, unused: 0 }` if the status endpoint returns an error, so the account page keeps loading cleanly when an admin has just disabled the feature.


bjoernch:
Thanks for the CI run. I just pushed `33fa32f` which addresses the three findings:

- **gosec G602** (`app_config_service.go:282`) — the recovery-codes change accidentally carried a small refactor of the `UpdateAppConfigValues` key/value loop that wasn't needed. Reverted to the original `i := 1; i-1 / i` form so gosec can see the bound.
- **testifylint require-error** (`recovery_code_service_test.go:98`) — switched `require.Error` + `assert.ErrorAs` to a single `require.ErrorAs`, since a wrong error type should abort the subtest.
- **svelte-check** (`recovery-code-pdf.ts:237`) — narrowed `encodePdf`'s return to `Uint8Array<ArrayBuffer>` so it satisfies `BlobPart` under TS 5.7's widened `ArrayBufferLike` typings.

Locally `svelte-check` is green and the recovery-code service tests still pass. Could you kick CI again when you have a moment? Thanks!

kmendell:
Im going to leave this up to @stonith404

ItalyPaleAle:
I would say this feature requires some consideration.

In the typical scenario, recovery codes are used to re-gain access to an account when the second factor is lost (like TOTP). In this case, there's no second factor, and the recovery codes are essentially just as good as a password. The security posture is not the same: here we have 10 long-lived secrets that we'd need to trust users to manage properly (and they will most likely not), and it's a bit the antithesis of passkeys.

I would lean towards no, personally, the risk seems too high.

Pocket ID already has [account recovery procedures](https://pocket-id.org/docs/troubleshooting/account-recovery) which do not involve long-lived secrets.

backend/internal/service/recovery_code_service.go
```diff
@@ -0,0 +1,245 @@
...
+		Error
+	if err != nil {
+		return nil, err
+	}
+
+	rawCodes := make([]string, 0, recoveryCodeBatchSize)
+	records := make([]model.RecoveryCode, 0, recoveryCodeBatchSize)
+
+	for i := 0; i < recoveryCodeBatchSize; i++ {
+		raw, err := utils.GenerateRandomUnambiguousString(recoveryCodeRawLength)
```
ItalyPaleAle, line 66: This would need to be significantly longer, with 16 characters and the limited alphabet it's not a lot of entropy, less than 70 bits...

backend/internal/controller/recovery_code_controller.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// redeemHandler godoc
+// @Summary Redeem recovery code
+// @Description Exchange an emergency recovery code for a session.
+// @Tags Recovery Codes
+// @Param code body redeemRecoveryCodeInput true "Recovery code"
+// @Success 200 {object} dto.UserDto
+// @Router /api/recovery-code [post]
+func (c *RecoveryCodeController) redeemHandler(ctx *gin.Context) {
```
ItalyPaleAle, line 103: This method should require at least the user's email in addition to the code.

Also, we'd probably want it strongly rate-limited by user name, not just by IP.

ItalyPaleAle (changes requested):

stonith404:
I agree with @ItalyPaleAle. Recovery codes are like more secure passwords in the end. Passkeys make Pocket ID users phishing resistant, if we introduce this feature users can be phished with the recovery codes. Additionally I think the chance for average users is pretty low that the recovery codes are stored in a safe place.


I think the current login methods are sufficient for account recovery. We even warn users if they only have configured one passkey. If Pocket ID runs in an organization the user can just contact the administrator and the administrator can create a one-time login code. Alternatively the admin can enable the option to sign in with an email code. The latter is not really recommended but still more secure than recovery codes.

Because of that I'm closing this PR, but thanks for your contribution. In the future I recommend you to create a feature request first.

PR #1454, opened 2026-04-24
fix: add missing /api prefix to app config swagger routes

aclerici38:
It looks like these 2 routes were missing `/api`, I came across this when trying to use an auto-generated client from the swagger spec which is based on the workflow the website uses https://github.com/aclerici38/pocket-id-go-client/blob/96b20c6da9d24227d89dcfe950fd6188da816f04/Makefile#L20-L33

thanks!


ItalyPaleAle (approved):

stonith404 (approved):
Thanks!

PR #1458, opened 2026-04-29
fix(oidc): delete refresh tokens on end-session to prevent reuse after logout

### LINKED ISSUE #1390: 🐛 Bug Report: Refresh tokens not deleted from database on end-session
### Reproduction steps

1. Create an application with Public Client enabled and PKCE enabled. Also give it a Logout Callback URL
2. Have a user log in (calls to oidc/authorize and oidc/token)
3. The client now has a refresh token
4. Have the user end the session (call to end-session)

### Expected behavior

All refresh tokens for that user_id should be deleted from the "oidc_refresh_tokens" table. It should no longer be possible to use the refresh token.

### Actual Behavior

The refresh token is still in the "oidc_refresh_tokens" table. It can still be used.

Looking at the source code, in oidc_service.go, in createTokenFromRefreshToken, the used refresh token is being deleted from the database. But, in  oidc_controller.go, in EndSessionHandler, no refresh tokens seem to be deleted.

### Pocket ID Version

v2.4.0

### Database

SQLite

### OS and Environment

Docker, Windows 11 Pro 25H2

### Log Output

_No response_

wucm667:
Fixes #1390

When a user calls the OIDC end-session endpoint, refresh tokens stored in the `oidc_refresh_tokens` table were not deleted. This allowed tokens to be reused after the user had logged out, which is a security issue.

## Root Cause

The `ValidateEndSession` method in `oidc_service.go` verifies the ID token and returns the logout callback URL, but never deletes any existing refresh tokens for the user. The `EndSessionHandler` in `oidc_controller.go` only clears the access token cookie.

## Fix

Add a single database query in `ValidateEndSession` to delete all refresh tokens for the authenticated user before returning the callback URL:

```go
s.db.WithContext(ctx).Where("user_id = ?", userID).Delete(&model.OidcRefreshToken{})
```

This deletes ALL refresh tokens for this user across ALL clients, which is the correct behavior for a session-wide logout.

## Changes

- `backend/internal/service/oidc_service.go`: Add refresh token deletion in `ValidateEndSession`
- `backend/internal/service/oidc_service_test.go`: Add `TestOidcService_ValidateEndSession_DeletesRefreshTokens` unit test


backend/internal/service/oidc_service.go
```diff
@@ -1229,6 +1229,11 @@ func (s *OidcService) ValidateEndSession(ctx context.Context, input dto.OidcLogo
 		return "", &common.OidcMissingAuthorizationError{}
 	}
 
+	// Delete all refresh tokens for this user
```
ItalyPaleAle, line 1235: This method now has multiple queries, please wrap them in a transaction for consistency

ItalyPaleAle (changes requested):

backend/internal/service/oidc_service.go
wucm667: Done! Wrapped both queries in a transaction using tx := s.db.Begin() with defer tx.Rollback(), matching the pattern used in other methods like Authorize.

ItalyPaleAle (approved):

kmendell (approved):
LGTM, Thanks!

stonith404:
@kmendell @ItalyPaleAle Tbh I'm not sure if this change makes sense. With this change all refresh tokens get deleted. So if a user signs out on one device, the refresh tokens of every other devices will get invalidated as well, which is a behavior the user probably doesn't expects or wants.

My suggestion would be to link the refresh token to the ID token. With the `id_token_hint` we can know which refresh token we have to delete. The `id_token_hint` is optional though, if it's not provided we would just keep the refresh token. 

Whats your take on this?

ItalyPaleAle:
@stonith404 you're right, I had not considered that. We could use the jti claim? I'd get it from the access token instead of id token tho maybe

stonith404:
@ItalyPaleAle With a [RP initiated logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) we unfortunately don't have access to the access token. 

In my opinion the only way is to store the JTI of the ID token in the refresh token table. I don't see another way how we could make the connection between a refresh token and session of the user.

kmendell:
@stonith404 yeah i as well didnt think of that...  your suggestion makes sense to me, nothing off the top of my head to work around that either without doing research , but seems like the simpliest route forward

PR #1462, opened 2026-05-02
fix: invalidate cache when changing image

GameTec-live:
Today I encountered an issue when attempting to change the Logo of a OIDC client. It saved, but didnt "persist".

This change should (and in my testing did) correctly invalidate the cache when changing a clients logo so that the new image is visible immediately and doesnt change back to the old one randomly.

This seems to specifically happen when when one uses URLs, so for example pastes in the URL from dashboardicons or similar.


ItalyPaleAle (approved):

kmendell (approved):
Thanks!

PR #1466, opened 2026-05-04
fix(oidc): fall back to Basic auth when PKCE puts client_id in body

### LINKED ISSUE #1465: 🐛 Bug Report: Basic auth fallback skipped when PKCE puts client_id in body
### Reproduction steps

When an OIDC client uses `client_secret_basic` together with PKCE, some libraries (e.g. [jumbojett/openid-connect-php](https://github.com/jumbojett/OpenID-Connect-PHP)) send `client_id` and `code_verifier` in the POST body while keeping `client_secret` only in the `Authorization: Basic` header. This is permitted by [RFC 6749 §2.3.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1) (only "MUST NOT use more than one authentication method in each request").

1. Configure an OIDC client in Pocket ID with PKCE enabled and a client secret.
2. Use a relying party that combines `client_secret_basic` with PKCE and puts `client_id` in the body alongside `code_verifier` (e.g. jumbojett/openid-connect-php).
3. Complete the authorization code flow and let the RP exchange the code at `/api/oidc/token`.

### Expected behavior

Pocket ID reads `client_secret` from the `Authorization` header, accepts the request, and returns tokens. The same fallback already works for the case where both `client_id` and `client_secret` are absent from the body.

### Actual Behavior

Token exchange fails with `400 Bad Request` / "Client id or secret not provided" because the Basic auth fallback in `oidc_controller.go` only triggers when **both** `client_id` and `client_secret` are empty in the body:

https://github.com/pocket-id/pocket-id/blob/v2.6.2/backend/internal/controller/oidc_controller.go#L205-L208

```go
// Client id and secret can also be passed over the Authorization header
if input.ClientID == "" && input.ClientSecret == "" {
    input.ClientID, input.ClientSecret, _ = utils.OAuthClientBasicAuth(c.Request)
}
```

The same pattern exists at the device authorization endpoint (L707-L709). When the body contains `client_id` (which RFC 6749 explicitly permits and PKCE libraries commonly do), the header is never read and the secret stays empty.

This is distinct from #1315: `client_secret_basic` does work, but only when the body omits `client_id`.

### Pocket ID Version

v2.6.2 (also reproduced on current `main`)

### Database

SQLite

### OS and Environment

Docker on Unraid, Traefik in front. Reproduced with jumbojett/openid-connect-php as the RP.

### Log Output

```
WRN Request with errors: Error #01: Client id or secret not provided
```

### Proposed fix

Fall back to Basic auth whenever `client_secret` is empty, and only override `client_id` from the header if the body didn't supply one. Happy to send a PR; I have the change running locally.

mgabor3141:
Fixes #1465.

[RFC 6749 §2.3.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1) lets a client send `client_secret_basic` credentials in the `Authorization` header while still putting other parameters (including `client_id`) in the request body. Some OIDC libraries do exactly this when PKCE is in play: e.g. [jumbojett/openid-connect-php](https://github.com/jumbojett/OpenID-Connect-PHP) sends `client_id` and `code_verifier` in the body and keeps `client_secret` in the header.

The current Basic auth fallback in `oidc_controller.go` only triggers when **both** body fields are empty:

```go
if input.ClientID == "" && input.ClientSecret == "" {
    input.ClientID, input.ClientSecret, _ = utils.OAuthClientBasicAuth(c.Request)
}
```

So the moment a body `client_id` is present, the header is never read, the secret stays empty, and token exchange fails with "Client id or secret not provided".

This PR falls back to Basic auth whenever the secret is missing, and only overrides `client_id` from the header when the body didn't supply one (avoiding any accidental cross-field override). Same fix applied to the device authorization endpoint.

### Changes

- `backend/internal/controller/oidc_controller.go`: relax the Basic auth fallback condition in `createTokensHandler` and `deviceAuthorizationHandler`.
- `backend/internal/controller/oidc_controller_test.go`: add a sub-test covering the PKCE shape (body has `client_id`, header has secret). The existing "Uses Basic Auth Credentials When Body Credentials Missing" case still passes unchanged.

### Verified with

jumbojett/openid-connect-php as the relying party against a patched Pocket ID build, running for ~2 months without regressions on the standard `client_secret_post` and header-only flows.


ItalyPaleAle (approved):

kmendell (approved):
LGTM, Thank you!

PR #1478, opened 2026-05-18
chore: use goreleaser for binaries and static image builds

kmendell:
https://goreleaser.com/customization/general/

TODO:

- [x] Add discord release webhook
- [x] Binary signing?


kmendell:
@stonith404 a decision we could make is to drop the full dockerfile build (normal dockerfile) and juts make everything use goreleaser binairies, so everything would be a static binary built from that. So our normal dockerfile would look more like the Dokcerfile-prebuilt

ItalyPaleAle:
I think this is good but we should keep codesigning pleae!

kmendell:
Yeah I just wasn't sure if we should use cosign or something instead , unless that not what you mean

ItalyPaleAle:
I'd stick to the built in one in Actions. That uses cosign under the hood and publishes attestations automatically.

kmendell:
@ItalyPaleAle done, i also added my apple developer stuff for noterizing macos binaries

.github/workflows/release.yml
```diff
@@ -66,93 +74,48 @@ jobs:
...
+          args: release --clean --skip=validate --parallelism=4
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: gh release upload ${{ github.ref_name }} backend/.bin/*
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+          MACOS_SIGN_P12: ${{ secrets.MACOS_SIGN_P12 }}
+          MACOS_SIGN_PASSWORD: ${{ secrets.MACOS_SIGN_PASSWORD }}
+          MACOS_NOTARY_KEY: ${{ secrets.MACOS_NOTARY_KEY }}
+          MACOS_NOTARY_KEY_ID: ${{ secrets.MACOS_NOTARY_KEY_ID }}
+          MACOS_NOTARY_ISSUER_ID: ${{ secrets.MACOS_NOTARY_ISSUER_ID }}
```
ItalyPaleAle, line 68: Wait what are these?

kmendell: Apple Developer Certificates so the macOS binaries are properly notarized and signed on macOS, (i pay for an apple developer account)

ItalyPaleAle: Ah, I didn't know you were paying for it!

.goreleaser.yaml
```diff
@@ -0,0 +1,220 @@
...
+  use_existing_draft: true
+  replace_existing_artifacts: true
+  prerelease: auto
+  make_latest: true
+  mode: keep-existing
+  name_template: "{{ .Tag }}"
+  footer: |
+    ---
+
+    Released by [GoReleaser](https://github.com/goreleaser/goreleaser).
```
stonith404:
```suggestion
```
I would remove this footer.

.github/workflows/release.yml
```diff
@@ -66,93 +74,48 @@ jobs:
...
-      - name: Install frontend dependencies
-        run: pnpm --filter pocket-id-frontend install --frozen-lockfile
-
-      - name: Build frontend
-        run: pnpm --filter pocket-id-frontend build
-
-      - name: Build binaries
-        run: sh scripts/development/build-binaries.sh
-
       - name: Build and push container image
```
stonith404: Don't we already build the image with Goreleaser or am I missing something?

.github/workflows/release.yml
```diff
@@ -66,93 +74,48 @@ jobs:
...
-  publish-release:
-    runs-on: depot-ubuntu-latest
-    needs: [build]
-    env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v6
-      - name: Mark release as published
+      - name: Binary attestation
```
stonith404, line 72: Currently we only attest the binaries, we have to do it for the Docker images as well.

stonith404 (changes requested):
Sorry for the late review. Please see my comments. 

Have you tested the release workflow in a private repository already? I think we would have to test this in a private repository first because it's very rare that actions work as expected on first try 😅

.github/workflows/release.yml
kmendell: The goreleaser images only use static binaries, so if we simplify down to only the static and distroless dockerfiles, we could only use go releaser yeah, otherwise wed have to keeo that part.

kmendell: This should be fixed now.

kmendell:
All review notes should be addressed now, and the test run is here: https://github.com/kmendell/pocket-id/actions/runs/27875990116/job/82496652818

The only difference i did on the fork is not use the depot runners since i cant install them into a personal github account. Otherwise its identical. 

Discord message also was sent 
image

stonith404 (approved):
LGTM, thanks!

PR #1479, opened 2026-05-19
feat: support systemd socket activation

deviant:
This adds support for [systemd socket activation](https://0pointer.de/blog/projects/socket-activation.html). In this mode, Pocket ID is not responsible for creating its listening socket; instead receiving one from systemd, fully prepared and ready to use from the moment it starts up. This confers various benefits, such as simpler and more robust network bringup code, and the ability to launch Pocket ID on demand (upon receiving a connection). However, what I am personally most interested in is the flexibility it allows for customizing the socket itself, and how this customization can make sandboxing more ergonomic.

As a concrete example: I am running Pocket ID in a strongly locked-down configuration, making use of many of [systemd's hardening options](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Sandboxing) to do so. One of these options, `PrivateUsers=`, abstracts over Linux *user namespaces* — a widely-used sandboxing technology, which is used by (and indeed fundamental to) container security. User namespaces isolate processes from the user table of the host system, making it more difficult for a compromised program to cause damage to the rest of the system.

Additionally, I am reverse-proxying Pocket ID behind Caddy.

I would like to expose Pocket ID such that its socket is only accessible to Caddy, and not to any other (unprivileged) processes. One simple way I can achieve this is by using a UNIX socket: by setting its *group* to `caddy`, and its *mode* to `0660` (i.e., user & group have read/write permissions; everyone else have no privileges). Whilst Pocket ID has no built-in support for setting the group of its socket, this can be accomplished in other ways, such as by adding a *default ACL* of `u:caddy:rw` to the containing directory. However, inside of a user namespace, this is no longer possible: from Pocket ID's perspective, the `caddy` user/group do not exist!

Along to the rescue come systemd socket units. I can define a simple configuration in **pocket-id.socket**:

```ini
[Unit]
Description=Pocket ID Socket

[Socket]
ListenStream=/run/pocket-id/socket
SocketGroup=caddy
SocketMode=0660

[Install]
WantedBy=sockets.target
```

And now I can run Pocket ID with full sandboxing, without any messy hacks or workarounds necessary. I'm sure that others would also like to be able to do this, as this feature has been [requested previously](https://github.com/pocket-id/pocket-id/pull/615#issuecomment-2943904318). It's quite useful!

Personally, I want Pocket ID to be running all the time, and not only after systemd detects something trying to connect to its socket, so I also have this in my **pocket-id.service**:

```ini
[Unit]
Requires=pocket-id.socket
After=pocket-id.socket
```

And now the socket is created whenever I start Pocket ID.

You might note that I chose to add a new configuration variable, `SYSTEMD_SOCKET`, which enables this feature. The reason I chose to do this instead of simply autodetecting when socket activation is being used (which is possible, through testing for the presence of the `LISTEN_FDS` and `LISTEN_PID` environment variables) is that in the event of misconfiguration, I have to imagine that most people wouldn't want Pocket ID to quietly start listening on a TCP socket instead. I certainly wouldn't! With this flag, it will loudly complain if something is amiss, instead of "failing open".

Lastly, I cleaned up the listener creation code a bit. It seemed to me that the logic around that had been abstracted wrongly, with too much effort placed on deduplicating certain code paths, and had turned into somewhat of a mess as a result. I reorganized it in the way that makes the most sense to me, and I think it's a lot simpler and easier to read now. I hope this is okay to roll into the same pull request!

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/aws/smithy-go v1.25.0
 	github.com/caarlos0/env/v11 v11.4.0
 	github.com/cenkalti/backoff/v5 v5.0.3
+	github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf
 	github.com/disintegration/imageorient v0.0.0-20180920195336-8147d86e83ec
 	github.com/disintegration/imaging v1.6.2
 	github.com/dunglas/go-urlpattern v0.0.0-20241020164140-716dfa1c80b1
--- a/backend/go.sum
+++ b/backend/go.sum
@@ -65,6 +65,8 @@ github.com/containerd/errdefs v1.0.0 h1:tg5yIfIlQIrxYtu9ajqY42W3lpS19XqdxRQeEwYG
 github.com/containerd/errdefs v1.0.0/go.mod h1:+YBYIdtsnF4Iw6nWZhJcqGSg/dwvV7tyJ/kCkyJ2k+M=
 github.com/containerd/errdefs/pkg v0.3.0 h1:9IKJ06FvyNlexW690DXuQNx2KA2cUJXx151Xdx3ZPPE=
 github.com/containerd/errdefs/pkg v0.3.0/go.mod h1:NJw6s9HwNuRhnjJhM7pylWwMyAkmCQvQ4GpJHEqRLVk=
+github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf h1:iW4rZ826su+pqaw19uhpSCzhj44qo35pNgKFGqzDKkU=
+github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/cpuguy83/go-md2man/v2 v2.0.6/go.mod h1:oOW0eioCTA6cOiMLiUPZOpcVxMig6NIQQ7OS05n1F4g=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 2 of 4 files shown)


ItalyPaleAle:
> and the ability to launch Pocket ID on demand (upon receiving a connection).

This is not something we'd want, to be honest.

Pocket ID does have a set of background tasks that are meant to run on a schedule, and if the process is not active, they would not run either. Also, Pocket ID is not optimized for quick cold starts, as it performs a variety of actions (incl. database migrations) upon startup.

deviant:
> > and the ability to launch Pocket ID on demand (upon receiving a connection).
> 
> This is not something we'd want, to be honest.
> 
> Pocket ID does have a set of background tasks that are meant to run on a schedule, and if the process is not active, they would not run either. Also, Pocket ID is not optimized for quick cold starts, as it performs a variety of actions (incl. database migrations) upon startup.

Yeah, that’s why I only have the socket configured as a dependency of the service, and enable the service, not the socket. :-) Systemd socket activation is a bit of a misnomer because while it does allow for services to be activated on socket connection, that doesn’t have to be how you use it, as there are other useful things it allows for.

I think this is a case of one of these things where the technology was originally created to solve one problem (parallelizing service startup), but gained additional functionality and purpose along the way as features were added to the surrounding software. If I understand correctly, systemd began as a “launchd for Linux”, where the primary purpose was to speed up the boot process, but now it has a set of robust sandboxing options as well (plus numerous other things… 😅) The name has stuck, but now the feature solves additional problems that are outside of its original scope.

> > Personally, I want Pocket ID to be running all the time,

I realize I did not explain why when I said this, but it was because of the reasons you gave above. In addition, I would really like to be sure that my SSO is always online and available, so I don’t run into any unpleasant surprises. Sorry about not being more clear about that!

deviant:
As an addendum to this: I think it would make most sense to explain this in the documentation as "Whether to manage the listening socket using systemd. [...]" instead of "Whether to use systemd socket activation" or something equally less specific. This way, the language encourages users to use the feature correctly, instead of doing something they probably shouldn't. I think this is a great point, thanks for bringing it up. :-)

stonith404 (approved):
Looks good to me, thanks.

PR #1480, opened 2026-05-19
feat: delete OAuth refresh token on RP initiated logout

### LINKED ISSUE #1390: 🐛 Bug Report: Refresh tokens not deleted from database on end-session

stonith404:
PR #1458 has been reverted because it deletes all refresh tokens of an user if the user signs out. With this PR only the refresh token connected with the ID token provided with the `id_token_hint` gets deleted. If no `id_token_hint` is provided by the client the refresh token can't be deleted.

Fixes https://github.com/pocket-id/pocket-id/issues/1390

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -311,6 +311,12 @@ func (oc *OidcController) EndSessionHandler(c *gin.Context) {
 	// The validation was successful, so we can log out and redirect the user to the callback URL without confirmation
 	cookie.AddAccessTokenCookie(c, 0, "")
 
+	// Callback URL can be empty if none is configured
+	if callbackURL == "" {
+		c.Redirect(http.StatusFound, common.EnvConfig.AppURL+"/logout")
+		return
+	}
+
 	logoutCallbackURL, _ := url.Parse(callbackURL)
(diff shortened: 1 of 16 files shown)


kmendell (approved):
LGTM, apologies for the old PR i didnr realize the impact

stonith404:
No worries

Issue #564, opened 2025-05-24
🚀 Feature: Custom scopes


zumoshi:
### Feature description

Currently pocket-id supports custom claims which is great. I've already used the custom claims to great success both in my own custom apps and in 3rd party applications.

However, all the claims are under the profile scope. I would like to be able to define custom scopes which show separately (same way Email/Profile do atm) while signing in to verify what info an app is asking for and isolate the claims to apps that need them.

![Image](https://github.com/user-attachments/assets/40089fb4-d18f-4dc1-b73e-8d0a9516c87e)

### Pitch

I have two main reasons for wanting this.

1. Reducing the size of resulting jwt, since most apps use the id_token as a cookie or header which is repeated on every request. if I use 10 apps and they each need a claim themselves they would need to resend data of the other 9 apps which they don't need on every request.
2. Security concerns. from what I understand the JWT of id_token is signed with a global key, not a per oidc-client key. so an id_token generated while logging into one is valid for another. If one app had a security vulnerability (e.g. storing id_token in local storage as opposed to HTTP Only cookie and XSS), the leaked key could potentially be used on all other apps protected by pocket until it expires.

My specific case:
App 1 is a panel I made myself for management of a multi-user Caddy reverse proxy setup, and a claim has the filename for the user's caddyfile.
App 2 is a K3S instance, and uses a claim for RBAC to determine which role to assign to the logged in user.

I have a lot less trust in my own code than that of kubernetes, so I would prefer that the claim needed for K3S be missing from the JWT generated for my app. This could be achieved if I could add two scopes separate from profile, lets say scope caddy an scope kuber. then have each app request the respective one and verify while clicking sign in that I'm allowing the correct one.

ItalyPaleAle:
Not the thing you asked for, but...

id_tokens are not meant to be used to maintain a session: https://oauth.net/id-tokens-vs-access-tokens/

> From what I understand the JWT of id_token is signed with a global key, not a per oidc-client key. so an id_token generated while logging into one is valid for another.

That's what the `aud` claim should be for. In Pocket ID (per OAuth specs), access tokens and ID tokens have the client ID of the app as value for `aud` https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3

zumoshi:
Interesting read. Thanks for sharing, I'll keep that in mind when implementing OIDC on my own code.

[...]

PR #1486, opened 2026-05-23
feat!: replace custom claims with custom fields

stonith404:
This is a preparation PR for implementing custom scopes #564. 

### Changes
This PR introduces "Custom Fields" which replace "Custom Claims".  Custom Fields allow admins to extend the data schema of users and groups. It's basically a more powerful version of Custom Claims.

Custom fields can be added in the application configuration:
Screenshot 2026-05-23 at 16 12 31@2x

The forms get naturally extended, like it's a native Pocket ID field:
Screenshot 2026-05-23 at 16 14 43@2x

With this change I also saw the opportunity to implement custom fields from LDAP (#742). If a custom field was added, Pocket ID automatically tries to get the custom field by its key from the LDAP server.

### Custom scopes (#564)
The next step would be to implement custom scopes. The goal is to allow an admin to create custom scopes per OIDC client. To each scope the user can assign custom claims. There will be three options to define the claims:
1. Select native or custom fields from Pocket ID
2. Use the expression language [CEL](https://cel.dev) to aggregate data from different native or custom fields
3. HTTP hook where Pocket ID sends an authenticated GET (or maybe POST) request to an endpoint which returns the claims.


backend/internal/dto/custom_field_dto.go
```diff
@@ -0,0 +1,42 @@
...
+
+type CustomFieldValueDto struct {
+	CustomFieldID string `json:"customFieldId"`
+	Key           string `json:"key,omitempty"`
+	Value         string `json:"value"`
+}
+
+type CustomFieldValueCreateDto struct {
+	CustomFieldID string `json:"customFieldId" binding:"required_without=Key" unorm:"nfc"`
+	Key           string `json:"key,omitempty" unorm:"nfc"`
```
ItalyPaleAle, line 11: For ID and key, should we restrict to a subset of ASCII? In which case, unorm wouldn't matter

backend/internal/service/custom_field_service.go
```diff
@@ -0,0 +1,566 @@
...
+	switch key {
+	case "given_name",
+		"family_name",
+		"name",
+		"email",
+		"email_verified",
+		"preferred_username",
+		"display_name",
+		"groups",
+		TokenTypeClaim,
```
ItalyPaleAle, line 50: Nit: this is the only constant and it looks odd

ItalyPaleAle (commented):
Quick review for now, I need to spend more time studying this as it's a larger change

kmendell:
I think this overall looks good i like the approach to it, i may have missed it, but this is a breaking change, but do users exisisting custom claims get migrated somehow? or required for them to be recreated?

Issue #1376, opened 2026-03-09
🚀 Feature: Passkey Autofill (Conditional UI)


elliott-levin:
### Feature description

Today, users need at least two clicks to authenticate. First, they click the "Authenticate" button on the Pocket ID page. Then, the browser takes over and presents a list of passkeys.

The "Conditional UI" webauthn feature allows users to skip the first step. The login page advertises that it accepts passkeys, and the user skips directly to selecting a passkey.

### Pitch

I think this would significantly reduce the friction of logging in, especially for less-technical users who are still learning about passkeys. Some other websites utilize this feature and I personally think it's the best workflow.

It would be a good idea to allow this feature to be enabled/disabled via configuration, since it might not be for everyone.

stonith404:
@fabianwimberger tried to implement this feature but in my opinion it doesn't really bring an advantage. Implementing the conditional UI requires to display an input field which doesn't make sense if we don't support sign in with email and password. 

In my opinion the current flow is already very fast, you literally have to do one click to show the passkey list, on keyboard devices you can even just click the "Enter" key. Because of that I'm closing this issue.

PR #1491, opened 2026-05-27
feat: add passkey autofill login

fabianwimberger:
This PR adds WebAuthn conditional UI support to the login page so browsers and password managers can offer saved passkeys automatically when the page loads. Users can still use the existing Authenticate button when autofill is unavailable or dismissed.

The implementation only starts conditional mediation when the browser reports passkey autofill support, and it completes authentication through the existing login flow.

Refs #1376

### DIFF
--- a/frontend/src/routes/login/+page.svelte
+++ b/frontend/src/routes/login/+page.svelte
@@ -2,12 +2,18 @@
 	import { goto } from '$app/navigation';
 	import SignInWrapper from '$lib/components/login-wrapper.svelte';
 	import { Button } from '$lib/components/ui/button';
+	import { Input } from '$lib/components/ui/input';
 	import { m } from '$lib/paraglide/messages';
 	import WebAuthnService from '$lib/services/webauthn-service';
 	import appConfigStore from '$lib/stores/application-configuration-store';
 	import userStore from '$lib/stores/user-store';
 	import { getWebauthnErrorMessage } from '$lib/utils/error-util';
-	import { startAuthentication } from '@simplewebauthn/browser';
+	import {
+		browserSupportsWebAuthnAutofill,
+		startAuthentication,
+		type AuthenticationResponseJSON
+	} from '@simplewebauthn/browser';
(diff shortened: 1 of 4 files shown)


fabianwimberger:
I pushed a fix for the E2E failures.

The auth helper was still assuming that passkey login only starts after clicking `Authenticate`. With conditional autofill enabled, the virtual passkey used in Playwright can complete the pending autofill request immediately, so the helper reached `/settings` before trying to click the button.

The helper now handles both paths: use the autofill redirect if it already happened, otherwise click `Authenticate`.

Should be green now once rerun.

stonith404:
Tbh I don't really see an advantage of this implementation. It doesn't make the sign in process faster but just makes it more confusing. Clicking on the input field just suggests the passkeys inline instead of a popup when clicking the sign in button. In the end the same passkeys are suggested and the exact same amount of steps are required.

Can you elaborate why you think this has an advantage?

Adobe Express - Screenshot 2026-05-29 at 11 48 27

stonith404:
Hm okay the behavior is different depending on the browser. In Safari it would save the user one click. I have to look into it a bit more, but I'm not really a fan of displaying an input field on the login page, because we don't offer email/password login.

Screenshot 2026-05-29 at 11 53 42

kmendell:
> Hm okay the behavior is different depending on the browser. In Safari it would save the user one click. I have to look into it a bit more, but I'm not really a fan of displaying an input field on the login page, because we don't offer email/password login.
> 
> Screenshot 2026-05-29 at 11 53 42

I agree, this creates a confusing UX for users with little to no benefit in my opinion

fabianwimberger:
Thanks for the feedback. I agree with you. I’ll close this PR for now. We can revisit this later once passkeys can be prompted without requiring a login field.

Before:
before-login


After (current PR):
after-login


PS: I also tried a hidden autofill anchor locally, but my password manager did not offer the passkey until pressing `Authenticate`, so that does not seem like a viable alternative either.

PR #1494, opened 2026-05-29
feat: add structured logging and audit trail for security violation attempts 

crazyrokr:
## Objective
  This implementation introduces a centralized security event logging mechanism via slog middleware to capture critical authentication failures (e.g., `401 Unauthorized`, `403 Forbidden`) with context-rich metadata - including client IP, user agent, and request path - facilitating integration with external security observability tools like **CrowdSec** or **Fail2Ban**. Furthermore, the audit trail has been expanded with persistent `SIGN_IN_FAILED` events, ensuring that failed authentication attempts, including those within failed transactions, are reliably recorded in the database, thereby providing administrators with crucial insights for threat mitigation while maintaining consistency through targeted transactional management. The error handler was also extended to natively intercept **WebAuthn** protocol errors and map them to appropriate HTTP status codes with security logging.
##  Assumptions
   - Brute-force detection tools (e.g., **Fail2Ban**, **CrowdSec**) primarily monitor structured application logs (stdout/stderr).
   - Centralizing security logging in the global error handler ensures consistent coverage across all authentication methods.
   - `401 Unauthorized` and `403 Forbidden` status codes are the primary indicators of security-relevant failures in this application.
   - **WebAuthn** protocol errors (from _go-webauthn/webauthn/protocol_) should be intercepted at the middleware layer and mapped to semantically correct HTTP status codes rather than surfacing raw internal error details to clients.
   - Audit log entries for failed sign-ins must survive transaction rollbacks and are therefore written using a direct database handle, not the caller's transaction.
##  Changes Made
1. Added `AuditLogEventSignInFailed` event type and `CreateSignInFailure` service method for persistent failure logging. 
2. All authentication entry points (WebAuthn, one-time token, JWT) now use named returns with deferred `CreateSignInFailure` calls to guarantee every error path is captured.
3. A new structured logging helper (`logSecurityEvent`) emits WARN-level slog entries with client metadata, and `webAuthnErrorToHTTPStatus` maps 15 WebAuthn protocol errors to proper HTTP status codes. 
4. Dependency injection was updated across the router, auth middleware, and JWT middleware to pass `AuditLogService` through the call chain. 
5. Integration tests cover end-to-end audit log creation for failed sign-ins, WebAuthn protocol errors, and non-security paths, plus unit tests for `CreateSignInFailure` persistence.
##    Regression
  All backend tests pass with no breakage in existing auth or service logic.
##  Residual Risk
   - Log Volume: A large-scale brute-force attack will increase log volume; ensure log rotation or aggregation is configured in the deployment environment.
   - Audit Table Growth: The `audit_logs` table will grow faster with failed attempts, but the existing `ClearAuditLogs` background job (retention based on `AUDIT_LOG_RETENTION_DAYS`) mitigates this.
   - **WebAuthn** Error Surface: The generic error message returned to clients for **WebAuthn** protocol errors avoids information leakage, but the mapping table may need updating if the upstream `go-webauthn` library adds new error types in future versions.
   - Defer-based audit logging: The defer pattern captures the `userID` at the point it is resolved; if a new early-return error path is added before the `userID` is assigned, the audit entry will have an empty `userID`. Care should be taken to always assign `userID` as early as possible in these methods.


ItalyPaleAle (approved):
LGTM!

stonith404:
@crazyrokr I’m not convinced it makes sense to persist failed login attempts in the database. I’d just log them through the normal logging system. With passkeys and login codes, failed login attempts aren’t really a meaningful security signal. Brute-forcing is impossible because of the rate limiting, so a failed attempt by itself doesn’t indicate much.

PR #1495, opened 2026-05-30
feat: Incognito client login

### LINKED ISSUE #939: 🚀 Feature: isolated client authentication
### Feature description

Feature Request to add an option to the create login code page and/or authentication prompt page when using the authorization code flow to use a ‘isolated login’. This login should not set any cookies on the device upon authentication by the user or include any refresh token to the client.

This could be achieved by a seperate button for the user to select when authenticating.

This is distinct from requiring reauthentication on each authorization per client, which is already supported.

### Pitch

When logging into a client service on a ‘less trusted’ device you dont have full control of (eg shared public library computer, friend or family devices) - for example to show photos from an immich album - the user may forget to close a browser or have sign outs incorrectly implemented.

This may lead to unwanted/accidental access to other services and the pocket id instance itself via sso (assuming no requires reauthentication is set).

Having a ‘isolated authentication’ would not stop continued access to the intended service in this scenario, as I believe this is up to the client to invalidate/sign out, but it would prohibit access to any other services or the pocket id instance itself. 

The user could use this isolated login in these instances to access a specific service only.

James18232:
This PR closes #939.

Given there has been significant time since this issue was raised, I decided to have a go at implementing this feature as intended to assist in the event there is a demand, or for use as inspiration.

This feature creates a new incognito button (only on the login code page) that can be used to login without enabling SSO functionality. This is for use when accessing a single service on a device you do not control that does not have a passkey available (eg family members laptop). This function avoids giving that device access to all services and the pocket id instance via the pocket id session cookie, which is currently a (unintended?) side effect.

screenshot of new login code page:
incognito

This is achieved by changing the session cookie expiry to a nominally low value of a few seconds. This makes the pocket id instance itself unusable, and so this feature is limited to clients only.

Please note this is not a trust boundary, just an improvement on existing. I think the pocket ID session cookies need improvement - they are globally scoped and have no cryptographic signing/verification before use. For example even with this PR, a user can simply change the max age in the browser and obtain full sso once again (subject only to a client having require reauthentication enabled). 

A further improvement would be proper server side signing/tracking of cookies and verification before use to enable a proper 'one-time-use cookie' scoped to a specific client.


James18232:
closing this PR as it is superceded by #1562

PR #1496, opened 2026-05-31
fix: Issues with loading new  font

MelvinSnijders:
## What does this PR do?
The new Gloock font wasn't actually loaded by the browser, because the name used was wrong. This PR fixes that.
Browsers defaulted back to their default `serif` font, which causes some styling differences in browsers.
Also, as a small addition, I've added the `ENCRYPTION_KEY` config option to the development `.env` example

## Screenshots
Old:
image

New:
image

## AI Usage
None

## Contributing Guidelines

### DIFF
--- a/backend/.env.development-example
+++ b/backend/.env.development-example
@@ -7,3 +7,6 @@ APP_ENV=development
 APP_URL=http://localhost:3000
 PORT=1411
 MAXMIND_LICENSE_KEY=your_license_key
+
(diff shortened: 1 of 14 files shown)


stonith404 (approved):
Thanks! Not sure how I didn't catch this...

MelvinSnijders:
> Thanks! Not sure how I didn't catch this...

If I was using Chrome instead of Firefox I wouldn't have seen it, the difference on Chrome is very little, haha 🙈

PR #1499, opened 2026-06-01
feat: prompt admin with PKCE client support hint

### LINKED ISSUE #1416: 🚀 Feature: prompt admin when client supports PKCE
### Feature description

Add a visible prompt to admins when Pocket ID detects that a confidential client supports the optional PKCE extension.

### Pitch

PKCE is an optional but useful security enhancement required for public clients, but only optional for confidential clients (in the oauth spec Pocket ID uses).

The issue that arises is that users are rarely able to determine if a confidential client supports using PKCE, and many users probably have no idea where to look. Clients may also provide support later and users may not be aware PKCE support has since been added.

Pocket ID does know this however, as a client will send a code_challenge as part of the authorization flow which is an indication that the client supports PKCE. 

As such, if a user authorisation event fires for a confidential client that does not have PKCE enabled, Pocket ID can alert the admin that PKCE is supported.

This could be in the form of say an orange background for the PKCE toggle and a note that it is supported on hover, or some other mechanism..

I considered pocket ID just auto turning on PKCE in these instances but i don't like the idea of client settings within Pocket ID being controlled/changed by the client.

James18232:
## What does this PR do?
Adds a prompt for server admins when a confidential client appears to support the PKCE extension but is not enabled.

closes #1416 

This can be reset by toggling off PKCE and saving, and will be rechecked next time the user authenticates to the client.

## Screenshots
example_prompt

## Implementation

- new db column in oidc_clients table to track PKCE support for clients, including updates to models and dto as needed.
- new backend logic to check and log PKCE support during authorisation flow, as well as reset flag
- new backend/frontend service logic to fetch/store outcome
- new frontend logic to display prompt to admins

I have deliberately left the ui button component alone and instead wrapped/singled out css changes to this form only, as this is the only exception to the standard component i could see, however this may be benefitial to add to the component as an optional extra..

Same with the prompt color, it may be better to use the selected accent color rather than a fixed color, but i decided this stands out more and reflects the importance of the information being conveyed - but don't have any firm view.

## AI Usage
Used an LLM (chatGPT/Gemini) to understand the relevant parts of the codebase, assist in initial implementation of the feature with opinionated prompts and to troubleshoot issues. All submitted lines have been reviewed and checked by me, and all functionality tested/verified on a test install.

## Contributing Guidelines


backend/internal/service/oidc_service.go
```diff
@@ -924,6 +949,10 @@ func updateOIDCClientModelFromDto(client *model.OidcClient, input *dto.OidcClien
 	client.IsPublic = input.IsPublic
 	// PKCE is required for public clients
 	client.PkceEnabled = input.IsPublic || input.PkceEnabled
+	// Reset any pkce support prompt if previously flagged
+	if !input.PkceEnabled {
```
kmendell, line 222: If the value is already set i feel the client should be except from all of this then no?

James18232: This is an edge case protection - when a client tries to implement pkce but then later rolls back the change due to issues.

An admin would investigate, uncheck pkce for the client and it would work again, but without this addition the changes act like a fuse and the admin will still always get the prompt for that client, with no easy way to reset it.

backend/internal/dto/oidc_dto.go
```diff
@@ -19,6 +19,7 @@ type OidcClientDto struct {
 	PkceEnabled        bool                     `json:"pkceEnabled"`
 	Credentials        OidcClientCredentialsDto `json:"credentials"`
 	IsGroupRestricted  bool                     `json:"isGroupRestricted"`
+	PkceSupported      bool                     `json:"pkceSupported"`
```
ItalyPaleAle: I'd add `omitempty`

backend/internal/service/oidc_service.go
```diff
@@ -143,6 +143,14 @@ func (s *OidcService) Authorize(ctx context.Context, input dto.AuthorizeOidcClie
 		return "", "", &common.OidcMissingCodeChallengeError{}
 	}
 
+	// If a code challenge is provided but confidential client is not using PKCE,flag PKCE support
+	if input.CodeChallenge != "" && !client.PkceEnabled {
```
ItalyPaleAle: We should check if the client has already been flagged to avoid executing a query that wouldn't change anything

frontend/src/routes/settings/admin/oidc-clients/oidc-client-form.svelte
```diff
@@ -73,6 +74,7 @@
 		logoutCallbackURLs: z.array(callbackUrlSchema).default([]),
 		isPublic: z.boolean(),
 		pkceEnabled: z.boolean(),
+		pkceSupported: z.boolean(),
```
ItalyPaleAle:
```suggestion
		pkceSupported: z.boolean().optional(),
```

frontend/src/routes/settings/admin/oidc-clients/oidc-client-form.svelte
```diff
@@ -209,13 +211,19 @@
...
-		<SwitchWithLabel
-			id="pkce"
-			label={m.pkce()}
-			description={m.public_key_code_exchange_is_a_security_feature_to_prevent_csrf_and_authorization_code_interception_attacks()}
-			disabled={$inputs.isPublic.value}
-			bind:checked={$inputs.pkceEnabled.value}
-		/>
+		<div
+			class="rounded-lg transition-all duration-200"
+			class:[&_[data-switch-root]]:ring-2={!$inputs.pkceEnabled.value && $inputs.pkceSupported.value}
```
ItalyPaleAle: IMHO just adding the color is a bit weak as signal and can be confusing. We should at least make the tooltip say "It looks like the client supports PKCE" - something like that

frontend/src/routes/settings/admin/oidc-clients/oidc-client-form.svelte
James18232: i don't understand the benefit of this? this is always available with a default false value?

ItalyPaleAle: I may be wrong but this may be needed if you add `omitempty`, see above

backend/internal/service/oidc_service.go
James18232: thx. Added an extra check to avoid calling the update function if already flagged, and simplified the update function as a result.

frontend/src/routes/settings/admin/oidc-clients/oidc-client-form.svelte
James18232: thanks and noted. I played around a bit and a tooltip is probably too weak - I have added a written prompt to go with the visual prompt instead.

James18232:
E2E test did not complete - might need a rerun, but already passed on earlier commits.

stonith404:
I added an info box, to make it a bit more visible. What do you think?
Screenshot 2026-06-26 at 15 47 46

stonith404 (approved):

James18232:
> I added an info box, to make it a bit more visible. What do you think? 

A banner is a bit generic and risks cluttering the page IMO, and it repeats the description already on the page for pkce. At the end of the day I just want an admin to turn on pkce, and the pkce toggle is where the action is at. 

It does avoid custom one off css changes though, which is a positive (edit - i see it keeps a pale outline - i like this but is maybe not obvious enough - i prefer a more obvious ‘interact with me’ color.

Up to you really - i didnt have a firm view how this should be presented to an admin, i just threw something in how i prefer these types of notifications (direct, relevant and as minimal as possible to get the job done).

kmendell (approved):

ItalyPaleAle (approved):
LGTM but please fix the CI failures!

PR #1520, opened 2026-06-16
refactor: use fosite for OAuth 2.0 logic

### LINKED ISSUE #1471: Bug Report: id_token_hint should be sufficient when no IdP UI session exists (OIDC RP-Initiated Logout § 5)
## Summary

`EndSessionHandler` rejects valid OIDC RP-initiated logout requests when the user has no active Pocket-ID UI session, even though the `id_token_hint` validates and contains a usable `sub` claim. Per [OpenID Connect RP-Initiated Logout 1.0 § 5](https://openid.net/specs/openid-connect-rpinitiated-1_0.html), `id_token_hint` is sufficient authorization for the logout — an active IdP UI session should not be required.

## Impact

Real-world example: a homelab using `traefik-oidc-auth` (the in-process Traefik plugin) as the OIDC client. When the user kills the Pocket-ID UI session out-of-band (e.g. visits `/logout` on Pocket-ID directly), every subsequent RP-initiated logout from any RP fails. The plugin's own session cookie is therefore never cleared (the `Action=Logout` callback never fires), so the user remains logged in to all gated services for up to the cookie lifetime (60 min in our case). This breaks the entire SLO promise.

## Reproduction

Pocket-ID v2.6.2.

1. RP performs a successful login (creates `UserAuthorizedOidcClient` row)
2. Manually clear the Pocket-ID UI session cookie (browser dev-tools, or hit `/logout` on Pocket-ID directly)
3. RP initiates logout: `GET /api/oidc/end-session?id_token_hint=<valid>&post_logout_redirect_uri=<registered>&state=<x>`
4. Pocket-ID logs:

   ```
   WRN Error getting logout callback URL, the user has to confirm the logout manually
       app=pocket-id version=2.6.2 error="missing authorization"
   INF Request status=302 method=GET path=/api/oidc/end-session
   INF Request status=200 method=GET path=/logout
   ```

5. User lands at `/login?redirect=/logout` — the RP never gets the post-logout redirect.

## Root cause (code reference)

In `backend/internal/service/oidc_service.go`, [`ValidateEndSession`](https://github.com/pocket-id/pocket-id/blob/main/backend/internal/service/oidc_service.go) requires `userID` to come from the gin-context middleware:

```go
err = s.db.
    WithContext(ctx).
    Preload("Client").
    First(&userAuthorizedOIDCClient, "client_id = ? AND user_id = ?", clientID[0], userID).
    Error
if err != nil {
    return "", &common.OidcMissingAuthorizationError{}
}
```

The handler bails to `/logout` when this errors:

```go
callbackURL, err := oc.oidcService.ValidateEndSession(c.Request.Context(), input, c.GetString("userID"))
if err != nil {
    slog.WarnContext(c.Request.Context(), "Error getting logout callback URL, the user has to confirm the logout manually", "error", err)
    c.Redirect(http.StatusFound, common.EnvConfig.AppURL+"/logout")
    return
}
```

When `userID == ""` (no UI session), the lookup can't succeed even when the ID token is fully valid and contains the user's `sub`.

## Expected behavior (per spec)

> § 5: The `id_token_hint` parameter ... is used as a hint about the End-User's current authenticated session with the Client.

The token has been validated for signature, issuer, audience, and (intentionally) accepts expired tokens via `acceptExpiredTokens=true` in `JwtService.VerifyIdToken`. That's the proof the OP needs. An active UI session should not be additionally required.

## Suggested fix

Two-line change in `ValidateEndSession`: when `userID == ""`, derive it from the validated token's `sub` claim.

```go
// existing token validation above remains unchanged

if userID == "" {
    sub, ok := token.Subject()
    if !ok || sub == "" {
        return "", &common.TokenInvalidError{}
    }
    userID = sub
}

// existing UserAuthorizedOidcClient lookup follows
```

This preserves the existing security model (the `UserAuthorizedOidcClient` lookup still gates the redirect on prior consent), it just lets the OP accept the token as the proof of identity when there's no UI session.

I'd be happy to send a PR with this change + a unit test covering the no-session path. Let me know if the approach looks right.

## Environment

- Pocket-ID v2.6.2 (chart `anza-labs/pocket-id` v2.1.1)
- Talos OS 1.13 / Kubernetes 1.36, Cilium 1.19
- RP: `traefik-oidc-auth` v0.19.0 ([sevensolutions/traefik-oidc-auth](https://github.com/sevensolutions/traefik-oidc-auth))
- IdP-RP cookie path: shared `Domain=.xx.lv` cookie covering all gated subdomains

🤖 Filed via Claude Code

### LINKED ISSUE #1502: 🐛 Bug Report: Error responses do not follow specification
### Reproduction steps

The error responses sent by the token endpoint do not follow the specification for error codes as defined by [RFC6749](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2).

It stats that the error field **must** be one of the pre-defined values, but Pocket ID sends its own error codes. This means that clients cannot reliably detect and handle error cases.

Such a case is detecting if the refresh token is expired or has already been used. Typically the server would respond with `invalid_grant` and therefore the client knows the issue lies with the refresh token itself (so it knows it needs an entirely new login) and not another configuration error.

### Expected behavior

The `error` field in the response should be one of the enumerated error codes.

### Actual Behavior

Error codes are just arbitrary messages defined in [backend/internal/common/errors.go](https://github.com/pocket-id/pocket-id/blob/main/backend/internal/common/errors.go).

### Pocket ID Version

v2.8.0

### Database

Any

### OS and Environment

Any

### Log Output

_No response_

### LINKED ISSUE #1514: 🐛 Bug Report: OIDC core spec non compliance - max_age/auth_time
### Reproduction steps

Flagging a oidc spec violation where the /authorize endpoint does not looks for and handle the optional max_age authorization request parameter.

The core spec notes this is optional in a request, but the authorization server MUST handle this parameter if it is present in a request, making this mandatory.

refer 3.1.2.1 of the core spec - https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

I understand Pocket ID is aiming for compliance with the Open ID Connect core spec, so this should be handled properly.

A client that is checking for a complaint response may fail as no auth_time is included in the return token, which is required when max_age or auth time is requested.

It acts like a client controlled time based forced reauthentication flag, so should be straightforward to include.

### Expected behavior

Handle a max_age and auth_time in authorisation requests in accordance with the OIDC core spec.

### Actual Behavior

No check is made for either of these parameters, and they are not handled at all by Pocket ID.

### Pocket ID Version

v2.8.0

### Database

SQlite

### OS and Environment

Docker/ubuntu

### Log Output

_No response_

stonith404:
## What does this PR do?
This PR migrates a big part of the OAuth 2.0 logic to [Fosite](https://github.com/pocket-id/fosite). I see the following advantages by this migration:
- Easier to maintain: The library is pretty good designed and extensible. It's easy to add additional RFCs to it. 
- Security: The library was developed by [Ory](https://www.ory.com/) which has a good reputation. I think it's more likely that our custom implementation - maintained by a team of three - has more vulnerabilities than Fosite. Additionally with this change we can focus on security on other parts of the application.
- Spec violations: The OIDC spec and OAuth 2.0 RFC is huge and it's almost impossible to not miss anything. Fosite is very well-tested and with this migration we probably will fix a few spec violations automatically.

Along with the migration to Fosite I decided to change from a per layer module structure (controller, service, etc.) to a per feature module structure. So everything related to OIDC is now inside the `oidc` module, not split in many modules. I think it makes sense to migrate every other part of the application step by step in the future. 

Fixes #1514
Fixes #1471
Fixes #1502

## Review
I know the PR is huge and almost impossible to review :/ In terms of functionality I'm pretty confident that everything is good. I did the following tests:
- OIDC Conformance test
- Ran existing e2e and unit tests
- Added new e2e and unit tests
- Ran 3 reviews with Claude Code in Ultracode mode. Burned 15 mio tokens, which is around 140$ if I would pay for API usage 😅

If you want to review the PR I would focus on checking the new project structure and maybe the most critical features.

## AI Usage
Used Codex for generating boilerplate code e.g implementing the interfaces in `store.go`. Everything was manually reviewed and improved afterwards. Additionally I've used Claude Code to generate unit tests and to review the code.

## Contributing Guidelines


James18232:
Massive PR/refactor...

I guess i will stop working on my PR for max_age/auth_time then.. i was nearly done too 😅.

I was running through the oidc spec because, and Pocket ID was really only missing around the edges - for example I cant even find a client that actually utilizes max_age or then checks it against auth_time in the response.. I have queries around token introspection etc that i haven't had time to interrogate, but feel there may be other areas Pocket ID currently falls down. 

The current situation does have rolling your own auth vibes though.. key elements are not (webauthn comes to mind).

The oidc spec really isn't that big/complex - just an extension to oath and it is unlikely to change substantially. It may be better getting up to speed and submitting for oidc certification to confirm all is in order.

I agree that keeping everything in a oidc module is preferable - jumping around service layers makes you dizzy. 

Regarding Fosite (no shade - cool project), they do not incorporate the latest oidc errata set 2 (latest release is Dec 2024 - 18 months ago), and include auth flows that are generally deemed insecure and should not be in Pocket ID imo (including implicit flow for example). Opinionated deviation from the standards is okay sometimes.

This might just be a more complex path to maintain moving forward, especially if you start branching off of the main ORY over time - just my two cents.

kmendell:
I will look over this as well along with @ItalyPaleAle

stonith404:
Callback URL wildcard matching has been restored, so the branch doesn't contain any breaking changes.

backend/internal/bootstrap/router_bootstrap.go
```diff
@@ -135,9 +137,12 @@ func registerRoutes(r *gin.Engine, db *gorm.DB, svc *services) error {
 	controller.NewScimController(apiGroup, authMiddleware, svc.scimService)
 	controller.NewUserSignupController(apiGroup, authMiddleware, middleware.NewRateLimitMiddleware(), svc.userSignUpService, svc.appConfigService)
 
+	optionalBrowserAuth := authMiddleware.WithAdminNotRequired().WithSuccessOptional().WithApiKeyAuthDisabled().Add()
+	browserAuth := authMiddleware.WithAdminNotRequired().WithApiKeyAuthDisabled().Add()
+	svc.oidcModule.RegisterRoutes(baseGroup, apiGroup, optionalBrowserAuth, browserAuth)
+
 	registerTestRoutes(apiGroup, db, svc)
 
-	baseGroup := r.Group("/", apiRateLimitMiddleware)
```
kmendell, line 140: This may be a dumb question since i haven't gone through everything yet, but is the rate limiting moved to another location? or is it handled by fosite? (which may be the more logical thought)

backend/internal/middleware/cors.go
```diff
@@ -43,7 +43,7 @@ func isCorsPath(path string) bool {
 	switch path {
 	case "/api/oidc/token",
 		"/api/oidc/userinfo",
-		"/oidc/end-session",
```
kmendell, line 46: i know this isn't where its defined, but is switching the end-session endpoint to a new URL a breaking change potentially?

backend/internal/oidc/tx.go
```diff
@@ -0,0 +1,43 @@
...
+func withTx(ctx context.Context, db *gorm.DB, fn func(ctx context.Context) error) error {
+	if _, ok := ctx.Value(txContextKey{}).(*gorm.DB); ok {
+		return fn(ctx)
+	}
+
+	tx := db.WithContext(ctx).Begin()
+	if tx.Error != nil {
+		return tx.Error
+	}
+	defer tx.Rollback()
```
ItalyPaleAle: Nit: consider checking the error returned by tx.Rollback. It could be an error indicating that the transaction was already committed, which should be ignored; otherwise, log it

backend/internal/oidc/federated_client_auth.go
```diff
@@ -0,0 +1,176 @@
...
+	if !ok {
+		return nil, errNoFederatedClientAssertion
+	}
+
+	jwksURL := federatedIdentity.JWKS
+	if jwksURL == "" {
+		jwksURL = strings.TrimRight(issuer, "/") + "/.well-known/jwks.json"
+	}
+
+	jwks, err := a.fetchJWKSet(ctx, jwksURL)
```
ItalyPaleAle, line 151: In the old code we had a cache of JWKS, using httprc. This one seems to be fetching the JWKS on every call, which would be incredibly slow (and if the endpoint were down, it would cause a downtime)

backend/internal/oidc/federated_client_auth.go
```diff
@@ -0,0 +1,176 @@
...
+		jwt.WithIssuer(issuer),
+		jwt.WithSubject(subject),
+		jwt.WithAudience(audience),
+		jwt.WithKeySet(jwks, jws.WithInferAlgorithmFromKey(true), jws.WithUseDefault(true)),
+	)
+	if err != nil {
+		return nil, fosite.ErrInvalidClient.WithHint("Invalid client assertion.").WithWrap(err)
+	}
+
+	// Enforce single use of the assertion to prevent replay
```
ItalyPaleAle: Were we doing this before? A problem I can see is that tokens issued by cloud providers can be cached and re-used for a bit. For example, on Vercel tokens are [cached](https://vercel.com/docs/oidc#in-vercel-functions) for up to 45 minutes

backend/internal/oidc/federated_client_auth.go
```diff
@@ -0,0 +1,176 @@
...
+	exp, _ := parsed.Expiration()
+	if err := a.clients.SetClientAssertionJWT(ctx, jti, exp); err != nil {
+		return nil, fosite.ErrInvalidClient.WithWrap(err)
+	}
+
+	return client, nil
+}
+
+func (a *federatedClientAuthenticator) fetchJWKSet(ctx context.Context, jwksURL string) (jwk.Set, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, jwksURL, nil)
```
ItalyPaleAle: Should probably set a timeout on this context, like 10s

backend/internal/oidc/federated_client_auth.go
```diff
@@ -0,0 +1,176 @@
...
+		jwt.WithAudience(audience),
+		jwt.WithKeySet(jwks, jws.WithInferAlgorithmFromKey(true), jws.WithUseDefault(true)),
+	)
+	if err != nil {
+		return nil, fosite.ErrInvalidClient.WithHint("Invalid client assertion.").WithWrap(err)
+	}
+
+	// Enforce single use of the assertion to prevent replay
+	jti, ok := parsed.JwtID()
+	if !ok || jti == "" {
```
ItalyPaleAle: We should make the jti optional IMHO

ItalyPaleAle (commented):
As discussed this is a very hard PR to review (for the size, but also because the new code lacks comments), so I've mostly focused on spot-checking a few things... Overall it seems good

backend/internal/bootstrap/router_bootstrap.go
stonith404: It just moved up, see here:
https://github.com/pocket-id/pocket-id/blob/255bbfdb184ad843f7e6687985a723c71f3ea2cc/backend/internal/bootstrap/router_bootstrap.go#L125

backend/internal/middleware/cors.go
stonith404: Good catch, but `/oidc/end-session` was actually never correct, `cors.go` referenced to the wrong route.

backend/internal/oidc/federated_client_auth.go
stonith404: Hm okay good call. I was assuming this makes sense for federated client auth as well because for classic OAuth client assertions it's recommended to add replay protection. I didn't know that cloud providers cache their tokens. In that case I have to remove the replay protection.

ItalyPaleAle: Can we make it optional? Like a config option enabled by default? (And if the token has no jti, then no replay protection regardless)

Issue #1517, opened 2026-06-13
🚀 Feature: Dynamic Client Registration


squidink7:
### Feature description

Currently clients must be able to configure their client ID to be used with pocket-id, which some (e.g. Stalwart) do not provide the option to. [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591) describes dynamic client registration which these clients can use to register with OIDC providers dynamically.
Ideally this would come with a toggle to allow administrators to control the ability for these clients to register themselves.

### Pitch

As mentioned above it would allow some clients that cannot be easily configured to store a client ID to work with pocket-id, and would also allow certain applications to dynamically register themselves such as mail clients or Forgejo instances which would otherwise need to be manually added in pocket-id before they can be used.

ovizii:
I used to use OIDC with PocketID and stalwart but since the update from 0.15 to 0.16 they seem to have changed stuff and after researching I ended up here as this seems to be the answer.

jfroy:
This would also be helpful for self-hosting MCP servers (e.g. using toolhive). Most clients (e.g. harnesses like Claude Code, Codex, etc) do not have a client-id. The MCP client ecosystem leans on DCR.

stonith404:
We added support for Client ID Metadata Documents ([CIMD](https://client.dev)), which is a new standard that replaces Dynamic Client registration. Even the MCP protocol marks DCR as deprecated now in favor of CIMD. 

For all of those who are participating in this issue, do you still have the need to use DCR, or does CIMD satisfy your needs?

squidink7:
If CIMD is indeed a replacement for DCR then that seems like the appropriate path to take, it doesn't immediately solve my initial issue of being able to use this with Stalwart but I feel like the best thing would be for them to implement CIMD rather than make any changes here, so all good from me 👍

stonith404:
I'm closing this issue in that case. If you have the need for DCR feel free to comment then we can reopen the issue.

jfroy:
For what it's worth, the OpenAI codex CLI still only support DCR.

```
❯ codex mcp add flux-mcp --url <...>
Added global MCP server 'flux-mcp'.
Detected OAuth support. Starting OAuth flow…
Error: Registration failed: Dynamic registration failed: Registration failed: Dynamic client registration not supported
```

contre95:
@stonith404 
> For all of those who are participating in this issue, do you still have the need to use DCR, or does CIMD satisfy your needs?

[...]

PR #1526, opened 2026-06-17
feat: implement OAuth Client ID Metadata Document (#1525)

### LINKED ISSUE #1525: 🚀 Feature: OAuth Client ID Metadata Document
### Feature description

Implement https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-01 to support MCP client registration (see https://modelcontextprotocol.io/specification/draft/basic/authorization/client-registration).

### Pitch

OAuth Client ID Metadata Document is the preferred method for MCP client registration.

jfroy:
## What does this PR do?

This PR implements https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-01. The functionality is gated by an environment variable (disabled by default).

This RFC is part of the MCP protocol for client registration. See https://modelcontextprotocol.io/specification/draft/basic/authorization/client-registration.

closes #1525

## AI Usage

Implemented with Claude Opus 4.8 assistance. All code reviewed by myself. Backend tests run on workstation. Deployed on homelab cluster and tested with ToolHive and Claude Code CLI.

## Contributing Guidelines

Have you read the [Contributing Guidelines](https://github.com/pocket-id/pocket-id/blob/main/CONTRIBUTING.md)? YES


jfroy:
Rebased on top of the fosite refactor.

jfroy:
Looking at the test failures.

jfroy:
OK, should all be resolved. I took the time to setup and run the e2e tests locally. Sorry for the churn!

jfroy:
From a design perspective, there are 2 topics to review

- Using "~[base64 encoded client ID]" for CIMD clients URLs (between frontend and backend). It's kind of ugly, but it maintains API compatibility and the URL structure while allowing the full CIMD client ID, which is a full URL, to pass through proxies and the frontend router.
- Allowing editing of CIMD clients from the admin UI without any restrictions. Once created, CIMD clients are treated the same as any other. Perhaps the admin UI should have some fields disabled or restricted.

There may also be additional security work, particularly [§6.1 Relationship between redirect_uris and client_id or client_uri](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-01#name-relationship-between-redire).

There is also no handling of changes to client metadata beyond audit logging. Stricter policies may be useful, as discussed in [§6.3 Changes in Client Metadata](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-01#name-changes-in-client-metadata).

Finally, [§6.7 Displaying Logos to End-Users](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-01#name-displaying-logos-to-end-use) is not implemented by this PR.

stonith404:
Thanks for your contribution. I've never heard of CIMD before but I took some time now to read into it. I haven't fully reviewed it yet but here are some points that I would like to suggest/discuss:

- I agree encoding the client ID as base64 isn't very clean but I can't really find out a better solution, so this is fine for me.
- I think we should definitely treat CIMD clients differently, the only source of truth should be the client, an admin shouldn't be able to change metadata of the client. Data that isn't provided by the CIMD client like group restrictions etc. should still be editable though. Additionally, I would mark CIMD clients, so that they can be filtered out in the OIDC client list.
- Even though this isn't defined in the spec don't you think it would make sense to create to create an option to whitelist which URLs are allowed to create a CIMD client? We could use the same URL wildcard matcher like we do with the callback URLs. Currently everyone could create clients and therefor completely fill the database.
- The migration to Fosite is pretty new, we haven't even created a release yet and therefor we didn't update the contribution guide yet, sorry. But I think it would be nice that we move as much of generic OAuth 2.0 logic to our [Fosite fork](https://github.com/pocket-id/fosite). For example the logic for resolving the client (with all the SSRF protection etc.) could easily be moved there.

jfroy:
The spec is a draft, so this is very new and not settled stuff. But because MCP adopted it, it's kind of on a fast track IMO.

> * I think we should definitely treat CIMD clients differently, the only source of truth should be the client, an admin shouldn't be able to change metadata of the client. Data that isn't provided by the CIMD client like group restrictions etc. should still be editable though. Additionally, I would mark CIMD clients, so that they can be filtered out in the OIDC client list.

Sounds reasonable. I'll iterate on that.

> * Even though this isn't defined in the spec don't you think it would make sense to create to create an option to whitelist which URLs are allowed to create a CIMD client? We could use the same URL wildcard matcher like we do with the callback URLs. Currently everyone could create clients and therefor completely fill the database.

Sounds like a reasonable policy knob for the administrator. On the flip side, the point of having dynamic clients is to avoid requiring the administrator to pre-register.

Perhaps there also should be some sort of TTL/GC process for unused dynamic clients (this will also come up for dynamic client registration #1517).

[§6.8 Client ID Domain Trust](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-01#name-client-id-domain-trust) has a few more ideas in that space.

> * The migration to Fosite is pretty new, we haven't even created a release yet and therefor we didn't update the contribution guide yet, sorry. But I think it would be nice that we move as much of generic OAuth 2.0 logic to our [Fosite fork](https://github.com/pocket-id/fosite). For example the logic for resolving the client (with all the SSRF protection etc.) could easily be moved there.

I didn't know you had forked. Some functionality could be moved there, indeed. The interaction with the database probably should stay in pocket-id.

stonith404:
> Sounds like a reasonable policy knob for the administrator. On the flip side, the point of having dynamic clients is to avoid requiring the administrator to pre-register.

I agree but if we would use the same wildcard matching logic as with callback URLs this would be pretty flexible. An admin could just enter `*` as an URL and this would remove any restrictions. This is the kind of input I have in mind:
Screenshot 2026-06-29 at 12 19 02

> Perhaps there also should be some sort of TTL/GC process for unused dynamic clients (this will also come up for dynamic client registration https://github.com/pocket-id/pocket-id/issues/1517).

Yes this would make sense. I would recommend to create a cleanup job that deletes a CIMD client that hasn't been updated more than 6 months.

jfroy:
Foasite PR: https://github.com/pocket-id/fosite/pull/1

I'll update this PR once the fosite side is in so I can update the fosite package in go.mod.

jfroy:
I've pushed an updated commit rebased on top of the fosite PR that includes the requested additional functionality (client ID URL allow list, client type column for filtering and sorting in admin UI, disabling edits in client details for dynamic clients, and database cleanup job).

jfroy:
Rebased. I have a [mcp branch](https://github.com/jfroy/pocket-id/tree/mcp) that picks up from this one and adds [RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707), [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414), and [OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html).

stonith404:
@coderabbitai review

jfroy:
Thanks for the patches! I can do another review pass once CRAI is settled.

stonith404:
@jfroy Thanks! I'm planning to move more logic related to CIMD into the Fosite fork tomorrow, but after that I'm done with my changes. I'll let you know as soon as you can review it.

stonith404 (approved):
Everything looks good from my side. Before we merge it we just have to replace your library in `go.mod`. 

Let me know if you agree with my changes.

jfroy:
I'll take a look probably tomorrow. And yep, need to land the fosite changes first then we can fix up `go.mod`.

kmendell (approved):
Also looks good to me from the quick pass I did earlier tonight.

stonith404:
Did some additional hardening based on Claudes feedback. Additionally I removed automatic client rentation because it's risky in my opinion. If a client gets configured by an admin that it is only accessible by some user groups and the client gets deleted because of the rentation policy, the client can re-register itself and all restrictations are gone.

jfroy:
🚀

PR #1542, opened 2026-06-26
feat: add OAuth APIs with scoped permissions

stonith404:
## What does this PR do?

This PR adds support for "APIs" to Pocket ID, modeled after Auth0's API/permissions concept. This allows clients to request specific scopes for a specific audience. 

This can useful if you have e.g an SPA which needs to access resources in the backend. The SPA can request the access token and send it to the backend. The backend only needs to validate the tokens audience, scope and signature - no session management needed in the backend.

In a follow-up PR I'm planning to add a "native Pocket ID" API. This will allow clients manage settings in Pocket ID in behalf of the user. This is useful e.g when you want that a user can update it's name directly in your app.

##  Screenshots

Create an API:
Screenshot 2026-06-26 at 14 58 31

Edit an API / add permissions to an API:
Screenshot 2026-06-26 at 14 56 22

Assign an API and permissions to a client:
Screenshot 2026-06-26 at 14 56 37

Screenshot 2026-06-26 at 14 56 49

## AI Usage
Used Claude Code to bootstrap the backend API module and write unit and e2e tests. Reviewed every file and tested with [oauth.tools
](https://oauth.tools)

## Contributing Guidelines


stonith404:
@coderabbitai review

backend/internal/api/dto.go
```diff
@@ -0,0 +1,63 @@
+package api
+
+import (
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+)
+
+// apiResponseDto is the full representation of an API including its permissions
+type apiResponseDto struct {
+	ID          string                     `json:"id"`
```
ItalyPaleAle, line 9: We could make the Audience be be ID too since it's unique and immutable

backend/internal/api/dto.go
```diff
@@ -0,0 +1,63 @@
...
+	Name        string                     `json:"name"`
+	Audience    string                     `json:"audience"`
```
ItalyPaleAle: Would "Resource" be a better name here? Since this identifies a resource provider (audience is what tokens are requested for)

backend/internal/api/models.go
```diff
@@ -0,0 +1,36 @@
+package api
+
+import (
+	"github.com/pocket-id/pocket-id/backend/internal/model"
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+)
+
+type API struct {
```
ItalyPaleAle, line 9: Why is this model not placed together with the other models?

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
+package api
```
ItalyPaleAle, line 1: Why is this service not placed in the service folder?

backend/internal/api/module.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return infos, nil
+}
+
+// RegisterRoutes mounts the admin CRUD endpoints
+// adminAuth is passed in as a gin handler so the module does not import internal/middleware
+func (m *Module) RegisterRoutes(apiGroup *gin.RouterGroup, adminAuth gin.HandlerFunc) {
+	apis := apiGroup.Group("/apis")
+	apis.Use(adminAuth)
+	{
```
ItalyPaleAle: Nit, if you want to use scopes properly
```suggestion
	{
	    apis := apiGroup.Group("/apis")
	    apis.Use(adminAuth)
```

backend/internal/api/module.go
```diff
@@ -0,0 +1,81 @@
...
+		apis.GET("/:id", m.handler.get)
+		apis.PUT("/:id", m.handler.update)
+		apis.DELETE("/:id", m.handler.delete)
+		apis.PUT("/:id/permissions", m.handler.updatePermissions)
+	}
+
+	// The per-client API-access allow-list lives on a separate path so it does not collide with the /apis/:id wildcard
+	access := apiGroup.Group("/api-access")
+	access.Use(adminAuth)
+	{
```
ItalyPaleAle: Nit, if you want to use scopes properly
```suggestion
	// The per-client API-access allow-list lives on a separate path so it does not collide with the /apis/:id wildcard
	{
	    access := apiGroup.Group("/api-access")
	    access.Use(adminAuth)
```

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+
+	"github.com/pocket-id/pocket-id/backend/internal/common"
+	"github.com/pocket-id/pocket-id/backend/internal/model"
+	datatype "github.com/pocket-id/pocket-id/backend/internal/model/types"
+	"github.com/pocket-id/pocket-id/backend/internal/utils"
+)
+
+// reservedPermissionKeys are the scope and claim names owned by Pocket ID's built-in identity layer
+// A custom API permission must not reuse one, otherwise its scope string would collide with a standard OIDC scope or claim
+var reservedPermissionKeys = map[string]struct{}{
```
ItalyPaleAle: Suggest creating a function like "isPermissionKeyReserved(name string) bool" which has a `switch` inside

This avoids allocating a map in the package

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+	"profile":        {},
+	"email":          {},
+	"email_verified": {},
+	"groups":         {},
+	"offline_access": {},
+}
+
+// permissionKeyPattern restricts permission keys to RFC 6749 scope-token characters, which are printable ASCII without space, double-quote or backslash
+// This keeps a key safe as a space-delimited value in the token scope claim and free of the control character used to qualify consent records
+var permissionKeyPattern = regexp.MustCompile(`^[\x21\x23-\x5B\x5D-\x7E]+$`)
```
ItalyPaleAle: Can probably create a function here that loops through each byte and finds the disallowed one. Would be faster to run and easier to manage than a regex?

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func newService(db *gorm.DB, issuer string) *Service {
+	return &Service{db: db, issuer: issuer}
+}
+
+// isIssuerAudience reports whether the audience refers to Pocket ID itself (the issuer)
+// A custom API must not claim the issuer as its audience, otherwise its tokens would be indistinguishable from Pocket ID's own identity tokens
+func isIssuerAudience(audience, issuer string) bool {
+	return issuer != "" && strings.EqualFold(strings.TrimRight(audience, "/"), strings.TrimRight(issuer, "/"))
```
ItalyPaleAle: 1. Issuer never has the trailing `/` since it comes from the config
2. We don't need unicode-aware comparison here

```suggestion
	return issuer != "" && strings.ToLower(audience) == issuer
```

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+
+func (s *Service) List(ctx context.Context, search string, listRequestOptions utils.ListRequestOptions) (apis []API, response utils.PaginationResponse, err error) {
+	query := s.db.
+		WithContext(ctx).
+		Preload("Permissions").
+		Model(&API{})
+
+	if search != "" {
+		like := "%" + search + "%"
+		query = query.Where("name LIKE ? OR audience LIKE ?", like, like)
```
ItalyPaleAle, line 65: Do we need this LIKE? since it won't use any index it will be very slow, scanning each row and each value inside

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+		Error
+	if err != nil {
+		return nil, err
+	}
+
+	return valid, nil
+}
+
+func (s *Service) getInternal(ctx context.Context, id string, tx *gorm.DB) (api API, err error) {
+	err = tx.
```
ItalyPaleAle: This needs a FOR UPDATE clause since sometimes we retrieve the row to update it

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+		}
+		if err = tx.WithContext(ctx).Create(&newPermission).Error; err != nil {
+			return API{}, err
+		}
+	}
+
+	err = tx.WithContext(ctx).
+		Model(&API{}).
+		Where("id = ?", api.ID).
+		Update("updated_at", new(datatype.DateTime(time.Now()))).
```
ItalyPaleAle, line 236: Is this needed? I thought Gorm set it automatically, but i may be wrong

backend/internal/api/service.go
```diff
@@ -0,0 +1,429 @@
...
+		Joins("JOIN api_permissions ON api_permissions.id = g.api_permission_id").
+		Joins("JOIN apis ON apis.id = api_permissions.api_id").
+		Where("g.oidc_client_id = ?", clientID).
+		Scan(&rows).
+		Error
+	if err != nil {
+		return nil, nil, err
+	}
+
+	scopeSeen := make(map[string]struct{}, len(rows))
```
ItalyPaleAle, line 363: Suggest preallocating scopes and audiences too with capacity `len(rows)`:

```go
scopes = make([]string, 0, len(rows))
audiences = make([]string, 0, len(rows))
```

backend/internal/oidc/store.go
```diff
@@ -89,7 +90,19 @@ func (s *Store) GetClient(ctx context.Context, id string) (fosite.Client, error)
 		return nil, err
 	}
 
-	return Client{OidcClient: client}, nil
+	client := Client{OidcClient: clientModel}
+
+	//  Populate the API scopes and audiences for the client if the API access provider is available
+	if s.apiAccess != nil {
+		apiScopes, apiAudiences, err := s.apiAccess.ClientAPIScopes(ctx, id)
```
ItalyPaleAle: We probably need a transaction here?

ItalyPaleAle (commented):
First pass!

backend/internal/api/service.go
stonith404: I'm planning to migrate everything to a feature based module structure, instead of layer based. I already did the same with the Fosite migration, everything is inside the `oidc` module there.

backend/internal/api/dto.go
stonith404: URLs as IDs isn't a good solution from my experience because they include slashes. This causes problems with URL params like `GET /api/https://my-audience.com`.

backend/internal/api/models.go
stonith404: Same reason like [here](https://github.com/pocket-id/pocket-id/pull/1542#discussion_r3486360864).

backend/internal/api/service.go
stonith404: We do use a `LIKE` everywhere else for searching trough a table. In my opinion it makes sense that a search query searches with `LIKE` or do you have a better solution that is still user friendly?

backend/internal/api/dto.go
ItalyPaleAle: Good point!

backend/internal/api/service.go
ItalyPaleAle: Uhm that is true. The "right" approach would be to add a full-text serach index, but it's probably an overkill.

A somewhat simpler approach could be to:

1. Add an index on the `LOWER(name)` and `LOWER(audience)`
2. Use a `LIKE` with a `%` at the end only and the `LOWER` function on the query too

With these two it should _in theory_ be possible to use an index (but we should validate it with an `EXPLAIN` query to be sure)

stonith404: That sounds like a plan, we shouldn't do this in this PR though, because we would have to adapt this in other places too.

ItalyPaleAle: How about doing it in this PR just for this case?

backend/internal/api/service.go
stonith404: It's needed here because we are no updating the "API" entity itself, we are updating the permissions associated to the API. Updating a relation doesn't update the timestamp.

ItalyPaleAle: ACK. This is the kind of stuff that I'd suggest adding as a comment

stonith404:
Thanks for the review @ItalyPaleAle. 

You've suggested that we implement it similarly like in Auth0.  Do you think the implementation looks good or is something important missing compared to Auth0?

kmendell:
It looks good to me, leaving to italy for a final review though.

ItalyPaleAle:
Ok I did a full review and also done more tests with oauth.tools!

Overall this is solid work :) I opened a [PR](https://github.com/pocket-id/pocket-id/pull/1561) against this branch with some fixes, but nothing major.

A few additional things for your consideration:

1. When defining an API, there's no difference between scopes that are for delegated access and client access. On Auth0, for example, when you assign a API to an app, you mark it as either User-Delegated Access, or Client Access (a scope can be added to both). This is something we probably want too, since the scopes can be different   
   image

2. When the token is requested with the client_credentials, if there's no list of scopes requested we could include all scopes for the resources by default. This is something that I believe Entra ID does too, for convenience.
3. In the `/authorize` (OBO flow) and `/token` (client credentials) endpoints, can we make the query string arg `audience` as an alias for `resource`? A lot of people use that, and it's even in the official [examples](https://auth0.com/docs/get-started/apis/scopes/sample-use-cases-scopes-and-claims#request-custom-api-access) for Auth0

stonith404:
@ItalyPaleAle Your feedback should be implemented. 

> When the token is requested with the client_credentials, if there's no list of scopes requested we could include all scopes for the resources by default. This is something that I believe Entra ID does too, for convenience.

I have implemented this into Fosite directly.

> In the /authorize (OBO flow) and /token (client credentials) endpoints, can we make the query string arg audience as an alias for resource? A lot of people use that, and it's even in the official [examples](https://auth0.com/docs/get-started/apis/scopes/sample-use-cases-scopes-and-claims#request-custom-api-access) for Auth0

I did some more research about the `audience` claim and I think this shouldn't be an alias for `resource` because there is an actual `audience` claim available for the `/token` endpoint that is different to the `resource` parameter. See [the spec](https://datatracker.ietf.org/doc/html/rfc8693#name-token-exchange-request-and-)

ItalyPaleAle (approved):

kmendell (approved):

PR #1545, opened 2026-06-27
feat: login code input boxes

### LINKED ISSUE #1532: 🚀 Feature: login code - UX change - individual code boxes
### Feature description

Feature request to update the existing login code combined input text box with login code fields of the right length.

Mock up below to explain:

Image

### Pitch

This will improve the user experience and make pocket ID easier to use, as the user intuitively knows to look for the right length code.

This may also help with automatic targeting on devices to speed up code entry.

James18232:
## What does this PR do?

Implements new login code pages with input boxes for the login code. The existing user login code url generation is also maintained.

closes #1532

The primary motivation for this was the login code ui proposed in pr 1402 by @Norrodar, and the base of the new login component is taken from that implementation, modified to work with the login code for the main auth flow.

This code-input component hides an input from the user and displays it back in a user friendly way. This stops issues on mobile devices with keeping focus on the input and works well as a OTP type component. This is where the real magic is and so credit to them.

The main modification i have made to this component were:

- to make it a ui component only, with all auth functions handled by the parent page instead
- remove any changing of input prior to display (uppercase etc)
- the box layout altered to accommodate 6 and 16 digits upper/lowercase alphanumeric codes, which are the two currently supported login codes for Pocket ID.
- on smaller screens the component switches to a strict grid column layout and the large code is split over two rows with an svg arrow to visually link the two together.

In addition to the new component, modifications to the front end were made to replace the existing login code form:

routes/interaction:
- modification of the interaction page to flag when long codes are permitted, as fosite now hides this behind an interaction id.  

components/login-wrapper:
- modification of the login-wrapper component to accept the flag from the interaction page and pass a parameter to the alternative login code page, and add extra width for the long login code.

routes/login/alternative/code:
- modification of the alternative login code page to incorporate the new code-input component, handle code submission and control visibility of code input elements.

## Outstanding queries

- I have left an unused onsubmit input parameter in the new component from the original component, as the auth for PR 1402 happens inside this component. I left this in the event the component can be common if this PR is adopted, but the layout is a lot different and it may be better to remove this and add a separate component for that purpose..

## Screenshots
large screen (Pocket ID) - short code:
pid_desktop_short_code_strip-metadata_compress

large screen (Pocket ID) - long code:
lc_pid_long_desktop_strip-metadata

large screen (Client) - short code:
client_desktop_short_compress_strip-metadata

small screen (Pocket ID) - short code:

lc_pid_strip-metadata

small screen (Pocket ID) - long code:
lc_pid_long_mobile_strip-metadata

small screen (Client) - short code:
lc_client_strip-metadata

## AI Usage
AI was used quite heavily in this PR (LLM - gemini/chatgpt), primarily to assist in coding css classes/divs/layout for the fake code input boxes within the new code-input component. I had a lot of difficulty getting these ui components to sit correctly/look right with the different possible permutations and tailwind classes not matching raw css equivalents, as well as issues trying to stop password managers attempting to auto fill the one time code.

I have reviewed all changes and tested these on a test instance, with all permutations and login options.

## Contributing Guidelines


stonith404:
Thanks for this PR, two things:
- I think it would be better if we use this [shadcn component](https://www.shadcn-svelte.com/docs/components/input-otp) as we are using shadcn for all other components
- I would fallback to a classic input for longer codes, in my opinion the input takes too much space and doesn't look good for 16 characters.

James18232:
Sure - i will look into the shadcn option.

And yes there is only so much you can do with a 16 character code on a limited display - that was the best i could think of. 

A standard input there makes sense, and it will rarely be used manually anyway.

James18232:
I have switched to a shadcn OTP component and reverted to a classic input for long codes as requested. It is a simpler implementation as a result.

revised

stonith404 (approved):

frontend/src/lib/components/login-wrapper.svelte
```diff
@@ -113,7 +106,7 @@
 	</div>
 {:else}
 	<div
-		class="flex min-h-dvh items-center justify-center bg-cover bg-center text-center"
+		class="flex h-screen items-center justify-center bg-cover bg-center text-center"
```
James18232, line 109: sorry, should have mentioned this was to fix a mobile ui bug where the login code is nearly off-screened on a mobile - firefox only (webkit - not an issue on safari though). the h-screen does not handle a keyboard popping up on a mobile very well and the layout jumps around. switching to min-h-dvh fixes this issue and works cross browsers as far as i tested.

Issue #1549, opened 2026-06-28
Tracking issue: High Availability / horizontal scaling support


ItalyPaleAle:
This is a tracking issue that includes things that need to be fixed/changed before Pocket ID can scale horizontally on multiple replicas.

A baked-in assumption is that we're going to leverage an actor framework for synchronizing state across multiple replicas and for orchestrating background jobs.

## Core blocker

- [x] **Remove single-instance app lock** — DB `application_lock` exits any second replica on startup.
  -  Initial implementation #1624

## In-memory state to share/coordinate

- [x] **App config cache** — `atomic.Pointer` config only refreshes in the process that wrote it
   - #1604 
- [x] **Instance ID**  — currently maintained inside App Config
   - #1579 
- [x] **Rate limiter** — per-process IP-based limiter
   - #1556
- [ ] **Minor read caches** — version check, app-images extension map
  - No need to address the external JWKS cache, can stay per-process
- [ ] **JWT signing keys** - coordinate creation of keys upon startup of a new cluster
   - Necessary fixes in #1733 

## Background jobs (run on exactly one instance)

- [x] **LDAP sync** — must not run concurrently across replicas
   - #1651
- [x] **SCIM sync** — must not run concurrently across replicas
   - #1680 
- [x] **API key expiry job** — single run cluster-wide
   - #1687 
- [x] **Analytics job** — single run to avoid duplicate pings
   - #1556
- [x] **GeoLite update job** — single run, tied to shared GeoLite storage below
   - #1647 
- [x] **Orphaned temp file cleanup** — single run (filesystem backend only)
   - #1556

## Local-filesystem dependencies

- [ ] **Filesystem storage backend** — local disk isn't shared, require DB/S3 backend for multi-instance
- [ ] **Default profile picture caching** — generated avatars written to local disk per-instance
- [x] **GeoLite DB on local disk** — per-instance copy + download
   - #1647 

See note: https://github.com/pocket-id/pocket-id/pull/1647#issuecomment-5186855645

## Migrate to actor + alarm (and delete the cleanup job)

Each of these is currently expired by `db_cleanup_job.go`. Move expiry into the owning actor's alarm and remove the matching cleanup function.

- [x] **WebAuthn sessions** — actor with expiry alarm, drop `ClearWebauthnSessions`
   - #1698
- [x] **One-time access tokens** — actor with expiry alarm, drop `ClearOneTimeAccessTokens`
   - #1611 
[...]

PR #1556, opened 2026-07-01
Integrate Francis actor framework for background jobs, cron scheduling, and rate limiting

ItalyPaleAle:
This PR is part of #1549

## What does this PR do?

Introduces [Francis](https://francis.italypaleale.me/) as the actor framework that will eventually allow Pocket ID to scale horizontally.

This initial PR focuses on bootstrapping Francis so it's integrated into Pocket ID. It is limited to using "built-in actors" (that ships with Francis itself) for now, which are used for Pocket ID's background work.

Treat this initial PR as a demonstration of what an actor framework can do 😄 

- Bootstrap a Francis actor system wired into the DB and app lifecycle
- Migrate the analytics cron job and file cleanup jobs to run as Francis actors (as example of jobs that can be orchestrated by Francis)
- Move HTTP rate-limiting (signup, user, WebAuthn) onto Francis instead of the old in-process limiter
- Swap `github.com/glebarez/sqlite` for `github.com/libtnb/sqlite`, since the former pulls in a forked `modernc.org/sqlite` which was causing conflicts

Note: import/export for data stored in Francis will be added later.

## AI Usage

This PR was written in part by hand (especially the initial bootstrap) and in part with help from Opus, under my direct supervision.


backend/internal/utils/crypto/key.go
```diff
@@ -0,0 +1,25 @@
...
+	"crypto/sha3"
+	"errors"
+	"fmt"
+	"hash"
+
+	"github.com/pocket-id/pocket-id/backend/internal/common"
+)
+
+// DeriveKey derives a key using HMAC from the configured EncryptionKey and a seed
+func DeriveKey(envConfig *common.EnvConfigSchema, seed string) (key []byte, err error) {
```
stonith404: Can't we pass only the encryption key as string as param instead of the whole env config struct?

kmendell:
Im gonna spin up a test insatnce and test this out later tonight or tomorrow.

backend/internal/utils/db_migration_util.go
```diff
@@ -127,20 +153,29 @@
...
 		}
 	}
 
-	return maxVersion, nil
+	if maxVersion > uint64(^uint(0)) {
+		// We do not support 32-bit systems
+		panic("32-bit systems are unsupported")
+	}
+
+	return uint(maxVersion), nil
```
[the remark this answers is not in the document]
ItalyPaleAle: This is wrong, there's an explicit check to enforce we're on a 64-bit system

stonith404 (commented):
Cool, thank you! One comment but the rest looks good to me.

backend/internal/utils/crypto/key.go
ItalyPaleAle: Done, I've also added new tests with fixtures to prevent people from accidentally changing how keys are derived (we do it in 4 places today)

stonith404 (approved):

PR #1561, opened 2026-07-03
[feat/apis-1542] Fixes for OAuth APIs/scopes

ItalyPaleAle:
**This PR is built on top of #1542**

## Description

A few fixes:

1. Tokens issued to a custom audience _only_ cannot invoke the `/userinfo` endpoint in Pocket ID. When requesting the `openid` scope (or `email`, `profile`), tokens are now created with the added audience of `https://<pocket-id-address>/`, which grants access to the `/userinfo` endpoint too (so, tokens may have more than one audience).  Also makes the `/userinfo` endpoint enforce that tokens have the audience of Pocket ID itself, in addition to the openid scope.
2. Fixes some minor bugs, including the log message "transaction already committed" after successful transactions for fosite
3. UI: offline_access should be rendered as a built-in API
4. UI: device flow consent page cna't show description for custom audiences
5. Reject duplicate API permission keys rather than just collapsing them, since that indicates a request error
6. Ensure that errors do not disclose available APIs
7. Reject multiple `resource` parameters
8. Added more tests (also switched keys in test from RSA to ECDSA, since they generate "instantly")

## AI usage

This PR includes a mix of hand-written changes and changes made by Opus under my supervision, especially for writing the added tests. I have also used Claude to do some additional code review, which highlighted a few of the bugs/issues below.

### DIFF
--- a/backend/internal/api/service.go
+++ b/backend/internal/api/service.go
@@ -177,7 +177,9 @@ func (s *Service) UpdatePermissions(ctx context.Context, id string, input apiPer
 		return API{}, err
 	}
 
-	// Reject keys with invalid characters or that collide with Pocket ID's reserved scopes and claims before persisting anything
+	// Reject keys with invalid characters, that collide with Pocket ID's reserved scopes and claims, or that repeat within the request before persisting anything
+	// A duplicate key would otherwise be silently coalesced last-wins into the map below, dropping a row behind a 200
+	seen := make(map[string]struct{}, len(input.Permissions))
 	for _, permission := range input.Permissions {
 		if !isValidPermissionKey(permission.Key) {
 			return API{}, &common.ValidationError{Message: fmt.Sprintf("the permission key %q contains invalid characters", permission.Key)}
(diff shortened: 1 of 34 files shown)


stonith404:
> Tokens issued to a custom audience should not have the standard OIDC claims like openid, email, etc. That info is not meant for external audiences, and it can be provided by the id_token anyways if requested. Also makes the /userinfo endpoint enforce that tokens have the audience of Pocket ID itself, not an external API.

I would prefer to add Pocket ID (`https://pocket-id-instance.com/userinfo`) as a second audience if a resource (external API) is requested plus the openid scope. This allows the client to call the `userinfo` endpoint even though a external resource was requested.

Auth0 does this as well. Here is how the access token payload looks like if you make a request to `/authorize` with `resource=my-custom-api`:

```
{
  "iss": "https://dev-yb9xmfe4.us.auth0.com/",
  "sub": "auth0|67695ced074719d14fc899fd",
  "aud": [
    "my-custom-api",
    "https://dev-yb9xmfe4.us.auth0.com/userinfo"
  ],
  "iat": 1783059860,
  "exp": 1783146260,
  "scope": "openid email a-custom-scope",
}
```

ItalyPaleAle:
@stonith404 Your comment makes sense and I've updated the PR with the change.

One difference, however: for audience I think we should use `https://<pocket-id-address>/` and NOT `https://<pocket-id-address>/userinfo`

- First, our userinfo endpoint is not at `/userinfo`
- But more importantly, we should be ok just using `https://<pocket-id-address>/` : tokens also need the openid scope to request the userinfo endpoint (that was pre-existing), so we don't need a separate audience (which makes things more complicated). This way we could also add more APIs to pocket ID itself in the future (should the need arise) and we just need the scope, not a new audience

PR #1562, opened 2026-07-03
feat: Incognito client login

### LINKED ISSUE #939: 🚀 Feature: isolated client authentication

James18232:
## What does this PR do?
This PR closes #939 

This is a revised PR that implements a new approach to this feature request, updated to work post fosite.

This feature creates a new incognito button (only on the login code page) that can be used to login without enabling SSO functionality. This is for use when accessing a single service on a device you do not control that does not have a passkey available (eg family members laptop). This function avoids giving that device access to all services and the pocket id instance via the pocket id session cookie, which is currently a (unintended?) side effect.

screenshot of new login code page:
## Screenshots
incog_strip-metadata

## Implementation
key change:

This is now achieved via the introduction of a new jwt access token type which i have called an "isolated-token". 

When a user logs in with incognito mode, a new isolated-token is minted. This isolated token includes the client it is linked to via a "permitted_clients" custom claim in the jwt, and this is used to enforce isolated authentication for that client in Pocket ID. An isolated-token will not work on other clients or pocket id itself.

This is not single use, but is limited to a single client.

The backend and frontend both needed updates to implement this. in particular:

backend:

- tweaked the auth middleware to permit both access and isolated tokens as valid token types for oath login flows
- a check is undertaken as part of the authorisation handler that resets a user if trying to authenticate to any other client, sending them back to the start of the interaction flow.
- the JWT minting and verifying process has been updated to accommodate the new token type

frontend:

- new incognito button and associated plumbing to enable the functionality client side

## Questions

There is strong reasoning to remove the incognito button and make this the default login code mode (perhaps with a remember me button as identified in the issue). I have kept the existing functionality in tact for now however ..

## AI Usage
AI used (LLM's - chat gpt/gemini) to help develop and refine the code as well as troubleshoot issues. All changes to the codebase have been checked and tested by me on a test instance to verify all functionality.

## Contributing Guidelines


James18232:
I have updated this implementation to move the isolated code generation/election server side and on the trusted device, as on reflection having any part of the process on an unknown client device is a bad architectural decision.

The user can instead generate a login code on a client page within pocket ID, and this code is used like the regular code on the now unchanged login code page in exchange for an isolated-token. This also keeps the ui largely the same as existing and reuses existing components.

The remainder of the logic is the same.

James18232:
I think i am happy with this implementation now.

i forgot not all users are admins, so have just tweaked the existing login code to include a dropdown to pick a client instead.

example

James18232:
separate to this but related, is it possible to change the name of completeInteractionStep to something else like performInteractionStep?  i always think its the 'final' interaction step rather than a placeholder for any number of interaction steps. Trying to follow along with all the to and fro between interaction pages and authorize endpoints makes my head hurt..

PR #1564, opened 2026-07-05
fix: various bugs in observability / OTel

ItalyPaleAle:
## What does this PR do?

This PR contains a number of fixes in the observability (OTel) space and related improvements.

1. **(Sort of breaking change)** Config parameters METRICS_ENABLED and TRACING_ENABLED are redundant. Tracing can be enabled by setting OTEL_TRACES_EXPORTER and OTEL_METRICS_EXPORTER for metrics (set them to "otlp" for OTel)
2. Fix shutdown sequence when OTel is configured, sometimes causing locks not to be released. Because servicerunner (both the one in Pocket ID 2.x and in the main branch) shut down all services (by canceling context) when the first service returns, during shutdown the context can be canceled quickly while app lock is releasing a lock.
3. Wire in tracing from the frontend, so we can see full traces for each page view (see example below) - these are proxied through the backend
4. Fixed: logs were not flushed to OTel upon shutdown, causing some logs to be lost
5. Add tracing to queries performed via Gorm

Example of a trace: page view, showing network requests to the server, and even SQL queries

image

## AI Usage

Mostly my work, Claude was helped on the frontend side and with the tracing proxy


kmendell (commented):
LGTM, I did some quick tests on a local checkout and seems fine. 

Ill leave it up to elias though, as while i get the want for frontend traces, part of me also thinks they arent that useful? but to some it may be  useful i suppose. 

that just adds a decent amount of frontend deps, for what i feel is little benefit for tracing.

ItalyPaleAle:
> that just adds a decent amount of frontend deps

I can maybe make them dynamically imported only if tracing is enabled!

> for what i feel is little benefit for tracing

The reason why I had to integrate tracing in the frontend is to have a single trace root span for each page, so requests created by the same page are correlated. Eventually it could be nice to add other spans that are not HTTP requests too! Or maybe even integrate full frontend observability like RUM!

backend/internal/controller/telemetry_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+type telemetryController struct {
+	targetURL string
+	client    *http.Client
+}
+
+// NewTelemetryController registers an endpoint that receives OTLP/HTTP JSON trace payloads from the browser SPA and forwards them to the OpenTelemetry collector the backend already exports to.
+// Keeping this same-origin means the SPA needs no CORS-enabled, browser-facing collector.
+//
+// The endpoint lives under /internal rather than /api: it is intentionally unauthenticated (traces are produced before/without a session) and not traced itself, but it is rate-limited.
```
stonith404:
```suggestion
```

These comments don't really make sense I would remove them.

backend/internal/job/scheduler.go
```diff
@@ -120,14 +87,86 @@ func (s *Scheduler) RegisterJob(ctx context.Context, name string, def gocron.Job
...
+)
+
+func jobWithObservability(jobName string, job jobFn) jobFn {
+	return func(ctx context.Context) error {
+		// Generate a random job ID
+		jobIDObj, err := uuid.NewRandom()
+		if err != nil {
+			return fmt.Errorf("failed to generate job ID: %w", err)
+		}
+		jobID := jobIDObj.String()
```
stonith404:
```suggestion
		jobID := uuid.NewString()
```

backend/internal/middleware/rate_limit.go
```diff
@@ -45,14 +46,15 @@ type RateLimitPolicy struct {
...
 	return []RateLimitPolicy{
-		{Name: RateLimitAPI, Rate: 1, Per: time.Second, Burst: 100},
-		{Name: RateLimitSignup, Rate: 1, Per: time.Minute, Burst: 10},
-		{Name: RateLimitWebauthnLogin, Rate: 1, Per: 10 * time.Second, Burst: 5},
-		{Name: RateLimitWebauthnReauthenticate, Rate: 1, Per: 10 * time.Second, Burst: 5},
-		{Name: RateLimitOneTimeAccessToken, Rate: 1, Per: 10 * time.Second, Burst: 5},
-		{Name: RateLimitOneTimeAccessEmail, Rate: 1, Per: 10 * time.Minute, Burst: 3},
-		{Name: RateLimitSendEmailVerification, Rate: 1, Per: 10 * time.Minute, Burst: 3},
-		{Name: RateLimitVerifyEmail, Rate: 1, Per: 10 * time.Second, Burst: 5},
+		{Name: RateLimitAPI, Rate: 100, Per: time.Second},
```
stonith404, line 49: Tbh. I like bursts more than constant rate limiting. Did you encounter any issues because you did this change?

backend/internal/controller/telemetry_controller.go
```diff
@@ -0,0 +1,86 @@
+package controller
```
stonith404: I think we should put the telemetry controller into the `tracing` package, as we are moving away from the layer based module structure.

backend/internal/middleware/rate_limit.go
ItalyPaleAle: The code before my last PR was using constant rate limits, it was my mistake when I ported it over to Francis to use burst.

I think if we want to use bursts, it's fine, but I wouldn't set the standard rate limiter to just 1 per period however, that seems a bit too low...

backend/internal/job/scheduler.go
ItalyPaleAle: No, they are different. `uuid.NewString()` could (in theory) panic.

backend/internal/middleware/rate_limit.go
kmendell: Wouldn't a panic be extremely rare? As its usually something else underlying that would cause the panic?

ItalyPaleAle: @kmendell DId you mean to comment on this? https://github.com/pocket-id/pocket-id/pull/1564#discussion_r3525076652

Yes, a panic there would be very rare - technically impossible in modern Linux kernels (the random reader returns an error only if the kernel doesn't have enough entropy, which in kernel 5.6 can't happen anymore IIRC)

backend/internal/job/scheduler.go
stonith404: Like you mentioned it's almost impossible that this would panice, and if it would panic the machine would probably have other issues. I'll leave the decision up to you.

backend/internal/middleware/rate_limit.go
stonith404: @ItalyPaleAle We were using [`x/time/rate`](https://pkg.go.dev/golang.org/x/time/rate) before which uses the token bucket strategy which supports bursts.

I just prefer a strategy that support bursts because automations might send a lot of requests during a short time period. But yeah I agree, one per period is a bit slow for most of the endpoints.

ItalyPaleAle: I can change the values, Any preference?

Something like this?

```go
{Name: RateLimitSignup, Rate: 2, Per: time.Minute, Burst: 10},
{Name: RateLimitWebauthnLogin, Rate: 1, Per: 10 * time.Second, Burst: 5},
{Name: RateLimitWebauthnReauthenticate, Rate: 1, Per: 10 * time.Second, Burst: 5},
{Name: RateLimitOneTimeAccessToken, Rate: 1, Per: 10 * time.Second, Burst: 5},
{Name: RateLimitOneTimeAccessEmail, Rate: 2, Per: 10 * time.Minute, Burst: 1},
{Name: RateLimitSendEmailVerification, Rate: 2, Per: 10 * time.Minute, Burst: 1},
{Name: RateLimitVerifyEmail, Rate: 1, Per: 10 * time.Second, Burst: 5},
{Name: RateLimitInternal, Rate: 20, Per: time.Second, Burst: 20},
```

backend/internal/job/scheduler.go
ItalyPaleAle: Updated

backend/internal/middleware/rate_limit.go
stonith404: Yes looks good, thanks. I'm thinking about increasing the limits for the webauthn login a bit, since this is almost impossible to brute force anyways.

What do you think about?
```
{Name: RateLimitWebauthnLogin, Rate: 1, Per: 5 * time.Second, Burst: 10},
```

One Time Access email is also a bit harsh in my option, I would recommend something like this:
```
{Name: RateLimitOneTimeAccessEmail, Rate: 2, Per: 10 * time.Minute, Burst: 5},
```

ItalyPaleAle: Done!

stonith404 (approved):

PR #1565, opened 2026-07-05
chore: use go-kit for sending emails

ItalyPaleAle:
## What does this PR do?

go-kit is a dependency that was pulled in when we added Francis, and it includes a variety of tools already, including an `emailer` package.

This PR changes the hand-rolled SMTP emailer to use go-kit's, which has equivalent capabilities (including LOGIN command) and extensive testing. Tested and confirmed working in my environment.

With this change, it will be possible to also add support (later) for other emailers that go-kit supports, including AWS SES and SendGrid, which use direct API calls rather than SMTP (in the case of SES, for example, using SMTP is not simple as it requires running Python scripts to convert an access key to a SMTP password)

## AI Usage

Hand-rolled

### DIFF
--- a/backend/internal/service/email_service.go
+++ b/backend/internal/service/email_service.go
@@ -1,24 +1,16 @@
 package service
 
 import (
-	"bytes"
 	"context"
-	"crypto/tls"
 	"errors"
 	"fmt"
 	htemplate "html/template"
-	"io"
-	"mime/multipart"
-	"mime/quotedprintable"
-	"net/textproto"
-	"os"
+	"net"
+	"net/url"
 	"strings"
 	ttemplate "text/template"
-	"time"
 
-	"github.com/emersion/go-sasl"
-	"github.com/emersion/go-smtp"
-	"github.com/google/uuid"
+	"github.com/italypaleale/go-kit/emailer"
 	"gorm.io/gorm"
 
 	"github.com/pocket-id/pocket-id/backend/internal/common"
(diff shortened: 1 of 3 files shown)


kmendell:
Would it be worth forking go-kit into the organization that way we have our own in the evnt we need updates or anything made/added? (Thinking outloud)

ItalyPaleAle:
@kmendell good point. Happy to make you and @stonith404 a maintainer of that and go-sql-utils

kmendell:
Yeah that works as well, not that its needed, but thinking in the event something does need to be done. I want to pull and test the changes but overall changes look good to me just want to confirm nothing weird happens with them.

backend/internal/service/email_service.go
```diff
@@ -82,224 +74,103 @@ func SendEmail[V any](ctx context.Context, srv *EmailService, toEmail email.Addr
...
-	if err == nil {
-		if err := client.Hello(hostname); err != nil {
-			return err
-		}
+// smtpConnString builds the SMTP connection string that go-kit's emailer expects:
+// smtp://<username>:<password>@<host>:<port>?fromAddress=<address>&fromName=<name>&tls=<none|starttls|tls>&insecureSkipVerify=<true|false>
+func smtpConnString(dbConfig *model.AppConfig) (string, error) {
+	host := dbConfig.SmtpHost.Value
+	if host == "" {
+		return "", errors.New("SMTP host is not configured")
```
kmendell, line 122: Nit - Should this be an error defined in internal/common/errors.go?

stonith404 (approved):

PR #1575, opened 2026-07-08
fix: /authorize endpoint crashes when list of scopes is empty

ItalyPaleAle:
## What does this PR do?

Fixes an issue we experienced using Pocket ID with Vercel Connect, where often the list of scopes is empty, causing the frontend to crash

## AI Usage

Hand-rolled code fix. Tests written by a new AI model


kmendell (approved):
LGTM!

PR #1579, opened 2026-07-10
refactor: manage instance ID in the KV table

ItalyPaleAle:
## What does this PR do?

This is part #1549 - Split-off from a larger PR I've been working on to move AppConfig to an actor.

The instance ID is currently stored in the app config database table, as an "internal value". Given the way we use the instance ID, this should not be added to an actor, since it's a value we need early in the bootstrap phase, and it's a read-only value that once set doesn't change.

Migrating the instance ID to the KV table.

- Existing clusters have the key migrated from the app config to the KV table automatically with a migration step (a SQL file)
- New clusters continue to generate the instance ID in the KV value with an atomic query, which is safe to be executed by multiple processes at the same time since it's atomic.

Concurrently, this PR updates francis to the latest beta, which contained some API changes for how actors are registered.

## AI Usage

Most of this was hand-written. New tests were contributed by Claude.


backend/Makefile
kmendell, line 1: Was this an intentional add?

kmendell (commented):
Besides that one comment, Everything looks good to me, all seems to be working locally as well. 

In the event i missed something, letting elias be the final call on this.

backend/Makefile
ItalyPaleAle: yes definitely intentional - having to remember to add `-tags unit,frontend_exclude` when running tests was a hassle rather than just `make test`. Also `make lint test` is convenient.

stonith404 (approved):

PR #1593, opened 2026-07-16
feat: restrict signup invite links to a specific email domain

ItalyPaleAle:
## What does this PR do?

When creating a signup token, add optionally the ability to restrict the domain name of users' emails for accounts created with that signup token

## AI Usage

Partly written with Opus, supervised and reviewed by me


stonith404:
The use case you described seems very niche, because of that I'm closing this. If someone really thinks this makes sense to implement, feel free to create a feature request.

PR #1402, opened 2026-03-29
feat: add QR-Login / Device Authorization Grant (RFC 8628)

Norrodar:
## Summary

This feature implements the **RFC 8628 Device Authorization Grant** as QR-Login for Pocket ID. Users can sign in on devices without keyboard or WebAuthn support (Smart TVs, consoles, set-top boxes) by scanning a QR code or entering an 8-digit code on a second device.

## Motivation

Pocket ID relies on WebAuthn/Passkeys as primary authentication. Devices like Smart TVs (e.g., LG WebOS) don't support WebAuthn. Without this feature, login on such devices is impossible. RFC 8628 Device Authorization Grant is the standard OAuth flow for exactly this use case.

### Backend

**New Endpoints:**
| Endpoint | Method | Auth | Description |
|----------|--------|------|-------------|
| `/oidc/device/authorize` | POST | Client-Auth | Creates Device Code + User Code |
| `/oidc/device/verify` | POST | User-Auth | Verifies User Code, authorizes device |
| `/oidc/device/info` | GET | User-Auth | Returns scope/client info for a User Code |
| `/oidc/device/exchange-session` | POST | None | Polls for authorization, sets session cookie |
| `/oidc/token` | POST | Client-Auth | Standard token endpoint with `grant_type=urn:ietf:params:oauth:grant-type:device_code` |

**Data Model:**
- `oidc_device_codes` table with `device_code` (UNIQUE), `user_code` (UNIQUE), `scope`, `expires_at`, `is_authorized`, `user_id`, `client_id`, `last_polled_at`, `interval_seconds`
- Device codes: 32 alphanumeric characters (190-bit entropy)
- User codes: 8 characters from `ABCDEFGHJKMNPQRSTUVWXYZ23456789` (no ambiguous characters)

**Self-Login Client:**
- Internal OAuth client `pocket-id-self-login` is auto-created on startup
- Used exclusively for the QR-Login flow on the login page itself
- `/exchange-session` endpoint restricted to this client only

**Security:**
- `FOR UPDATE` locks on all device code operations (race condition protection)
- Rate limiting via `slow_down` with interval increase (+5s per violation, max 60s)
- Device codes deleted immediately after single use
- Expired codes cleaned up by scheduled job
- `Cache-Control: no-store` + `Pragma: no-cache` on all token responses
- Already authorized codes cannot be re-verified or queried
- Admin toggle `qrLoginEnabled` blocks all device flow endpoints

[...]

PR #1594, opened 2026-07-16
feat: add qr code alternative sign in method

### LINKED ISSUE #1146: 🚀 Feature: Device Authorization Grant with QR code
### Feature description

I will preface this issue with the facts that I am a user of Pocket ID, by no means a developer, and not even a novice when it comes to OIDC / OAuth and the specifications on which they are developed, so my request may not even be possible within the specification. However, I _think_ what I am trying to describe is [RFC 8628](https://datatracker.ietf.org/doc/html/rfc8628), with specific emphasis on the "_input constrained_" element of the standard.

The feature has been posed in a comment in a previous issue ([#267](https://github.com/pocket-id/pocket-id/issues/267#issuecomment-3033113211)), however the issue was closed before the comment was made and I don't feel that the use case was fully understood before it was closed, hence raising it here.

1. User opens the client (e.g. Home Assistant, Jellyfin) on a device which doesn't support passkeys (e.g. smart TV, in-vehicle web-browser). User is forwarded to Pocket ID
2. User selects _Sign in with QR code_ option (which is supplement to the existing _Sign in with login code_ option)
5. User scans the QR code on a device which supports passkeys, e.g. their smartphone, which forwards the user to Pocket ID with a short-lived code
6. User authenticates on their compatible device and client (on incompatible device) is successfully logged in

### Pitch

I have two specific (albeit edge) cases where this functionality would improve quality of life - I concede that this functionality is already possible using the existing short-code workflow, but implementing this feature would make the expereince more fluid in most cases and improve user experience.

1. **Jellyfin on HDMI TV dongle**: when travelling I access my media remotely through Jellyfin. Currently I use the login code method but entering the code into the login screen can be awkward and prone to mistakes. This is particularly annoying on hotel WiFi where inter-device connectivity is blocked, I usually have to enter the code using the remote and the non-qwerty, square on-screen keyboard. 
2. **Home Assistant on in-vehicle display**: I sometimes access Home Assistant from my in-vehicle display where I have a specific dashboard setup for things like garage door control. Similarly to the previous example, I currently use the login code method which is possible, but isn't a smooth flow.

I do have a few other examples but they make me look even more lazy than I am making myself look.

The current process requires the user to make the following steps:

1. Open Pocket ID on a compatible device
2. Authenticate on compatible device
3. Create login code
4. Navigate to client on incompatible device
5. On incompatible device, select _Sign in with login code_ on client
6. Paste login code and submit

The proposed process would require the user to take the following steps:

1. Navigate to client on incompatible device which forwards to Pocket ID
2. On incompatible device, select _Sign in with QR code_ on client
3. Scan QR code on compatible device, which opens Pocket ID authentication page
4. Authenticate on compatible device

The proposed alternative is 2 fewer steps (slightly subjective how many, but I'd argue its objectively fewer steps), but regardless of this the interaction required by the user is far more fluid (e.g. no manual code entries).

This example assumes the user typically uses a mobile device as their compatible device and is therefore able to scan the QR code. The proposal is that this feature compliments the existing _Sign in with login code_ feature - not replaces it.

stonith404:
## What does this PR do?
This PR adds a third alternative sign in method and closes #1146. This login method generates a QR Code on the device the user wants to sign in and the QR Code can then be scanned from another device to approve. Thanks to @Norrodar with his PR (#1402) which has inspired this PR.

## Screenshots
Screenshot 2026-07-16 at 22 50 26@2x
Screenshot 2026-07-16 at 22 46 09@2x
Screenshot 2026-07-16 at 22 50 11@2x

## AI Usage
Used Codex for bootstrapping the new backend module. Frontend was done completely by hand. The backend module has been optimized by hand and of course fully reviewed.

## Contributing Guidelines


stonith404:
@coderabbitai review

backend/internal/devicelogin/module.go
```diff
@@ -0,0 +1,59 @@
...
+	return &Module{
+		service: service,
+		handler: newHandler(service, deps.BaseURL, deps.AppConfig),
+	}
+}
+
+// RegisterRoutes mounts the public exchange and authenticated verification endpoints
+func (m *Module) RegisterRoutes(apiGroup *gin.RouterGroup, browserAuth, createRateLimit, verificationRateLimit gin.HandlerFunc) {
+	apiGroup.POST("/device-login/requests", createRateLimit, m.handler.createRequest)
+	apiGroup.POST("/device-login/requests/:id/exchange", m.handler.exchangeRequest)
```
Norrodar: Only one without rate limit?
Brute force isn't a realistic concern here so this is mostly about consistency with the other endpoints — and about not letting a misbehaving client hammer the endpoint with unthrottled requests. Since the endpoint is polled, the policy would need to be more permissive than the other two, something like Rate: 1, Per: 2 * time.Second, Burst: 10 to stay comfortably above the 3-second polling interval.

Norrodar (approved):
Looks really good to me. clean module, and the device-token cookie binding plus the mandatory reauthentication on approve are nice touches. I left one small **question about a missing rate limit** in the exchange endpoint, but nothing blocking.

backend/internal/devicelogin/service.go
```diff
@@ -0,0 +1,242 @@
...
+		return Request{}, "", err
+	}
+
+	now := time.Now().Round(time.Second)
+	request := Request{
+		DeviceTokenHash: utils.CreateSha256Hash(deviceToken),
+		Status:          RequestStatusPending,
+		ExpiresAt:       datatype.DateTime(now.Add(RequestDuration)),
+		UserAgent:       userAgent,
+		IpAddress:       ipAddress,
```
ItalyPaleAle: Why storing the IP address? If it's for audit logs, IMHO we should emit 2 separate lines: when the request is created and when it's completed.

backend/internal/devicelogin/service.go
```diff
@@ -0,0 +1,242 @@
...
+	case "deny":
+		request.Status = RequestStatusDenied
+	default:
+		return fmt.Errorf("unsupported device login decision %q", decision)
+	}
+
+	result := tx.
+		WithContext(ctx).
+		Model(&Request{}).
+		Where("id = ? AND status = ? AND expires_at > ?", request.ID, RequestStatusPending, datatype.DateTime(time.Now())).
```
ItalyPaleAle: For correctness, and for readiness for high availability, suggest storing times that are computed in the database rather than the app layer. So expires_at would use the server's time (e.g. current_timestamp). Reason is that with multiple replicas of Pocket ID, on different nodes, system clocks may be different (clock skew)

ItalyPaleAle (commented):
A few nits, overall LGTM.

One overarching comment: why do we need to poll for the login approval? Why can't we keep a long-lived HTTP request open and just wait for a signal on that one?
(great use case for an actor btw!)

Norrodar:
> Why can't we keep a long-lived HTTP request open and just wait for a signal on that one? (great use case for an actor btw!)

The polling needs to stay as a fallback in any case. The long-lived connection is awesome and feels modern, but I'm a little afraid it adds complexity on the network side, e.g. reverse proxies or stupid Smart TVs. And if long polling fails, the client should fall back to short polling anyway. So it would be a very nice-to-have on top. 🚀

backend/internal/devicelogin/service.go
Norrodar: The IP is stored for the review screen on approval (anti-phishing design): the approver sees the requesting device and its IP before confirming. Since the request is created on one device and reviewed on another one later, it has to be persisted.

Setting | Value
-- | --
Code | PKQM - 3WT7
Device | LG Web Browser on webOS TV
IP Address | 192.168.178.42

I guess a DNS lookup would take too long and it would be a point of failure in Docker/reverse-proxy setups anyway. The user agent (e.g. "LG Web...") is already sufficient in most cases.

grafik
(from stonith's example above)

ItalyPaleAle: ACK, I didn't think of the approval (I agree a reverse DNS lookup is unnecessary)

ItalyPaleAle:
> > Why can't we keep a long-lived HTTP request open and just wait for a signal on that one? (great use case for an actor btw!)
> 
> The polling needs to stay as a fallback in any case. The long-lived connection is awesome and feels modern, but I'm a little afraid it adds complexity on the network side, e.g. reverse proxies or stupid Smart TVs. And if long polling fails, the client should fall back to short polling anyway. So it would be a very nice-to-have on top. 🚀

Yes, even with long polling we still need to retry the poll if the request is abruptly closed.

The main difference is the in-memory coordination: with short polling each poll hits the database, while with long polling you keep the state in-memory (or in our case, in an actor) and can immediately react to state changes.

stonith404:
@ItalyPaleAle Could you review my recent changes? I tried to implement the long polling with an actor.

backend/internal/utils/string_util.go
```diff
@@ -23,13 +23,23 @@ func GenerateRandomUnambiguousString(length int) (string, error) {
 	return GenerateRandomString(length, charset)
 }
 
+// GenerateRandomUppercaseUnambiguousString generates a random uppercase string of the given length using unambiguous characters
+func GenerateRandomUppercaseUnambiguousString(length int) (string, error) {
```
ItalyPaleAle, line 27: (Non-blocker) Looks like this is omitting 1 (ambiguous with I) and O (ambiguous with 0)

What if here (and in GenerateRandomUnambiguousString) we did include 1 and 0, and then when we receive the code, if we detect I we convert it to 1, and O to 0?

The reason I'm suggesting this is to have 2 more characters in the charset, which offers (a bit) more entropy to the generated codes

backend/internal/devicelogin/service.go
```diff
@@ -0,0 +1,291 @@
...
+	case RequestStatusDenied:
+		return state.Status, &common.DeviceLoginDeniedError{}
+	default:
+		return "", &common.DeviceLoginRequestInvalidOrExpiredError{}
+	}
+}
+
+func (s *Service) actorIDForCode(code string) (string, error) {
+	code = strings.ToUpper(strings.TrimSpace(code))
+	derived, err := cryptoutils.DeriveKey(s.actorIDKey, "pocketid/device-login-request/"+code)
```
ItalyPaleAle: Why do we need a HMAC here? Couldn't we use just the code as actor ID?

backend/internal/devicelogin/actor.go
```diff
@@ -0,0 +1,396 @@
...
+		}
+		return a.completeApprovedExchange(ctx, state, input)
+	default:
+		return requestActorResult{Code: requestActorResultInvalid}, nil
+	}
+}
+
+func (a *requestActor) completeApprovedExchange(ctx context.Context, state requestActorState, input requestActorExchangeInput) (requestActorResult, error) {
+	var user model.User
+	err := a.db.WithContext(ctx).First(&user, "id = ?", state.UserID).Error
```
ItalyPaleAle: I would recommend keeping DB operations outside of the actor. Let the actor only own the state and then make the service perform DB invocations. Things can get a bit "spaghetti" otherwise

ItalyPaleAle:
@stonith404 This looks nice but I see that the service is still polling the actor. I'm thinking of adding a new primitive in francis (like a new builtin actor) that could be use to receive notifications on a specific event, let's call it "signal" for example.

The idea is:

- One ore more pages open a long-lived request to the "signal" actor
- The requests remain pending
- Another request can tell the actor to send a signal to all listeners
- All listeners (pending requests) would receive the signal right away
- Subsequent requests would receive the response immediately without waiting

This would allow avoding polling and making it so clients (browsers) can just keep a long-running HTTP request that is notified as soon as the login is approved.

If the request times out, clients can re-connect and they'll receive the same signal.

stonith404:
@ItalyPaleAle Okay yeah I agree that would be a cleaner solution. Although this would be cleaner, don't you think this would increase the complexity unnecessarely? Because in terms of performance I think the current approach would be sufficient, because we are polling from the actor state in the memory, not the database.

backend/internal/utils/string_util.go
stonith404: Good idea 👍

ItalyPaleAle:
The bigger advantage I see is not so much in terms of resources used but how quick responses are received by clients. With polling you still have to wait an average of (polling interval / 2), but with a signal it would be instant

stonith404:
Thanks for the PR feedback, I've made the changes.

> The bigger advantage I see is not so much in terms of resources used but how quick responses are received by clients. With polling you still have to wait an average of (polling interval / 2), but with a signal it would be instant

I mean that would be 125ms that could be saved on average, I don't really see a huge UX gain here 😅

kmendell:
Im going to put my two cents in here, I did my do-digience tests locally and it all looks good to me as far as functionality.

ItalyPaleAle:
I think it's fine to merge for now.

I spoke to @stonith404 yesterday about the long polling. When proper HA mode is supported and there are multiple replicas, requests to the actor could go over the network so polling every 250ms could be too much.

I am working on a new primitive in Francis to allow signals like described above but will take another couple of days.

We can always revise this feature later!

Norrodar:
Non-blocker:
consume deletes the state before the user lookup and token generation run so "restoreRequest" has to put it back on failure. Would it work to do the lookup and token generation first and consume last? Than the state stays untouched until the end and the restore path wouldn't be needed.

PR #1600, opened 2026-07-17
fix: sync group membership for users added at creation time

ItalyPaleAle:
## What does this PR do?

This fixes a bug with SCIM sync for users who are created with an invite link that includes an automatic group.

The fix involves setting the `updated_at` explicitly. Otherwise, users would be provisioned in the SCIM-connected service without groups in some cases.

## AI Usage

Co-authored with Opus


kmendell (approved):
LGTM, I assume this was the issue I saw messages about in discord? :P

ItalyPaleAle:
> LGTM, I assume this was the issue I saw messages about in discord? :P

indeed! we encountered this bug yesterday. thanks!

PR #1604, opened 2026-07-18
feat: use actors for db configuration

ItalyPaleAle:
## What does this PR do?

This is part of #1549 

It migrates app configuration to an actor, which always has the up-to-date view of the app config in the cluster.

## AI Usage

Critical parts have been hand-rolled (ie. most of the "appconfig" package, including the actor itself). AI assistance used to complete the migration throughout the codebase


backend/internal/appconfig/appconfig_actor.go
```diff
@@ -0,0 +1,200 @@
...
+	// Load the actor state
+	ctx, cancel := context.WithTimeout(parentCtx, 10*time.Second)
+	defer cancel()
+	state, err := a.client.GetState(ctx)
+	if err != nil {
+		return fmt.Errorf("error retrieving actor state: %w", err)
+	}
+
+	// If we already have a state, nothing else to do
+	if state != nil {
```
stonith404, line 63: Just a question; How can this happen? Does bootstrap run on every host even though it's a singleton?

backend/internal/appconfig/model.go
```diff
@@ -0,0 +1,316 @@
...
+		LdapAttributeGroupName:             "",
+		LdapAdminGroupName:                 "",
+		LdapSoftDeleteUsers:                "true",
+	}
+}
+
+// fromLegacyConfig builds an appConfigModel from a legacy config map
+// The map's keys correspond to the "json" tags on appConfigModel, and all values are strings that are cast to each field's type
+// Keys that are missing (or have an empty value) retain the default value
+func fromLegacyConfig(legacyCfg map[string]string) (*AppConfigModel, error) {
```
stonith404: Can we move the migration into a new module? A single file inside a `migration` module that runs on startup would be cleaner IMO. I don't really like to mix migration logic into business logic. Currently we have migration logic in three different files.

backend/internal/service/audit_log_service.go
```diff
@@ -64,7 +63,7 @@ func (s *AuditLogService) Create(ctx context.Context, event model.AuditLogEvent,
 }
 
 // CreateNewSignInWithEmail creates a new audit log entry in the database and sends an email if the device hasn't been used before
-func (s *AuditLogService) CreateNewSignInWithEmail(ctx context.Context, ipAddress, userAgent, userID string, tx *gorm.DB) model.AuditLog {
+func (s *AuditLogService) CreateNewSignInWithEmail(ctx context.Context, ipAddress, userAgent, userID string, tx *gorm.DB, dbConfig *appconfig.AppConfigModel) model.AuditLog {
```
stonith404: I would keep the dbConfig as a dependency of the service, instead of the method? Or at least pass `EmailLoginNotificationEnabled` directly to the method instead of the whole struct.

stonith404:
Thanks for the PR, it looks good to me but one question that spans over the whole PR. 

You introduced a new way to get config variables from the context. Why storing the app config in the context instead of passing the actor as a dependency to the handlers and services? In my opinion this is the same anti pattern as [storing transactions in the context](https://www.reddit.com/r/golang/comments/1awp5af/is_passing_database_transactions_as_via_context/).

ItalyPaleAle:
> You introduced a new way to get config variables from the context. Why storing the app config in the context instead of passing the actor as a dependency to the handlers and services? In my opinion this is the same anti pattern as [storing transactions in the context](https://www.reddit.com/r/golang/comments/1awp5af/is_passing_database_transactions_as_via_context/).

The reason I went this route was because fetching the config from the actor isn't free but it requires a call. When there's more than one replica the call _may_ go over the network (depending on placement). Additionally there's no guarantee that with multiple invocations the response is the same, as the config may be mutated meanwhile.

I have found situations where a request chain could request the config multiple times, in various methods throughout the call stack. I thought storing it in the context would be a convenient method so it's fetched in a middleware and then it's available for all methods in the call chain.

If you prefer, there is an alternative that's more explicit but a bit more verbose: each method in the service package (or one of its dependencies) would require passing the config object explicitly. Then the method in the controller is the one that explicitly fetches it from the actor.

backend/internal/appconfig/appconfig_actor.go
ItalyPaleAle: Every host calls bootstrap when it starts up. This is because there's no way to know if the singleton had been bootstrapped already:

https://francis.italypaleale.me/docs/singleton-actors/#the-actorbootstrapper-interface

ItalyPaleAle:
> If you prefer, there is an alternative that's more explicit but a bit more verbose: each method in the service package (or one of its dependencies) would require passing the config object explicitly. Then the method in the controller is the one that explicitly fetches it from the actor.

As discussed on discord, this is now done (I tasked Claude to do it, and upon review it looks correct)

backend/internal/service/audit_log_service.go
ItalyPaleAle: I think passing EmailLoginNotificationEnabled directly makes sense

backend/internal/appconfig/model.go
ItalyPaleAle: I moved it to a “migrations.go” file in this package, it seemed a cleaner solution than a migrations package which would cause some awkward dependencies. LMK if this is fine with you

stonith404 (approved):

PR #1611, opened 2026-07-22
feat: migrate one-time and signup tokens to an actor

ItalyPaleAle:
## What does this PR do?

Part of #1549 

Migrates the one-time access tokens and signup tokens to actors. This makes the periodic cleanup logic redundant too, as the actors themselves implement various cleanup logics.

## AI Usage

Co-authored with Claude since I don’t have my laptop with me, but with oversight and manual tweaks


backend/internal/service/e2etest_service.go
```diff
@@ -340,14 +320,8 @@ func (s *TestService) SeedDatabase(baseURL string) error {
...
-		accessToken := model.OneTimeAccessToken{
-			Token:     "one-time-token",
-			ExpiresAt: datatype.DateTime(time.Now().Add(1 * time.Hour)),
-			UserID:    users[0].ID,
-		}
-		if err := tx.Create(&accessToken).Error; err != nil {
-			return err
-		}
+		// One-time access tokens (including "one-time-token") are seeded into the actor state store
+		// after this transaction commits, in seedOneTimeAccessTokens.
```
stonith404:
```suggestion
```

backend/internal/service/one_time_access_actor.go
stonith404, line 1: Can we migrate everything related to one time access tokens to a separate module called `onetimeaccess`?

backend/internal/usersignup/testing_e2etest.go
stonith404, line 1: Can't we move this into the e2etest service? In my opinion it's not very clean to have functionality that is only used for tests in the package itself.

stonith404:
@ItalyPaleAle The sign up tokens are stored in a singleton actor while the one time access tokens are stored in separate actors. 

I guess you are storing sign up tokens in a singleton actor because we need to list them in the UI, is this the only reason? Wouldn't it be cleaner to implement the functionality into Francis to list actors? I'm very new to actors so I might be wrong here but for me storing all sign up tokens in a singleton actor seems like an anti pattern to me.

ItalyPaleAle:
> I guess you are storing sign up tokens in a singleton actor because we need to list them in the UI, is this the only reason? Wouldn't it be cleaner to implement the functionality into Francis to list actors? I'm very new to actors so I might be wrong here but for me storing all sign up tokens in a singleton actor seems like an anti pattern to me.

Yes, you are spot-on, I used a singleton actor because I needed to list individual sign-up tokens. I wouldn't quite say this is anti-pattern, but I agree it's not ideal.

I think an API to list actors (to be precise here, to list actor _state objects_) is possible. The only thing to consider is that this can be implemented in 2 ways:

1. The API returns the list of each actor state and its value. This is a single, atomic operation, but has the potential to return a lot of data. It would have to be used carefully (it's probably ok in this case, but I can see cases where it could be abused). Pagination would probably be required (with pagination there's always the risk of loss of consistency, but that should be acceptable here). to note: with this approach we could only _read_ state and never modify it.
2. The API returns just the list of actors and then we would need to invoke them one-by-one. This is slower (with N actors, there's N+1 invocations) and could cause idle actors to be re-activated just to read the state. It is more flexible, but also less consistent (between the call to list actors and the invocation to each actor, the list may have changed)

I am thinking option 1 would be the way to go, with the caveats listed above...

stonith404:
Thanks for your explanation! Yes, option one sounds really good. For exactly this use-case I wouldn't use pagination provided by Francis, since the list of sign up tokens won't be large and we still want to support sorting. But nevertheless it definitely would make sense to implement pagination into Francis.

How do you want to move forward? Do you think that is a change that is implemented pretty quickly into Francis? Because if not, I would recommend to strip out the signup token changes from this PR and create a follow up PR.

ItalyPaleAle:
@stonith404 your comments should now be addressed!

stonith404 (approved):
Thanks!

PR #1617, opened 2026-07-23
feat: support redirect URLs after passkey setup

### LINKED ISSUE #1553: 🚀 Feature: redirect_uri and cancel_uri query parameters on /signup/add-passkey
### Feature description

I’d like to submit a pull request and add support for optional `redirect_uri` and `cancel_uri` query parameters on `/signup/add-passkey`.

Proposed behavior:
- `/signup/add-passkey?redirect_uri=...` redirects to `redirect_uri` after successful passkey setup.
- `/signup/add-passkey?cancel_uri=...` redirects to `cancel_uri` when the user skips passkey setup.
- Both parameters are optional.
- If either parameter is missing, invalid, or not registered, Pocket ID falls back to `/settings/account`.
- External redirects are only allowed if the URL matches an existing OIDC client callback/redirect URL.
- Matching should use the existing callback URL matching logic, including wildcard support.
- Relative/internal URLs should not become an open redirect path through this feature.

Security considerations:
The feature must not allow arbitrary redirects. The supplied URL should only be used if it matches one of the callback URLs configured for an existing OIDC client/application. Unknown external URLs, malformed URLs, relative URLs, and unsafe schemes should fall back to `/settings/account` instead.

Implementation outline:
- Add a backend `helper/service` method that checks whether a requested callback URL is registered on any existing OIDC client.
- Reuse the existing callback URL matching logic, e.g. `utils.GetCallbackURLFromList`, so wildcard behavior stays consistent with OIDC redirect handling.
- Expose a small authenticated endpoint that returns whether a requested URL is registered, without exposing all OIDC clients to non-admin users.
- Update `frontend/src/routes/signup/add-passkey/+page.svelte` to use `redirect_uri` after successful `finishRegistration()` and `cancel_uri` when the user confirms skip
- Validate either URL through the backend helper endpoint, redirect only if everything is right, otherwise fall back to `/settings/account`
- Add backend unit tests for the new validation behavior and rely on the existing callback matcher tests for matching details.

**Please let me know if I can go ahead and submit a PR.**

### Pitch

Pocket ID can already be integrated well with external applications through OIDC and its API. One missing piece is returning users back to an external account-management flow after passkey setup.

For example, an organization may use a separate portal where users manage their profile, onboarding, or application access. That portal can create or invite a user through Pocket ID, then send them to `/signup/add-passkey` to complete passkey setup. Today, after adding the passkey, the user is always sent to /settings/account, which interrupts the external flow and forces the integrating application to somehow recover that behavior.

With the proposed changes, the user can be redirected to the company portal that shows a success screen and completes registration.

I believe this fits very well with Pocket ID’s API-first / integration-friendly model.

tobidwest:
## What does this PR do?

Closes #1553.

This PR adds support for optional `redirect_uri` and `cancel_uri` query parameters on `/signup/add-passkey`.

* `redirect_uri` is used after successful passkey registration.
* `cancel_uri` is used when the user confirms that they want to skip passkey setup.
* Both parameters are optional.
* Supplied URLs are validated against the callback URLs registered for existing OIDC clients.
* The existing callback URL matching logic is reused, including wildcard support.
* Invalid, malformed, unregistered, relative, or otherwise unsafe URLs fall back to `/settings/account`.

The PR also adds an authenticated backend endpoint for checking whether a requested redirect URL is registered without exposing the list of OIDC clients to non-admin users.

Backend tests cover the new redirect validation behavior.

## Screenshots

No visual UI changes.

## AI Usage

I used an LLM to help review and refine parts of the implementation and the accompanying tests. I manually reviewed, adjusted, and tested the code and understand the submitted changes.

## Contributing Guidelines

* [x] If I'm implementing a new feature, I have opened an issue first, or commented on an existing one, to discuss the implementation details.
* [x] I have read the [AI Usage Guidelines](https://github.com/pocket-id/pocket-id/blob/main/CONTRIBUTING.md#ai-usage-policy)

### DIFF
--- a/backend/internal/controller/oidc_controller.go
+++ b/backend/internal/controller/oidc_controller.go
@@ -38,6 +38,8 @@ func NewOidcController(group *gin.RouterGroup, authMiddleware *middleware.AuthMi
 	group.DELETE("/oidc/clients/:id/logo", authMiddleware.Add(), oc.deleteClientLogoHandler)
 	group.POST("/oidc/clients/:id/logo", authMiddleware.Add(), fileSizeLimitMiddleware.Add(2<<20), oc.updateClientLogoHandler)
 
+	group.GET("/oidc/redirect-uri/registered", authMiddleware.WithAdminNotRequired().Add(), oc.getRegisteredRedirectURIHandler)
+
 	group.GET("/oidc/clients/:id/preview/:userId", authMiddleware.Add(), oc.getClientPreviewHandler)
 
 	group.GET("/oidc/users/me/authorized-clients", authMiddleware.WithAdminNotRequired().Add(), oc.listOwnAuthorizedClientsHandler)
@@ -149,6 +151,19 @@ func (oc *OidcController) listClientsHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, dto.Paginated[dto.OidcClientWithAllowedGroupsCountDto]{
 		Data:       clientsDto,
 		Pagination: pagination,
+	})
+}
+
+func (oc *OidcController) getRegisteredRedirectURIHandler(c *gin.Context) {
+	redirectURI, err := oc.oidcService.GetRegisteredCallbackURL(c.Request.Context(), c.Query("redirect_uri"))
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	c.JSON(http.StatusOK, dto.OidcRegisteredCallbackURLDto{
+		Registered:  redirectURI != "",
+		RedirectURI: redirectURI,
 	})
 }
 
--- a/backend/internal/dto/oidc_dto.go
+++ b/backend/internal/dto/oidc_dto.go
@@ -73,6 +73,11 @@ type OidcUpdateAllowedUserGroupsDto struct {
 	UserGroupIDs []string `json:"userGroupIds" binding:"required"`
 }
 
+type OidcRegisteredCallbackURLDto struct {
+	Registered  bool   `json:"registered"`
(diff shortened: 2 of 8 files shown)


ItalyPaleAle (changes requested):
While I will let @stonith404 and @kmendell comment on whether we want this or not, from a concrete perspective I don’t think we should abuse of the redirect_uri’s infrastructure for this.

The redirect_uri has a precise meaning in the OAuth2 world and it indicates the endpoint where users are sent to after they authorized the app, and the endpoint receives the authorization code too (which can be exchanged for an access token). Most apps use a very specific endpoint like `/oauth2/callback` for this.

Here we are talking about redirecting users after they added a passkey (or canceled it). We should configure a separate allowlist for this, that doesn’t rely on the redirect_uri allowlist. 

The risk is that users will add to the redirect_uri allowlist endpoints that have nothing to do with OAuth2, but that can then receive authorization codes, and this is a security risk.

tobidwest:
@ItalyPaleAle Thank you for your response and good catch! I implemented it this way to keep it simple and maintainable but totally see the point. Would you want such an allowlist to be a frontend setting or rather an env variable?

I'm looking forward to a quick reply from @stonith404 or @kmendell on whether you want the feature or not before changing the behavior.

ItalyPaleAle:
> Would you want such an allowlist to be a frontend setting or rather an env variable?

It would have to be a frontend setting because it’s something you’d set per app (client). I think it’s fine to re-use the same algorithm for validating URLs, supporting wildcards, etc. But it should be a separate field (and I’m not sure if the cancel URL should be a separate one too - probably not necessary but I can be convinced otherwise)

PR #1618, opened 2026-07-23
feat: add custom claims conflict strategy

### LINKED ISSUE #1273: 🐛 Bug Report: Custom Claims with same key overrides previous defined values

kiaraly:
## What does this PR do?
More a poc and something to get my last missing feature started.
Adds an optional field to #1409 that specifies what happens when two custom group claims have the same key.
The options are: default (current behavior), first (when multiple keys are in group custom claims, does a natural sort and selects the first), last (similar but picks last), collect (merges all values into a list).

fixes #1273

Example:
User is in two groups which each define the custom claim `quota` with values 100GB and 50GB respectively. With this PR you can change it so the largest/smallest value gets selected instead of whichever group gets processed first.

## Screenshots
grafik

## AI Usage
Gemini wrote the UI part as I'm entirely clueless about doing frontend stuff. I'd need some help doing the UI properly.

## Contributing Guidelines


backend/go.mod
```diff
@@ -156,96 +132,61 @@ require (
 	github.com/lestrrat-go/httpcc v1.0.1 // indirect
 	github.com/lestrrat-go/option/v2 v2.0.0 // indirect
 	github.com/lib/pq v1.12.3 // indirect
+	github.com/maruel/natural v1.3.0 // indirect
```
ItalyPaleAle: Pls run go mod tidy

ItalyPaleAle (changes requested):
My .02: I think this is a bit overly complex also due to the fact that many of the options are about “first” and “last” and this would set the expectation that the order of groups is stable - while it is not at present, and we cannot guarantee that groups will be loaded from the database in a specific order all the time.

What if we simplified this by making it so if users are in groups that set the same value, we should just collect them in an array? (With no promise that the order of the values will be stable) This way we would not need new configs in the UI.

kiaraly:
> My .02: I think this is a bit overly complex also due to the fact that many of the options are about “first” and “last” and this would set the expectation that the order of groups is stable - while it is not at present, and we cannot guarantee that groups will be loaded from the database in a specific order all the time

This PR collects all group custom claims in a list and then does its selection logic, so it is independent of load order of the DB.

> What if we simplified this by making it so if users are in groups that set the same value, we should just collect them in an array? (With no promise that the order of the values will be stable) This way we would not need new configs in the UI.

While admittedly I haven't tried this I think most apps expect just one value for claims like quota. I'm not sure what happens when you give them a list instead. In either case that would move the burden of selecting the right value to the downstream app which might not expose a way to do this (immich comes to mind).

Issue #1574, opened 2026-07-08
🚀 Feature: Declaritive Configuration of User IDs and Client Secrets


schnz:
### Feature description

Widen the API to accept custom user IDs and client secrets:
* `POST /api/users`: Accept an optional user id (UUID v4) in `dto.UserCreateDto` or some sort of alternative field which is propagated as `sub` claim for OAuth clients.
* `POST /api/oidc/clients/{id}/secret`: Accept an optional secret that is used instead of a randomly generated one

Ideally, the secret would also be specifiable in the `POST /api/oidc/clients` endpoint. But from what I understand, the usual flow is to create the client first, then create the secret separately.

### Pitch

TL;DR: This allows for central management/deployment of users and clients, e.g., via Terraform.

#### Related Issues
#864 already introduced user-supplied OAuth client IDs. The suggestion for programmatic client secrets was discussed in #83 and #1166 already. Both of which were closed prematurely IMO.

#83 had a legitimate use-case, which was to the declarative provision of OAuth clients with backed-in client IDs and secrets such as the [ownCloud clients](https://doc.owncloud.com/server/latest/admin_manual/configuration/user/oidc/oidc.html#client-ids-secrets-and-redirect-uris). Back then, the issue was closed and the proposed solution was to change the secret directly in the database. This solution is suboptimal at best. I couldn't spot an explanation as of why this scenario shouldn't be handled by the API.

#1166 was closed as "completed" as part of #1229, which only addressed the static API key (which is great!) but _not_ the declarative client secret. My understanding is that there was no aversion against offering this feature in the API. But nobody seem to have followed-up on this again.

#### My Use Case
My use-case is somewhat similar to that of #1166, whose author wished to _centrally_ manage the client secrets which simplifies provisioning and/or bootstrapping new environments greatly. But #83 is also relevant for me. I currently refrain from deploying an ownCloud Infinite Scale instance due to in compatibility with my current SSO stack [which doesn't support the ownCloud Android authentication flow](https://github.com/ory/hydra/issues/4039) natively.

[...]

PR #1619, opened 2026-07-23
feat: declaritive client secret configuration

schnz:
This PR is part 1/2 which provides a bare minimum foundation to satisfy the needs described in https://github.com/pocket-id/pocket-id/issues/1574. See #1622 for part 2/2.

# Feature
* `POST /api/oidc/clients/{id}/secret` now accepts an _optional_ JSON body payload with a `secret` key that allows to deterministically set the client secret

## Points for Discussion
* Whether or not to integrate the new API capabilities in the frontend (i.e. specifying a custom client secret via the frontend). I didn't added any frontend integration for two reasons:
  1. For the majority of use-cases (or all?) where a custom user ID and/or custom client secret is needed, the creation of resources will happen directly through the API. Namely: Terraform, NixOS, etc.
  2. After scraping through multiple issues I was left with the impression that there seems to be a general concern that "inexperienced users" might set "insecure" values, if they are allowed to do so.
* Whether the current validation rules make sense: Client Secret must be at least 16 chars and may only consist of printable ASCII characters
* Whether or not to adapt `POST /api/oidc/clients` to allow passing of a custom secret directly when creating a new  client.

**Disclaimer:** No(!) AI was used in writing any of the code, nor this text.


ItalyPaleAle (dismissed):
Can you please split this PR into two, so changes for `POST /api/users` and `POST /api/oidc/clients/{id}/secret` are in separate PRs?

Reason why I am asking this is that (while ultimately it’s @kmendell and @stonith404 ’s choice) I can see the value in the first change (`POST /api/users`) but for me, allowing users to provide their own client secrets would be a hard no.

I understand your Terraform use case, but for client secrets you should be able to have Pocket ID generate them and then they should be saved as outputs

schnz:
Hi @ItalyPaleAle,

Thanks for the quick reaction. I could certainly split this PR into two separate ones. But if in the end only one of them gets merged, this renders the whole change useless for me. So I'd rather try to understand and address your concern. Could you elaborate on that?

I'll try to summarize my arguments from #1574 why deterministic generation of client secrets should be allowed:
* On-premise installations of oCIS can be used with the official Android/iOS/Web client. These clients ship with [hard-coded secrets](https://doc.owncloud.com/server/next/admin_manual/configuration/user/oidc/oidc.html#client-secret) which must be reflected by the IdP.
  * The only possible workarounds for this are (1) manipulate the secret in the pocket-id DB or (2) re-compiling (and possible signing) the Android/iOS app. Option 2 is out of the question. And instead of going with option (1), I'd rather stay with the Ory stack for now.
* Which brings me to my second point: Current popular on-premise OAuth/OIDC providers (Ory Hydra, Authentik, Keycloak, ...) allow for a client secret to be provided. I don't believe that this is a design flaw, but instead of feature which is demanded by end-users.
* Last but not least: Using Terraform outputs to retrieve and propagate secrets would make it impossible to have a single source of truth where secrets are managed (usually a secret manager). It would also make it impossible to replicate an identical clone of one environment. Some of my applications are deployed to multiple locations. I wouldn't want these deployments to depend on a Terraform state file.


I hope we can find some common ground or compromise which addresses our both demands.

ItalyPaleAle:
Hey @schnz thanks for the detailed answer. Eventually this will be up to Elias and Kyle to decide, but TBH a lot of the things you have written seem like red (or at least orange) flags to me.

> On-premise installations of oCIS can be used with the official Android/iOS/Web client. These clients ship with [hard-coded secrets](https://doc.owncloud.com/server/next/admin_manual/configuration/user/oidc/oidc.html#client-secret) which must be reflected by the IdP.

As I am looking at this I'm honestly horrified. Mobile apps are considered public clients in the OAuth2 specs so they should not even be using a client secret under any circumstance. 

This seems like a (major) bug (maybe even security flaw) in ownCloud, and I'm not sure it's something that Pocket ID should bend for.

(Public clients should not use client secrets and should either use the authorization code flow with PKCE, or use the device code flow. What ownCloud is doing is [explicitly forbidden by the spec](https://datatracker.ietf.org/doc/html/rfc6749#section-10.1))

> The only possible workarounds for this are (1) manipulate the secret in the pocket-id DB or (2) re-compiling (and possible signing) the Android/iOS app. Option 2 is out of the question. And instead of going with option (1), I'd rather stay with the Ory stack for now.

_maybe_ we could create some script or util to manually change the secret in the DB for these very rare cases. The extra friction would be on purpose to discourage horror stories like what ownCloud is doing

_...but_ I still lean towards not supporting this

> Which brings me to my second point: Current popular on-premise OAuth/OIDC providers (Ory Hydra, Authentik, Keycloak, ...) allow for a client secret to be provided. I don't believe that this is a design flaw, but instead of feature which is demanded by end-users.

I understand this, but I am still not convinced this is the right thing to do. If we let users set client secrets themselves they will:

- choose insecure client secrets
- re-use client secrets among apps (more below)
- do horrifying things like ownCloud does

I think we have the responsibility here of protecting users against themselves by providing a "secure by default" option.

> Last but not least: Using Terraform outputs to retrieve and propagate secrets would make it impossible to have a single source of truth where secrets are managed (usually a secret manager).

It's pretty common AFAIK for secrets to be generated by a provider and then managed by Terraform. The source of truth would be what Pocket ID generates, and if this needs to be put in a secret manager, terraform should be able to do that by getting the secret from Pocket ID and copying it into the secret manager.

Many cloud services behave the same way, including non-auth stuff like Azure Blob Storage or Vercel's AI Gateway (just to name a few I'm familiar with)

> It would also make it impossible to replicate an identical clone of one environment. Some of my applications are deployed to multiple locations. I wouldn't want these deployments to depend on a Terraform state file.

Yeah that's precisely what I worry about :) you should not re-use secrets across different environments like prod and dev. I really, really recommend against that (ask me how I know!)

---

By the way if you control the apps have you considered using OIDC federation in lieu of a client secret? This way you can manage the keys yourself. See: https://pocket-id.org/docs/guides/oidc-client-authentication

schnz:
I've splitted the PRs now.

Thanks for the elaboration, @ItalyPaleAle. It certainly helped me to understand your point of view. I do agree on the oCIS part (embedding well-known secrets in public clients is an antipattern) - but even so: it remains to be a practical problem without a proper solution.

I mostly disagree on the other parts and I could elaborate on it, if there is interest in that. Unfortunately the proposed solutions will not work in my setup (air-gapped environments with partially replicated deployments that are synced every once in a while).

However, I understand that it isn't my place to make this decision. I still _do_ hope that a compromise can be found as this is ultimately a determining factor of whether or not Pocket ID can replace Ory as IdP in my setup.

ItalyPaleAle:
Thanks! Waiting for Elias and Kyle to comment too

One last note:

> I do agree on the oCIS part (embedding well-known secrets in public clients is an antipattern) - but even so: it remains to be a practical problem without a proper solution.

I would suggest actually open the issue upstream with OwnCloud. This is not just anti-pattern (and against the specs), but more so it's unsafe as it risks tokens being stolen. OwnCloud should really implement PKCE and omit the client secret

James18232:
Sorry to hijack this, but i have experience with ocis and hardcoded secret change PR’s 😂 - i did initially exactly as @ItalyPaleAle suggested not to and manually altered the pocket id db to make it work with a public fixed secret. Felt so wrong lol - did get me more familiar with how bcrypt worked though..

i also put in a PR ages ago to allow custom secrets (with oCis as the underlying reason) but in the end agreed a public secret is no secret at all so its just a bad idea. We did allow custom client id at the time though (which i see has been merged for the cli after you split the pr)

Fyi, ocis actually does implement pkce so just make it a public client in pocket id with the expected client id set and it will work fine 👍.

Agree an upstream change is better but i recall seeing a forum post somewhere this was rebuffed in upstream some years ago..  strange they bothered to implement pkce but not follow the spec for secrets 🤷

stonith404:
@schnz Thanks for these PRs! 

I agree with @ItalyPaleAle that the best solution would be to use a key vault but for some simple installations that might be an overkill. Maintainers of the NixOS and Terraform packages should document that it's not recommended to set client secrets manually though.

stonith404 (approved):

schnz:
Thanks for merging this, despite the reservations!

helmut72:
@ItalyPaleAle 

> I think we have the responsibility here of protecting users against themselves by providing a "secure by default" option.
> (...)
> The source of truth would be what Pocket ID generates

How do I get the secret from Pocket ID after creating a new OIDC client via API? Then I would use this. Thanks.

Edit: Found it. Need to post /api/oidc/clients/test/secrets without any options...

ItalyPaleAle:
Yes. Also note the API changed in 2.14.0 :)

PR #1624, opened 2026-07-26
feat: migrate app locks to Francis

ItalyPaleAle:
## What does this PR do?

Part of #1549

As part of this feature, adds an HAEnabled config flag. This is currently not configurable via env var yet (until we're done with supporting HA).

When HA is disabled the actor host registers with a max-hosts limit of 1, so a second replica fails to register and the process exits

The `import` command now takes an exclusive-access lease via the Francis clusteradmin package

PS: By removing the app lock, we are now reducing the load on the database too

## AI Usage

Co-authored with Claude's help


kmendell (commented):
Overall looks good to me so far, i just got back though so still catching up on everything. So i didnt do a in depth review yet.

The only part i question right now is the HAEnabled in the kv table, is this only because its not a env yet? Wouldnt it make sense for it to follow the existsing way the env settings are set in the app config service? (again if its just temp ignore me)

ItalyPaleAle:
> The only part i question right now is the HAEnabled in the kv table, is this only because its not a env yet? Wouldnt it make sense for it to follow the existsing way the env settings are set in the app config service? (again if its just temp ignore me)

That's a good question.

My idea is that you'll start Pocket ID either in HA mode or not, but once that's set you will not be able to change it unless you export and re-import the data. It's persisted in the database to prevent it from being changed in the future.

I am still ironing out whether it will be possible to change the value "live" (without export/import) in the future, as there are some things to consider, so if that becomes possible we may be able to remove the value from the DB.

stonith404:
@ItalyPaleAle LGTM, thanks.

Just one question, can you elaborate why HAEnabled is immutable? I don't really see a reason why this is necessary since this environment variable just changes host limit and polling/health-check timings.

stonith404 (approved):

ItalyPaleAle:
I have removed the thing that makes HAConfig immutable.

Originally I was considering using it to switch to a "standalone" provider for Francis when HAConfig was 0, so then migrating to a regular provider would have required exporting/importing DB data. Upon further consideration, this wouldn't work because a standalone provider can't allow the import/export code to work.

PR #1645, opened 2026-08-04
feat: include Francis data in Pocket ID backups

ItalyPaleAle:
## What does this PR do?

Part of #1549

Data is exported in the portable format Francis uses

## AI Usage

Co-authored with Claude's assistance


kmendell (approved):
Overall looks good to me, Ill let elias look over as well befroe merging though , incase i missed something.

stonith404 (approved):
LGTM thanks. 

I adapted the e2e tests so that the example import is static and not dynamically generated. Additionally the test now checks if the imported binary is exactly the same size as the exported. Is this assertion safe? We can't directly compare the checksums because `actor_state_expiration_time` can changes between import and export.

stonith404:
Nope it's not safe 🙂
Screenshot 2026-08-05 at 21 17 30@2x

PR #1646, opened 2026-08-04
feat: add observability to actors

ItalyPaleAle:
## What does this PR do?

This is related to #1598 and it is meant to help troubleshooting errors

This PR updates Francis to beta 19 which includes support for observability: actor methods and queries are now logged (when debug logs are enabled) and included in traces.

**"breaking" change**: this disables Gorm query logging because it would have caused logs for Pocket ID's own queries to be logged twice. This comes with two small changes:

- Query parameters are not included by default anymore, but need an explicit `LOG_QUERY_ARGS=true` - this is because the parameters could contain sensitive values (SQL queries continue to be logged when log level is debug or when they are slow)
- Queries returning 0 rows were always logged by Gorm as errors, even though in almost all cases it was an expected behavior - this has been removed

## AI Usage

Partly hand-written, partly updated with Claude


stonith404 (approved):
LGTM, thanks!

PR #1647, opened 2026-08-04
refactor: move GeoLite to its own package

ItalyPaleAle:
## What does this PR do?

Related to #1549 

Move the GeoLite code to its own package `geolite`.

The GeoLite database file is considered "cache", which means that in a multi-node scenario, each node can maintain its own local copy of the database.

The cache _can_ be shared, but doesn't have to, when not convenient. Considering that the cache _can_ be shared, we take care of limiting the likelihood that more than one replica is downloading the database at the same time, and if that were to happen anyways we ensure the file is written atomically.

It remains possible to manually place a GeoLite database file for Pocket ID to load. This PR adds support for fsnotify to automatically reload the database when the file changes.

## AI Usage

Created with Claude's help


stonith404:
I'm not really sure if storing the Geolite DB in an actor is the right approach. 

Storing the database in memory will at least double the idle memory use of Pocket ID - from ~50mb to at least ~110mb. During a refresh it even climbs to 316mb until the garbage collector does its work.

Also it's pretty common that Pocket ID runs air-gapped, so we must still support the option to specify the db as a file. 

Can't we just use an actor for the refresh of the database and keep storing the database on disk? If there are multiple replicas without any persistent volume mount each replica has to download it's own db which is not clean, but still better than having such a significant memory increase.

ItalyPaleAle:
Understood on memory usage, makes sense.

I would like to avoid shared mountpoints because they're hard to manage in most topologies. Let me see if I can think of a separate design...

How strict is the requirement that in air-gapped environments, the DB must be loaded from a file? Could we say that people need to host their DB on a HTTP endpoint of their choosing (could be in the LAN)? Or maybe offer an endpoint to upload it?

stonith404:
The air-gapped environments are more common than I thought, based on some comments in issues. I understand that share mountpoints could be hard to manage but forcing users to upload the database via a script or spinning up an HTTP server is too complex in my opinion. 

My thesis is, that people that run Pocket ID in HA mode are probably very experienced and know how to set up firewall rules, so that they can just allow outbound requests to Geolite, or they know how to manage shared mountpoints. So I would still keep the support for referencing the Geolite DB as a file.

ItalyPaleAle:
@stonith404 understood.

I have completely rewritten the PR so now it doesn't use actors at all, it doesn't need it. It nevertheless includes some improvements that will make it better for use in HA scenarios. See the [updated description](https://github.com/pocket-id/pocket-id/pull/1647#issue-5062467607)

It was important for me to not require a shared data folder across all nodes. This isn't to say that you can't have a shared volume, if you have a way to do that, but it's _not required_. This is because depending on the topology, shared volumes can be challenging to get.

**On cache vs state**

I have not touched in this PR the default path for the GeoLite DB, which is `data/GeoLite2-City.mmdb`.

This means that in the `data` folder we now have a mix of things that are state (and sometimes are stored in the database) and cache:

- `data/GeoLite2-City.mmdb` -> cache
- `data/uploads/application-images` -> state
- `data/uploads/profile-pictures` (excluding `defaults` directory) -> state
- `data/uploads/profile-pictures/defaults` -> cache

Would it make sense (for v3) to rethink the layout of the data folder, and maybe split it into two: `data` (state only) and `cache` (cache only)?

The idea is that cache volumes are fine if lost. In a Docker scenario, you could use a named volume per each replica, which is fine if it's lost. On K8s, you could use an emptyDir volume.

The cache _can_ be shared among multiple replicas, but it doesn't have to be.

stonith404 (approved):
LGTM, thanks for the changes!

stonith404:
@ItalyPaleAle Yes, I agree. Your suggestions would make sense for v3.

PR #1653, opened 2026-08-05
fix: separate TLS inputs and preserve certificate reloads

### LINKED ISSUE #1595: 🐛 Bug Report: TLS_CERT autoreload doesn't work reliably, and the docs are confusing
### Reproduction steps

I've just tried to set up Pocket ID and was confused by the following statement in the documentation regarding the TLS_CERT environment variables: 

> Base64-encoded TLS private key. When set along with TLS_CERT, the server enables HTTPS. The server monitors the certificate and key files and will automatically reload them when they change.

If I set TLS_CERT to the raw base64-encoded key, how can it monitor and reload any file?

Then I grepped the codebase for TLS_CERT and found the following test case in env_config_test.go: "should fail when TLS cert is set without key". 

In that test case, it sets "TLS_CERT" to a file path, not an actual base64-encoded certificate. 

Then I thought, okay, it's just the docs that are confusing, TLS_CERT is supposed to take a file path, not an actual cert. 

And then I found the test case "should load TLS cert and key file contents", which clarifies that there's also the environment variables "TLS_CERT_FILE" and "TLS_KEY_FILE" that can be set to a file path, which are entirely undocumented in the docs?

Also, in env_config.go the function `validateTLSConfig`, which seems to expect the variable "TLS_KEY" to be a string with a file path/name, not "TLS_KEY_FILE"?

### Actual behavior

The config seems to be so confusing I'm not sure what the expected behaviour even is. Is it even supposed to be possible to set a raw base64 string as cert and key without passing a file path? Why are TLS_KEY and TLS_KEY_FILE used interchangeably in the test cases, sometimes containing a path and sometimes containing a key?

I'm assuming that the correct behaviour would be for the docs to clearly state that TLS_KEY and TLS_CERT can be set to the raw base64-encoded cert, without automatic monitoring, _or_ instead, you could set TLS_KEY_FILE and TLS_CERT_FILE to filepaths containing the SSL cert and key. 

That would mean the docs should be modified to include these two new environment variables, and more important, most of the test cases would need to be updated so they'd correctly use TLS_KEY when testing with raw key data, or TLS_KEY_FILE when testing with a file path. 

And if I'm understanding the environment code correctly, all the certificate reading in env_config.go would also need to be adapted. 

### Expected Behavior

Environment variables in the documentation are consistent with how they're implemented in env_config.go and how they are tested in env_config_test.go. One given variable should only ever contain _either_ raw data _or_ a file path.

### Pocket ID Version

v2.11.0 (checked on latest master 525946e94e32af99755a221310be668281d4f955).

stonith404:
## What does this PR do?

Fixes #1595.

TLS configuration previously mixed inline certificate data with certificate file paths. File-based certificate watching also stopped working after certificates were atomically replaced because fsnotify watched the original files rather than their parent directories.

## AI Usage
Used Codex for the fsnotify handling.

## Contributing Guidelines

### DIFF
--- a/backend/internal/bootstrap/router_bootstrap.go
+++ b/backend/internal/bootstrap/router_bootstrap.go
@@ -12,7 +12,6 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -467,21 +466,14 @@ func enrichRequestLog(c *gin.Context, record *slog.Record) *slog.Record {
 
 // tlsCertProvider holds certificates that can be dynamically reloaded
 type tlsCertProvider struct {
-	certMutex   sync.RWMutex
-	cert        *tls.Certificate
-	certFile    string
-	keyFile     string
-	forceReload atomic.Bool
+	certMutex sync.RWMutex
+	cert      *tls.Certificate
+	certFile  string
+	keyFile   string
 }
 
 // GetCertificate implements tls.GetCertificate interface for dynamic certificate loading
 func (p *tlsCertProvider) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
-	if p.forceReload.Load() {
-		p.certMutex.Lock()
-		p.forceReload.Store(false)
-		p.certMutex.Unlock()
-	}
-
 	p.certMutex.RLock()
 	defer p.certMutex.RUnlock()
 	return p.cert, nil
@@ -576,7 +568,6 @@ func (p *tlsCertProvider) StartWatching(ctx context.Context, watcher *fsnotify.W
 			if err := p.reloadCertificate(); err != nil {
 				slog.Error("Failed to reload TLS certificate", "error", err)
 			} else {
-				p.forceReload.Store(true)
 				slog.Info("TLS certificate reloaded successfully")
 			}
 
--- a/backend/internal/bootstrap/router_bootstrap.go
+++ b/backend/internal/bootstrap/router_bootstrap.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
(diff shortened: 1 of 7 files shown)


backend/internal/common/env_config.go
```diff
@@ -75,8 +75,10 @@ type EnvConfigSchema struct {
 	SystemdSocket   bool   `env:"SYSTEMD_SOCKET"`
 	LocalIPv6Ranges string `env:"LOCAL_IPV6_RANGES"`
 
-	TLSCertFile string `env:"TLS_CERT" options:"file"`
-	TLSKeyFile  string `env:"TLS_KEY" options:"file"`
+	TLSCert     string `env:"TLS_CERT"`
```
ItalyPaleAle, line 79: Nit: add a comment that we aren't adding `options:"file"` because we need to manage that by hand - make it explicit

```suggestion
    // TLS cert and key need special treatment with fsnotify, so we aren't using `options:"file"`
	TLSCert     string `env:"TLS_CERT"`
```

backend/internal/bootstrap/router_bootstrap.go
ItalyPaleAle: While this wasn't touched in this PR, it's related, and I don't understand this. Why are getting a lock to change an atomic value? Shouldn't this just be `updated := p.forceReload.CompareAndSwap(true, false)`?

But even then, I don't understand the purpose of `forceReload`?

ItalyPaleAle (approved):
Approved with one comment

kmendell (approved):
LGTM

Leseratte10 (approved):
Thanks. I haven't tested the changes yet but I read through the code and it looks like it'd resolve the issue I reported.

backend/internal/bootstrap/router_bootstrap.go
stonith404: Yeah `forceReload` is unnecessary, good catch.

ItalyPaleAle (approved):

PR #1660, opened 2026-08-06
fix: in non-HA mode, increase AlarmLeaseDuration

ItalyPaleAle:
## What does this PR do?

Mirrors the work in #1602 and #1624 for alarm poll interval and health checks when in non-HA mode.

As verified in #1598, Francis was still performing a query to renew alarm leases every 10s, which is too quick in non-HA mode. Relaxing the duration to 180s which means that renewals are every 170s.

Note: in case of a non-clean shutdown (e.g. app crashes), alarm leases aren't released, so now it can take an average of 1.5 mins (half of 180s) until a new instance can pick up the alarms. Given what we use alarms for, this should be acceptable.

## AI Usage

Hand-rolled

## Contributing Guidelines


stonith404 (approved):

ItalyPaleAle:
Looks like there's an incident with GitHub Actions https://www.githubstatus.com/
image

PR #1661, opened 2026-08-06
fix: brotli compression ignored

### LINKED ISSUE #1649: 🐛 Bug Report: brotli compressed content is never served
### Reproduction steps

Set `Accept-Encoding: br` and `Accept-Encoding: gzip`, GET a static asset and notice for the former, there is no `Content-Encoding` header. Only the latter includes `Content-Encoding: gzip`.

### Expected behavior

With `Accept-Encoding: br`, the response would have `Content-Encoding: br`.

### Actual Behavior

It doesn't. This is due to no `.br` files ever even ending up in the build.

It seems to come down to, some Claude investigation revealed convincingly, the fact that `vite-plugin-compression` isn't meant to be used [like this](https://github.com/pocket-id/pocket-id/blob/a52620666961d620b496b8718c2e4a6a9b485066/frontend/vite.config.ts#L21). There's [a shared cache for all instances](https://github.com/vbenjs/vite-plugin-compression/blob/077eff37f45f12896326d687aa0e08f99b2b7a07/packages/core/src/index.ts#L15), so [the first compression type target wins](https://github.com/vbenjs/vite-plugin-compression/blob/077eff37f45f12896326d687aa0e08f99b2b7a07/packages/core/src/index.ts#L84) and there is [never a `.br` file to be served](https://github.com/pocket-id/pocket-id/blob/a52620666961d620b496b8718c2e4a6a9b485066/backend/frontend/frontend_included.go#L197).

It's also been unmaintained for 4 years so it'd be good to migrate away. [`vite-plugin-compression2`](https://github.com/nonzzz/vite-plugin-compression) seems like a good option and it accepts a _list_ of algorithms.

### Pocket ID Version

v2.12.0

### Database

sqlite

### OS and Environment

Talos + Kubernetes

### Log Output

_No response_

stonith404:
Replaces #1655 and fixes #1649. 

Instead of lazily compressing in the backend or the `vite-plugin-compression` we use the built in `precompress `which creates compressed brotli and gzip files on build time.

### DIFF
--- a/backend/frontend/frontend_included.go
+++ b/backend/frontend/frontend_included.go
@@ -11,6 +11,7 @@ import (
 	"mime"
 	"net/http"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


ItalyPaleAle (approved):
LGTM, seems like a good approach!

ItalyPaleAle:
PS: I force-pushed the exact same commit to re-trigger GH Actions

PR #1662, opened 2026-08-06
feat: add settings for limiting what passkeys can be used

stonith404:
## What does this PR do?

This PR adds three configuration options to limit which kind of passkeys can be used to sign in and sign up.

## Screenshots
Screenshot 2026-08-06 at 20 49 54@2x

## AI Usage
Created with Codex, UI and actor default state logic modified manually. Everything has been reviewed by hand.

## Contributing Guidelines


backend/internal/webauthn/service.go
```diff
@@ -408,6 +417,37 @@ func (s *Service) updateWebAuthnConfig(dbConfig *appconfig.AppConfigModel) {
...
+		return protocol.Platform
+	case "cross-platform":
+		return protocol.CrossPlatform
+	default:
+		return ""
+	}
+}
+
+func allowsSyncedPasskeys(dbConfig *appconfig.AppConfigModel) bool {
+	return dbConfig.WebauthnAllowSyncedPasskeys != "false"
```
ItalyPaleAle:
```suggestion
	return dbConfig.WebauthnAllowSyncedPasskeys.IsTrue()
```

backend/internal/webauthn/service.go
```diff
@@ -408,6 +417,37 @@ func (s *Service) updateWebAuthnConfig(dbConfig *appconfig.AppConfigModel) {
 	s.webAuthn.Config.RPDisplayName = dbConfig.AppName.String()
 }
 
+func userVerificationRequirement(dbConfig *appconfig.AppConfigModel) protocol.UserVerificationRequirement {
+	if dbConfig.WebauthnUserVerification == "preferred" {
+		return protocol.VerificationPreferred
+	}
+
+	return protocol.VerificationRequired
```
ItalyPaleAle, line 425:
```suggestion
	switch dbConfig.WebauthnUserVerification {
	case "required":
		return protocol.VerificationRequired
	case "preferred":
		return protocol.VerificationPreferred
	default:
		// Should never happen
		return protocol.VerificationRequired
	}
```

ItalyPaleAle (approved):
LGTM, with two minor nits

backend/internal/webauthn/service.go
ItalyPaleAle: Technically someone could modify the value in the DB and change it to something different than "required" or "preferred"

stonith404: I agree, but your suggestion behaves exactly the same as the original code?

Leseratte10:
Thanks for this update. Is there a particular reason why this only allows you to set the UV flag to 'preferred' or 'required' but doesn't allow the admin to set it to 'discouraged'?

stonith404:
@Leseratte10 I just don't see a reason why an admin wants to discourage users to not use verification if they have configured it. Or do you see really a use case in this?

PR #1679, opened 2026-08-10
feat: support multiple client secrets per OIDC client

### LINKED ISSUE #1659: 🚀 Feature: Multiple client secrets per OAuth2 client
### Feature description

Allow each OAuth2 client/app to have multiple client secrets, or at least 2, which can be rotated independently.

Bonus if the secrets could made expire independently

### Pitch

This helps with rotating secrets for app without downtime.

Currently, if I want to rotate a secret, first I need to generate a new one in Pocket ID, which immediately invalidates the old one, breaking apps until they are updated.

ItalyPaleAle:
## What does this PR do?

Fixes #1659

Adds support for multiple client secrets which are helpful when we need to rotate a client secret without downtime. The design matches what was discussed in #1659 (1-20 client secrets, can have an expiration date set optionally).

A few additional notes on implementation details made in this PR, that were not surfaced in the initial discussion in the issue:

- Create a new tab "Credentials" for OIDC apps, where we manage both client secrets and federated credentials
- All client credentials are stored in the database in the existing `oidc_clients` table's `credentials` column. That was already a JSON document containing federated credentials, and now it has the client secrets too. Using the same JSON column allows us to be flexible, especially since we never query for those separately
- The first 4 characters of a client secret are stored in plain-text, so admins who have a client secret know which row it is
- Expired client secrets are not automatically garbage-collected. The reason is that admins may want to see the first 4 chars and know which ones to delete/rotate
- Client secrets are stored in the database hashed with plain SHA-256. This is a change from the "old" behavior that used bcrypt. Because client secrets are strings of 32 randomly-generated  chars, they have sufficient entropy and don't need more complex (slower) hashes like bcrypt.

## Screenshots

localhost_1411_settings_admin_oidc-clients_6e231e95-9731-470f-bc2f-acf6ebc71c27_user-group-selection-page=1

## AI Usage

Co-authored with Claude, reviewed/edited by hand

## Contributing Guidelines


kmendell (commented):
LGTM, I did some testing locally and everything seems correct, though i don't have a actual way to test this in a "prod" env.

Elias has the final vote on this one as i know he had ideas from the issue (all of which seem to be implemented here)

frontend/src/lib/types/oidc.type.ts
```diff
@@ -21,8 +21,23 @@ export type OidcClientFederatedIdentity = {
...
+};
+
+// The clear-text value of a secret is only returned when it is created and cannot be retrieved afterwards
+export type OidcClientSecretCreated = OidcClientSecret & {
+	secret: string;
+};
+
 export type OidcClientCredentials = {
 	federatedIdentities: OidcClientFederatedIdentity[];
+	secrets?: OidcClientSecret[];
```
stonith404: Can we default it to an empty list? `omitempty` would have to be removed on the backend dto.

backend/internal/model/oidc.go
```diff
@@ -88,6 +94,70 @@ func (c OidcClient) IsMetadataDocument() bool {
...
+	OidcClientSecretHashSHA256 OidcClientSecretHashAlgorithm = "sha256"
+
+	// OidcClientSecretHashBcrypt is only found on secrets migrated from the single-secret column that Pocket ID used before it supported multiple secrets
+	// Those hashes cannot be converted to SHA-256, so they are kept as-is and verified with bcrypt until the admin rotates the secret
+	OidcClientSecretHashBcrypt OidcClientSecretHashAlgorithm = "bcrypt"
+)
+
+// OidcClientSecret is a single client secret of an OIDC client, stored hashed in the credentials JSON document
+type OidcClientSecret struct {
+	// ID uniquely identifies the secret within the client, so it can be deleted without knowing its value
```
stonith404: Can we remove comments? They seem pretty unnecessary for me.

stonith404 (approved):
Please see my two comments, just nitpicks. I really like the UI!

PR #1635, opened 2026-07-30
refactor: standardize API error handling

stonith404:
## What does this PR do?
This PR replaces generic and ad hoc API error handling with structured application errors.
Here is an example:
```json
{
  "error": "User not found",
  "code": "user_not_found",
  "details": {},
  "request_id": "564f383e-baa6-4317-a57f-a931e0ee29f0"
}
```

The error handling in the backend itself has improved as well. This PR introduces a new `apperror` package which replaces the `common/errors.go` file. Yeah 100 files changed 🫠, but 80% of them are just repetitive refactorings.

Additionally, I made some more improvements to make error handling cleaner:
- Migrateed application handlers to return errors through httpserver.Handle instead of attaching errors to the Gin context manually.
- Added shared BindJSON, BindOptionalJSON, and FormFile helpers that distinguish malformed requests from validation failures.
- Explicitly map expected persistence failures in the relevant services instead of globally treating every `gorm.ErrRecordNotFound` as a public 404.
- Keeps OIDC authorization policy denials on Pocket ID’s interaction error page so users can see why access was refused.

## AI Usage
Codex assisted with research and the implementation. I manually reviewed every line and improved the design manually.

## Contributing Guidelines

Have you read the [Contributing Guidelines](https://github.com/pocket-id/pocket-id/blob/main/CONTRIBUTING.md)?

- [x] If I'm implementing a new feature, I have opened an issue first, or commented on an existing one, to discuss the implementation details.
- [x] I have read the [AI Usage Guidelines](https://github.com/pocket-id/pocket-id/blob/main/CONTRIBUTING.md#ai-usage-policy).

PR #1684, opened 2026-08-11
docs: add error responses to openapi spec

aclerici38:
## What does this PR do?
Adds `@Failure` annotations to capture the new defined error messages from #1635 in the generated swagger spec. This adds the error as a response in the website and any machinery that uses the generated swagger.yaml can now use the typed errors. The current spec doesn't declare any error messages and it's not clear outside the codebase the error responses are structured now.
This also exports the error body as `dto.ErrorDto` since swag can only reference exported types. The json tags are unchanged so the response is the same as before.

```diff
--- swagger.yaml  /api/custom-claims/user-group/{userGroupId} (before)
+++ swagger.yaml  /api/custom-claims/user-group/{userGroupId} (after)
@@ -25,6 +25,10 @@
         items:
           $ref: '#/definitions/dto.CustomClaimDto'
         type: array
+    default:
+      description: Error
+      schema:
+        $ref: '#/definitions/dto.ErrorDto'
   summary: Update custom claims for a user group
   tags:
     - Custom Claims
```

Those refs resolve to
```yaml
dto.ErrorDto:
  type: object
  properties:
    code:       { $ref: '#/definitions/apperror.Code' }
    details:    { type: object, additionalProperties: {} }
    error:      { type: string }
    request_id: { type: string }

apperror.Code:
  type: string
  enum:
    - internal_error
    - validation_failed
    - invalid_request_body
    - request_timeout
    - not_found
    - already_in_use
    # ... 56 total
  x-enum-varnames:
    - CodeInternal
    - CodeValidationFailed
    - CodeInvalidRequestBody
    - CodeRequestTimeout
    - CodeNotFound
    - CodeAlreadyInUse
    # ...
```

**NOTES**:
I didn't see a way to generate the spec with the actual http codes returned on error so I used the reserved `default` response , but it doesn't read super well in the docs as-is. AFAICT the only alternative is to write out each response code as a separate annotation, which I can do if preferred, but my assumption is that's not something you'd like to hand-maintain.

The second commit adds a blank import to any controllers that don't import `dto`. I can revert that if preferred and not have those errors documented. Swag only resolves annotation types through a file's imports, so `dto` needs to be imported wherever there are `@Failure` annotations.

On the website all the errors just contain a ref to `dto.ErrorDto`, I plan to look into submitting a PR to the website (regardless if this is accepted) to resolve all refs as the successful responses show them as well.

## Screenshots
Screenshot 2026-08-10 at 8 27 57 PM

## AI Usage
Argued with Opus 5 for awhile about the best way to do this, the implementation is mostly from Opus 5, driven by myself, with some small manual tweaks.

## Contributing Guidelines

- [ ] If I'm implementing a new feature, I have opened an issue first, or commented on an existing one, to discuss the implementation details.


stonith404 (approved):
Thank you!

stonith404:
Yeah I don't like the OpenAPI library we are using it's really a pain in the ass but I think this is a good temporary addition.

PR #1690, opened 2026-08-14
fix: flaky update application image e2e test

ItalyPaleAle:
Fixes e2e test failures like https://github.com/pocket-id/pocket-id/actions/runs/31771914920/job/94679431090 (unrelated from the PR, I've seen this fail in some cases)

Just need to wait for the DELETE to respond before verifying the logo is gone


stonith404 (approved):
Thanks!

Issue #1689, opened 2026-08-14
🚀 Feature: First Class Agent Identities


rioncm:
### Feature description

dd an AI Agent dedicated onboarding and authentication flow separate from the native passkey ceremony.

Pocket ID already provides the surrounding identity lifecycle: administrators can create users, issue one-time login links, assign groups and administrator status, and expose the resulting identity to applications through OIDC. The remaining gap is durable authentication for a `user` identity which is an agent only. 

The proposed user flow is:

```text
Administrator creates a normal Pocket ID user
  -> turns on Agent authentication 
  -> issues the existing one-time login code
  -> agent uses the code and registers a runtime public key
  -> agent proves possession of the private key for future logins
  -> Pocket ID issues its normal sessions and OIDC tokens
```

The administrator user form should expose **Agent authentication** as a binary on/off selector. It is not a separate agent identifier: the user's existing Pocket ID username and ID remain authoritative.

- Off: the user follows the existing passkey registration and authentication path.
- On: the user follows the runtime-key registration and authentication path.

The selector changes the authentication process only. It must not create a reduced identity class, change groups or claims, limit existing user or administrator functions, or prevent an agent-authenticated user from being an administrator. The boundary of user management and explicit permissios depends on human choice as it does in the real world.  

**PS:** I've already completed the implementation of this concept and am currently in testing. There is also a companion agent plugin to educate the agent on how to interact with Pocket ID. 

### Pitch

Managed software agents are increasingly durable organizational actors. They may need their own identity, group memberships, audit history, and access to OIDC applications, but many run headlessly and cannot complete Pocket ID's human-oriented passkey ceremony.

Using a human's Pocket ID session for an agent weakens identity separation and auditability. Giving the agent an opaque long-lived bearer secret would work technically, but makes the durable credential effectively a password. A registered asymmetric runtime key provides a closer analogue to a passkey: the private key remains with the runtime, Pocket ID stores only the public key, and the credential can be named, audited, and revoked.

A practical first implementation can remain narrow and reuse Pocket ID's existing account, bootstrap, session, OIDC, authorization, and audit concepts.

### Simplified implementation

1. Add an administrator-managed boolean to the user model, exposed in the UI as **Agent authentication** and defaulting to off.
2. Add a runtime credential associated with the normal user. Store a name, credential ID, Ed25519 public key, creation time, last-used time, optional expiration, and revocation time. Never accept or store the private key.
[...]

PR #1691, opened 2026-08-14
Completed development and documentation on first class agent

rioncm:
## What does this PR do?

Full Implementation of the First Class Agent proposed feature. 
Full documentation and code notation in .fork-docs

Enable a deliberately provisioned First-Class Agent to complete Pocket ID account onboarding from an administrator-issued one-time access token. Completed development is a focused identity-provider capability for agents and did not expand Pocket ID into an agent runtime, delegation system, policy engine, or general secret manager.

## Context
I am not a developer nor am I a vibe coder. My primary development focus is python and back end systems. `go` is a new language for me and beyond my skills to execute at the level of this project. However, I am an accomplished manager of such projects and reviewed all code (to the best of my ability), every commit, provided all development constraints, and confirmed the following of such constraints. 

Documentation in `.fork-docs` contains a change index and each change is commented with the index number to assist in review from those who should be the gate to merging. 

The work completed here primarily serves my own purposes and I would take no offense should it not be merged into the main branch.   

## Relevant Notes
- Database migrations and rollbacks are fully implemented and round-trip tested for both SQLite and PostgreSQL.
- Passkey and agent runtime-key authentication paths are mutually exclusive and enforced by both service logic and database constraints.
- Runtime private keys remain client-side; Pocket ID stores only public keys and non-secret credential metadata.
- Headless account completion, proof-of-possession login, replay protection, expiration, and revocation flows are fully covered.
- Runtime-authenticated accounts retain normal OIDC claims, roles, groups, and authorization behavior without introducing a separate identity model.
- Credential visibility, lifecycle management, and audit events are available through existing user and administrator controls.
Backend tests, frontend type checks, production builds, and end-to-end scenarios pass against both supported databases.
- Container built and deployed on K3s and currently in active use with multiple agent identities. 
- While this solves the identity problem for a durable agent is does not address per application agent functionality. 

## Screenshots
**Agent toggle in user**
CleanShot 2026-08-14 at 09 57 44

**Agent run time credentials**
CleanShot 2026-08-14 at 09 59 41

## Companion Agent plugin 
A companion skill and plugin is available at https://github.com/rioncm/pocket-id-agent-bootstrap

## AI Usage
Multiple AI agents and models were used for:
- concept clarification
- pre-development scope
- gap review
- development scope
- development
- code review
- agent endpoint and application endpoint testing
- final documentation

## Contributing Guidelines

## Post Script
Yes, this PR is a lot of words. No AI did not write it. This is only my second public PR and I want to ensure the feature, development, and my intent are crystal clear.


rioncm:
I'm not sure if you review the autoclosed PRs. If you don't I would like to
call your attention to this one.

It was closed automatically because I haven't been very active on public
projects and am still learning all the protocols. I was close to the file
limit but including several documentation pages pushed me over.  The issue
and PR present an interesting use case related to AI Agent Identity
management. Although it might not fit your vision for the project, I did
not want it to be missed simply because my GH profile was incomplete.

Thanks!

❌ max-changed-files PR has 54 changed file(s), exceeds maximum of 50
❌ conventional-title PR title "Completed development and documentation on
first class agent" does not follow conventional commits format
❌ description-max-length Description is 3734 chars, exceeds maximum of 2500
❌ max-additional-pr-template-sections PR has 4 additional section(s),
exceeds maximum of 3
❌ min-profile-completeness Profile completeness is 2/10, below minimum of 4
(missing: company, blog, location, email, hireable, bio, twitter, followers)

Rion Morgenstern

ItalyPaleAle:
@rioncm I'll be honest with you, I have no idea what this PR does. The PR description is 3,734 characters but it doesn't really explain what "First Class Agent" is or why we should consider it for Pocket ID.

Can you elaborate please? Is there a spec (RFC or equivalent) this complies with?

CC: @stonith404

rioncm:
@ItalyPaleAle

Sorry for the lack of clarity. This idea and its implementation stem from
work I am doing to integrate agents directly into the workplace environment
as "employees" with their own roles and permissions distinct from the
person directing their work.  If you like, I can post the container (50MB)
to Docker Hub or send you a direct link.  Again, I will not be offended if
this is rejected. I'm truly not sure if this direction reflects how others
view agents in the future of agentic work. Currently I am deploying Pocket
ID in two ways: alongside the existing identity system (Google Workspace)
in my workplace and as the primary identity system in my lab.

The information below would have made a much better "issue" than the one I
opened.

*Thesis:* Agents need a dependable authentication flow suited to their
tools and non-browser interfaces.

   - First-Class Agents are employee-like actors that should have durable,
   independently managed identities.
   - First-Class Agents should not share a human’s identity or
   authentication credentials.
   - Their authorization should be assigned independently, even when
   managed through the same roles, groups, and application policies used for
   human accounts.

*Problem:* Most identity systems primarily authenticate humans through
passwords, passkeys, multifactor authentication, and interactive browser
flows. OAuth Device Authorization enables clients on input-constrained
devices to obtain authorization, but it does not establish a durable agent
identity or agent-specific authentication credential.

AI agents often operate through credentials or sessions belonging to a
human resulting in conflation of identity and overly broad access.

*Solution:* Add an administrator-controlled authentication path to Pocket
ID for agent identities, using registered asymmetric keys and
proof-of-possession authentication.

*Implementation:* The addition of the IsAgent user flag that selects the
account’s authentication path. When IsAgent is enabled, Pocket ID prevents
passkey registration for that account and enables runtime-credential
registration and authentication using Ed25519 keys.

An administrator initiates enrollment through the existing one-time 'Login
Code' shared with the agent. The agent generates and retains its private
key locally, then registers the corresponding public key.

ID tokens report the authentication method as: "amr": ["pop"]

RFC 8176  defines the more specific swk and hwk values for software-secured
and hardware-secured keys. The OpenID Connect Extended Authentication
Profile separately registers pop for proof of possession when the identity
provider cannot assert how the key is secured.  Apart from the "amr":
["pop"] there are no changes to how Pocket ID implements OIDC. No isAgent or
other agent-specific claim is sent to configured applications.

*Result:* Agents have durable Pocket ID identities with independently
managed credentials, application assignments, groups, roles, and
permissions. Downstream applications use Pocket ID’s normal OIDC flow,
while administrators govern agent access separately from human accounts.
This simplifies administration, improves attribution, and reduces the need
for agents to reuse human credentials or inherit a human’s full access.



Rion Morgenstern

ItalyPaleAle:
It sounds like what you need is service-level authentication.

The way to do that in OAuth2 (and Pocket ID) is to use the client credentials flow. If you want to avoid client secrets, Pocket ID can work with federated credentials for client authentication already.

rioncm:
Service-level authentication and workload identity federation are
important, and they fit within the broader mental model I am using.
However, they do not fully address what I mean by a “First-Class Agent.”

Client credentials is a good fit for what I call a “Second-Class Agent”
(SCA): a durable agentic workload with its own client identity, but not an
account-level actor or durable persona. An SCA generally has a tightly
defined purpose and narrowly scoped access—for example, executing one step
in a business process or interacting with a limited set of API endpoints.
This makes it closer to a machine-to-machine service.

An SCA may perform work authorized by a person or a First-Class Agent,
although client credentials alone represents the SCA acting as itself
rather than carrying delegated user authority. The delegation relationship
would need to be expressed through another OAuth flow, token exchange, or
application-level authorization policy. An SCA may also use a smaller,
task-specific model when that is sufficient, reducing cost and compute
requirements.

A “First-Class Agent” (FCA), by contrast, is a durable, independently
managed actor or persona. In my model, an FCA is analogous to a
middle-manager employee, while an SCA is analogous to a line worker with a
specific assignment. An FCA can receive responsibilities, use multiple
applications, direct narrower agents, and have its access managed
independently over time.

I would deploy an FCA using a capable model in a fully isolated runtime,
such as a dedicated container or computer. Its runtime, credentials,
authenticated applications, and authorization grants should be separate
from those of its operator, manager, or any other user.

One shortcomming of my implementation is that it does not explicitly
identify the subject to downstream applications as an AI agent. The only
related signal is amr=["pop"], Administrators and relying-party developers
may want a separate claim—such as an actor-type claim—to distinguish agent
accounts from human accounts.



Rion Morgenstern

stonith404:
I agree with @ItalyPaleAle you would usually use the client credentials flow for that.

@rioncm Agent Identity is a really now topic because of that it is possible that I'm missing a use-case though. It would be cool if you share your project. You can either post it here or DM it via Discord (stonith404) if you prefer.

ItalyPaleAle:
@rioncm I just saw #1689 

I think I understand now the problem you're trying to solve. Essentially, Pocket ID implements OAuth2, and OAuth2 separates users (human identities) from apps/clients (machine/service identities). Your point is that in an agentic world, the line is blurred and we can have "human" identities (users) that are "non-human" (actually agents).

This seems to be first of all a "problem" with the current specs...

I am wary of adding an ad-hoc identity flow at this point, however, unless it becomes more of a standard.

To provide an alternative that would be spec-compliant, what if we were to complete implementing RFC 7523 to use **assertions for authorization grants**?

Today we implement that RFC (with a custom "extension" to make it compatible with cloud providers/k8s) for client authentication only (section 2.2), so assertions can be used in lieu of a client secret in the client credentials flow (or as client authentication in the authorization code flow).

The same RFC also specifies in section 2.1 that assertions could be used as authorization grants. This way you could present to the `/token` endpoint an assertion and get a token whose subject is a user, not a client.

The end result should be very similar to what you described in #1689, where an agent needs to prove possession of a private key. But this is more flexible because Pocket ID wouldn't need a per-user private key, and would just need to be configured to accept a public key in a JWKS somewhere. And this is an existing spec...

rioncm:
Thanks for the conversation about this. I think it is a very important
topic.  Here is a post I rushed out trying to communicate the broader
concept. https://blueplatypus.com/2026/08/17/an-agent-by-any-other-name/

I'm not sure I'd define my PR as an ad-hoc flow. It does not update or
modify any Pocket ID implementation of OIDC or OAuth. The best way to
describe the PR would be as a new channel.  This is similar to adding
password authentication along side the passkey implementation (though this
is not a recommendation to do so). The difference is that the channel has
an enablement flag, 'IsAgent', which locks it into the 'Runtime
Credentials' channel instead of the Passkey channel.

Without that flag, users are locked into the Passkey channel, which is the
standard for Pocket ID.  An added safeguard prevents an admin from flipping
that flag if there are existing Passkeys for regular users or Runtime
Credentials for non-regular (agent) users.

RFC 7523 Section 2.1 sounds very interesting for delegation to Second Class
Agents or sub-agents as I define them. I have been thinking about
delegation generally but have not spent any time fleshing out the
mechanics. I'll check this out for sure.

I know it's a big piece of code to review to verify my claims and it may
not be the right time or the right vision for Pocket ID. My intent is to
share not to direct or change the vision and application of the community.
The PR bridges a gap for me which has been on my radar for a long time. The
other identity systems I've experimented with are overly complex and Pocket
ID really hits a sweet spot.  I'm in the final stages of setting up a
consultancy to work with small to midsized businesses implementing agentic
work in their business processes. This new channel in Pocket ID may be the
missing part of my solution vision. I'm going to keep testing in my home
lab and current company to see if that is the case.

In my mind, this is an important concept and I'm happy to continue this
discussion. Also, I'm happy to answer questions about what the PR is and is
not. I'm also open to any specific direction you want to point me. I am a
curious fellow and recognize that I often "don't even know what I don't
know".


Rion Morgenstern

ItalyPaleAle:
I think the "ad-hoc" part here is that this is not a standard OAuth2 flow, nor a standard user authentication flow (like passkey or password). That's the root of my hesitation, i can understand the need for this, but I am concerned about adding a new protocol in Pocket ID that's "ad hoc" for this use case and not at least an industry standard.

The protocol you're suggesting, while not necessarily bad, is a custom CHAP protocol that relies on digital signatures, it's not a simple password (shared secret). I also think where this falls short is that it requires each agent to have its own private key, and even though the private key isn't sent to Pocket ID, it's a long-lived secret that needs to be managed somewhere.

> RFC 7523 Section 2.1 sounds very interesting for delegation to Second Class Agents or sub-agents as I define them

Why not for your "first class agents" too? Using RFC 7523 § 2.1 should allow you to get the exact same result as your custom CHAP protocol, while remaining OAuth2-compliant and well-defined in a RFC. It would also allow us to re-use most of the existing code in Pocket ID.

With assertions, your agents need to provide a JWT that's signed by _some private key_. This is the interesting aspect: they could own the private key and sign the assertions themselves (in which case this is pretty much the same case as your PR, except there's no challenge-response but uses standard JWTs), or in a better scenario could get the assertion from an external assertion provider. For example, if the agents run on K8s you can present a pod identity token (aka service account token). If you're using Tailscale, you can use my [tsiam](https://tsiam.italypaleale.me/). On a cloud provider (Azure has best support right now), the token can be signed by the cloud itself. The advantage with using assertions signed externally is that your agents do not have a private key themselves, which means one less secret to manage (and make sure doesn't leak) and more opportunities for auditing an agent's behavior.

ItalyPaleAle:
More context on how the assertions flow would work:

1. You decice on what assertion provider to use. I gave 3 options above (K8s service account tokens, cloud provider or other infra, tsiam for Tailscale), or you could build your own which authenticates agents in a way that makes sense for your use case and provides when invoked a JWT asserting the identity of the agent.
   - Token should contain as `sub` the user ID of the agent (in Pocket ID), `aud` as Pocket ID, `iss` as your assertion provider
   - Here I'm suggesting using an external service to provide assertions, but note that technically speaking, the agent itself could also sign its own assertions if it has a private key (which you need to make sure you're carefully protecting).
1. Client in Pocket ID is configured to accept assertions in place of authorization grants. The public key of the assertion provider is added to the client configuration.
   - Today (for client assertions) we support retrieving the public key from a JWKS endpoint only. We should (also for client assertions) allow uploading a fixed public key, as JWK or PEM.
1. Your agent needs a token:
   1. It obtains an assertion from the provider
   2. It invokes the `/token` endpoint directly presenting the assertion rather than the authorization code
   3. Pocket ID issues a token that represents the user for whom the assertion is signed (so in this case user identity = agent identity)
   4. You can use the token freely to invoke APIs

The part that I like about this is that there isn't a long-lived credential configured for each agent (unless you make the agents self-sign their assertions).

In the design you've suggested there's a long-lived credential, which is essentially a password (no expiration, must be protected carefully), with the only difference that's not symmetric/shared. But the risks about poor management of the credential remain.

rioncm:
That is super interesting. I'll map that to the infrastructure I am using
and report back.

Thanks for the direction!



Rion Morgenstern

PR #1713, opened 2026-08-22
feat: show admin UI warning for SQLite on networked filesystem

ItalyPaleAle:
## What does this PR do?

We tried including it in the docs. We tried showing a warning in the logs. We still get people opening issues caused by SQLite placed on NFS or other networked storage.

I don't think we can block it entirely, because of possible false positives, but we can show a bigger warning in the admin UI (similar to when there's updates), that cannot be dismissed. The warning links to docs that explain the problem.

The only way to dismiss the warning is to set the env var `DISMISS_SQLITE_STORAGE_WARNING=i accept the risks` (to the exact phrase, case-insensitive)

Docs PR: https://github.com/pocket-id/website/pull/295

## Screenshots

image

## AI Usage

Built with Claude help

## Contributing Guidelines


stonith404:
Since we want to deprecate the `controllers` module I think it would make more sense that we create a new module called `environment` (or similar). This will contain the version and sqlite warning endpoint. What do you think?

Also I think we should make `DISMISS_SQLITE_STORAGE_WARNING` just a bool. IMO the approach with the sentence is too aggressive.

ItalyPaleAle:
> Since we want to deprecate the `controllers` module I think it would make more sense that we create a new module called `environment` (or similar). This will contain the version and sqlite warning endpoint. What do you think?

Sounds good

> Also I think we should make `DISMISS_SQLITE_STORAGE_WARNING` just a bool. IMO the approach with the sentence is too aggressive.

I fear that by making it just a boolean, people will just ask Claude "fix it", Claude will do it, and people will not notice what they're dismissing. The amount of people that are ignoring the warning in the terminal, IMHO it needs to be stronger.

How about `DISMISS_SQLITE_STORAGE_WARNING=i accept the risks`?

stonith404:
Fair, yeah `DISMISS_SQLITE_STORAGE_WARNING=i accept the risks` works.

stonith404:
LGTM, thanks. Moved the alert into an alert component in top of the page content, feel free to merge it if you agree with the changes.
Screenshot 2026-08-24 at 22 34 49@2x

stonith404 (approved):

PR #1733, opened 2026-09-01
feat: use HS256 for Pocket ID's own session tokens + fixes

ItalyPaleAle:
## What does this PR do?

Pocket ID mints 2 types of tokens:

- Access tokens and ID tokens for external clients
- Session tokens for maintaining internal sessions

The two kinds have fundamentally different audiences: external consumers vs Pocket ID itself. However, they are currently signed with the same key.

While the current behavior isn't "wrong", it's unnecessary. For Pocket ID's own session tokens, we can use symmtric keys and HS256 for signing tokens, since no one outside of Pocket ID needs to validate the signature. The benefits are faster token verifications (and less CPU usage), and preventing people from possibly abusing Pocket ID's sessions for other purposes (no proof this is happening today, but it _could_).

Additionally, while working on this PR, I found and fixed a **bug** in how keys were stored in the database.

When Pocket ID starts up for the first time, it generates a new token signing key and stores it in the database. If two instances of Pocket ID start at the same time (something that shouldn't have been possible, but could be with HA mode - see #1549 ), they race each other to generate a key. The initial design wanted a first-write-win, where the first one to commit the key to the database won, and the other one crashed and restarted (then picking up the existing key). However, after changes were made (to support the key rotation command), this changed to "last-write-wins", which meant that 2 instances could run with two separate signing keys. This PR also fixes that, by having 2 separate methods for initial startup and key rotation. Additionally, in case of races, the replica that fails doesn't crash anymore, but just retries after a brief delay.

## AI Usage

Co-authored with Claude, reviewed and edited by me

## Contributing Guidelines


kmendell:
Glancing at it looks okay, im a bit busy atm, to do a full revie though hopefully i can find some time today or tomorrow.

kmendell (approved):
Overall Looks good to me

Issue #1473, opened 2026-05-14
Feature: security with session tracking, backchannel logout, and SSO


forkrails:
### Feature description

Pocket ID is great, and I appreciate the simple approach to OIDC and auth. However, there are certain design decisions which make me hesitate to use with family/friends.

Right now, it seems like if you log out of one app, it doesn't log out of the others? Pocket ID doesnt' seem to describe itself as an SSO provider anywhere I can find, so this is at least consistent. 

However, in general the expectation is that if login is unified, so is logout. If you log into two apps, log out of one expecting the other will log out, when it doesn't in reality, that seems like a serious security issue. As a admin, I can train myself to log out of all apps, maybe. But, for other users, that's just not guaranteed.

Implementing this would require some level of session tracking, which may be against the current statelessness goal (from what I can tell). However, its not so different from the refresh token implementation. Those are stateful of course, and maybe there are others too. 

It looks like most of the plumbing is already here, so not a big lift. This will create a meaningful improvement in security posture for non-technical users who otherwise expect true SSO.

### Pitch

Also, I really want to move away from keycloak 😂

forkrails:
Never mind, pocket id seems fairly slow in development, don't see this making it in.

m8426:
@stonith404 there are many upvotes on this issue

alecrubin:
I've been running Pocket ID as the IdP for Immich, with a second service planned, and this is the one gap that keeps offboarding manual. Removing someone from a group stops new logins but doesn't touch live sessions, so my runbook is "disable in Pocket ID, then go delete the user in the app". The second step is the one that actually revokes anything. Immich already implements the receiving side and accepts a `sub`-only logout token, so this would work end to end if Pocket ID could send one.

From a look at the code most of the pieces seem to be there already. `UserAuthorizedOidcClient` gives you the list of clients to notify, and `jwt_service` already holds the signing key and serves the JWKS, so minting a logout token looks like it would sit next to `GenerateAccessToken`. Delivery is the part with nothing to reuse: `internal/job` is cron actors rather than a work queue, so a v1 would probably be best effort POSTs at logout time with failures logged, and retries left for later if anyone wants them.

That plus a `BackchannelLogoutURL` field next to `LogoutCallbackURLs`, and `sub`-only tokens (no `sid`, no per-session tracking), would cover the common case without the session tracking work the title of this issue implies.

[...]

PR #1734, opened 2026-09-01
feat: add OIDC back-channel logout

alecrubin:
## What does this PR do?

Implements OIDC back-channel logout, so clients can end a user's session when Pocket ID revokes their access. This is the back-channel logout part of #1473 only. Session tracking and single logout when a user signs out of one app aren't covered, so that issue should stay open.

Logout tokens carry `sub` only, with no `sid` and no per-session tracking. Deliveries run as durable Francis jobs, so they survive a restart and failed sends are retried up to three times before being logged and dropped.

Adds a `backchannel_logout_url` column and form field on the OIDC client. `GenerateLogoutToken` mints a `logout+jwt` with the `events` claim, signed with the existing JWKS key. The `backchannellogout` package resolves targets from `UserAuthorizedOidcClient` and schedules one job per authorization, each POSTing the token form encoded, capped at 4 in parallel.

Tokens go out when a user is disabled or deleted, when a user revokes a client, when a client is deleted, and when someone loses access to a group restricted client. Group access uses a `NOT EXISTS` against the client's allowed groups, so a user who still reaches it through another group isn't logged out. Editing allowed groups from either the client or the group side, and turning group restriction on, take the same path. The LDAP sync is wired too, which is how membership changes for many deployments.

`backchannel_logout_supported` is advertised, with `backchannel_logout_session_supported` false since there's no `sid`. Signing out of Pocket ID sends nothing, because those sessions are stateless today. Happy to hold that metadata back if that seems misleading.

The URL is validated as an HTTP URL only, with no private address check, since self hosted RPs are the main use case. The field went in advanced options rather than the main URL grid, easy to move if you'd rather it sat with the other URLs.

## Screenshots

**The create form**, advanced options expanded:
create-mode-advanced

**The edit form**, advanced options expanded:
backchannel-logout-field

## AI Usage

Built with Claude Code. Design decisions were mine, I reviewed the diff, and the backend tests and golangci-lint were run locally before opening this.

## Contributing Guidelines


backend/internal/dto/oidc_dto.go
```diff
@@ -21,6 +21,7 @@ type OidcClientDto struct {
 	OidcClientMetaDataDto
 	CallbackURLs                        []string                 `json:"callbackURLs"`
 	LogoutCallbackURLs                  []string                 `json:"logoutCallbackURLs"`
+	BackchannelLogoutURL                *string                  `json:"backchannelLogoutURL"`
```
ItalyPaleAle: This doesn't need to be a pointer, right? An empty logout URL isn't a valid one, so we could just keep `string`

backend/internal/dto/oidc_dto.go
```diff
@@ -47,6 +48,7 @@ type OidcClientUpdateDto struct {
 	Description                         string                   `json:"description" binding:"omitempty,max=150" unorm:"nfc"`
 	CallbackURLs                        []string                 `json:"callbackURLs" binding:"omitempty,dive,callback_url_pattern"`
 	LogoutCallbackURLs                  []string                 `json:"logoutCallbackURLs" binding:"omitempty,dive,callback_url_pattern"`
+	BackchannelLogoutURL                *string                  `json:"backchannelLogoutURL" binding:"omitempty,http_url"`
```
ItalyPaleAle: Ditto

backend/internal/model/oidc.go
```diff
@@ -50,6 +50,7 @@ type OidcClient struct {
 	Description                         string
 	CallbackURLs                        datatype.StringList
 	LogoutCallbackURLs                  datatype.StringList
+	BackchannelLogoutURL                *string
```
ItalyPaleAle: Ditto

backend/resources/migrations/postgres/20260901120000_add_oidc_client_backchannel_logout_url.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE oidc_clients ADD COLUMN backchannel_logout_url TEXT;
```
ItalyPaleAle: Please make `NOT NULL`

backend/resources/migrations/sqlite/20260901120000_add_oidc_client_backchannel_logout_url.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE oidc_clients ADD COLUMN backchannel_logout_url TEXT;
```
ItalyPaleAle: Ditto, please make `NOT NULL`

frontend/src/lib/types/oidc.type.ts
```diff
@@ -54,6 +54,7 @@ export type OidcDiscoveryConfiguration = {
 export type OidcClient = OidcClientMetaData & {
 	callbackURLs: string[];
 	logoutCallbackURLs: string[];
+	backchannelLogoutURL?: string;
```
ItalyPaleAle:
```suggestion
	backchannelLogoutURL: string;
```

backend/internal/service/backchannel_logout_service.go
```diff
@@ -0,0 +1,264 @@
+package service
```
ItalyPaleAle: Per @stonith404, we are trying to avoid `internal/service` and create a separate package for each service. The package contains the `Service` as well as any related artifact. Eventually, the existing services should be migrated too...

backend/internal/service/backchannel_logout_service.go
```diff
@@ -0,0 +1,264 @@
...
+
+// targetsQuery selects the authorizations of clients that are registered for back-channel logout
+// Callers narrow it down to the users or the client whose access was revoked
+func (s *BackchannelLogoutService) targetsQuery(ctx context.Context, tx *gorm.DB) *gorm.DB {
+	return tx.
+		WithContext(ctx).
+		Model(&model.UserAuthorizedOidcClient{}).
+		Select("user_authorized_oidc_clients.user_id", "user_authorized_oidc_clients.client_id", "oidc_clients.backchannel_logout_url AS logout_url").
+		Joins("JOIN oidc_clients ON oidc_clients.id = user_authorized_oidc_clients.client_id").
+		Where("oidc_clients.backchannel_logout_url IS NOT NULL AND oidc_clients.backchannel_logout_url <> ''")
```
ItalyPaleAle:
```suggestion
		Where("oidc_clients.backchannel_logout_url <> ''")
```

backend/internal/service/backchannel_logout_service.go
```diff
@@ -0,0 +1,264 @@
...
+func (s *BackchannelLogoutService) notifyClients(ctx context.Context, targets []backchannelLogoutTarget) {
+	if len(targets) == 0 {
+		return
+	}
+
+	// The caller's request context ends with its HTTP response, so delivery detaches from its cancellation while keeping its values
+	go s.notifyClientsSync(context.WithoutCancel(ctx), targets)
+}
+
+func (s *BackchannelLogoutService) notifyClientsSync(ctx context.Context, targets []backchannelLogoutTarget) {
```
ItalyPaleAle: I would suggest moving this to be a Francis job, which makes it durable.

1. Create an actor type `BackchannelLogoutNotifier`
2. For each notification that you need to send, schedule (synchronously) a job with the actor
3. The actor will be the one that sends the notification in backgroun
4. Concurrency can be controlled by setting the [`WithConcurrencyLimit`](https://francis.italypaleale.me/docs/writing-actors/#registering-an-actor) when registering the actor

Docs for Francis jobs are here: https://francis.italypaleale.me/docs/jobs/

Why this?

1. The jobs are guaranteed at-least-once execution, and failed jobs can be retried automatically (you should set it so it doesn't retry forever, but with a cap of, say, 3 attempts)
2. The jobs are durable, so if Pocket ID crashes meanwhile, they are resumed after Pocket ID comes back online
3. You get concurrency control enforced by the actor runtime

backend/internal/service/user_service.go
```diff
@@ -196,7 +199,18 @@ func (s *UserService) UpdateProfilePicture(ctx context.Context, userID string, f
 }
 
 func (s *UserService) DeleteUser(ctx context.Context, dbConfig *appconfig.AppConfigModel, userID string, allowLdapDelete bool) error {
+	// The user's authorizations are gone after the delete, so the clients to notify must be resolved inside the transaction
+	notifyLogout := func() {}
 	err := s.db.Transaction(func(tx *gorm.DB) error {
+		if s.backchannelLogout != nil {
+			var err error
```
ItalyPaleAle: Use a different name for `err` inside the closure please

ItalyPaleAle (changes requested):
Thanks! A few initial comments from me

backend/internal/service/backchannel_logout_service.go
ItalyPaleAle: PS: Make sure you pull francis v0.1.0-rc.5 if you do this, as it contains a few relevant fixes

backend/internal/service/backchannel_logout_service.go
alecrubin: Moved it to `internal/backchannellogout`, laid out like the apikey and devicelogin packages. The one wrinkle was the import cycle with `internal/service`, so the service takes a small `TokenSigner` interface instead of the JWT service directly.

backend/internal/service/backchannel_logout_service.go
alecrubin: Reworked it as a job on rc.5. Each notification is dispatched to a `BackchannelLogoutNotifier` actor keyed by client and user, so sends for the same authorization stay in order. `WithConcurrencyLimit(4)` caps the parallel POSTs and `WithMaxAttempts(3)` bounds the retries, after which `JobFailed` logs it and gives up. Input that doesn't decode comes back as `ErrJobPermanentFailure` so it isn't retried.

backend/internal/backchannellogout/service.go
```diff
@@ -0,0 +1,267 @@
...
+		return
+	}
+	s.notifyClients(ctx, targets)
+}
+
+// NotifyUsersLostGroupAccess delivers logout tokens for group-restricted clients that the given users can no longer access
+// It must be called after the group membership change has been committed, and logs instead of failing because delivery is best effort
+func (s *Service) NotifyUsersLostGroupAccess(ctx context.Context, userIDs []string) {
+	s.notifyLostGroupAccess(ctx, userIDs, "")
+}
```
kmendell: I would get rid of the stub function and just make the other one public and call it directly, there's a few occurrences of this through these changes.

alecrubin: Collapsed the two into an exported `NotifyLostGroupAccess` that takes both filters. The other short ones each run their own query or have their signature pinned by Francis, so I left those, but happy to flatten more if you had others in mind.

alecrubin:
Gentle bump on this @ItalyPaleAle @kmendell. All the feedback so far is addressed. The one red check looks like a CI hiccup, the job failed setting up the Docker builder before any tests ran, so it probably just needs a re-run. Ready for another look whenever you get a chance.

ItalyPaleAle:
@alecrubin waiting on @stonith404 who was OOO for the last couple of weeks to confirm he's good with this feature before doing one final pass!

Abhirup0 (commented):
Reviewed the full diff, migrations, and test suite. The implementation is clean and well thought out:

- Capturing authorization targets inside the transaction before cascading deletes and dispatching the Francis jobs post-commit prevents both missing targets and phantom logout dispatches on rollback.
- Token claims, JOSE headers (`logout+jwt`), and discovery metadata strictly adhere to OIDC Back-Channel Logout 1.0 §2.4/§2.5.
- Disabling redirect following on the HTTP client (`http.ErrUseLastResponse`) correctly avoids dropping the POST payload.

One optional suggestion: in `sendLogoutToken`, when `res.StatusCode` is in the `4xx` range (e.g. `400 Bad Request` or `404 Not Found`), consider wrapping the returned error with `actor.ErrJobPermanentFailure`. Per OIDC Back-Channel Logout 1.0 §2.6, 4xx responses are not expected to succeed on retry, so marking them permanent avoids 2 redundant retries against dead or misconfigured client endpoints.

LGTM overall.

alecrubin:
Thanks for taking a look @Abhirup0. I'd rather keep retrying 4xx since the spec routes even transient RP failures through 400, and things like a stale JWKS cache after key rotation heal on retry. The attempts are capped at 3 either way.

alecrubin:
@ItalyPaleAle just for context, this feature came out of a conversation with @stonith404 on #1473 before I opened the PR, the approach and triggers here are the ones [he suggested](https://github.com/pocket-id/pocket-id/issues/1473#issuecomment-5407475383). Figured that thread might be useful background for the final pass.

stonith404:
@alecrubin Thanks for the PR! I'll try to review it as soon as possible.

stonith404 (approved):
Did some minor changes but besides that LGTM. Thanks!

PR #1756, opened 2026-09-16
feat: include passkey icons based on AAGUID

### LINKED ISSUE #1735: 🚀 Feature: show icon for passkey from AAGUID
### Feature description

Pocket ID should use https://github.com/passkeydeveloper/passkey-authenticator-aaguids to show the user an icon for their passkeys.

### Pitch

As the number of passkeys grows for a user it gets harder to parse the list. Icons specific to the authenticator would help here. We can get the icon from the AAGUID, if it's set, and fetch the entry from the community-maintained https://github.com/passkeydeveloper/passkey-authenticator-aaguids.

This is related to but orthogonal to #1638.

michaelbeaumont:
## What does this PR do?

Closes #1735 

Records and uses the AAGUID to select an icon instead of the placeholder. There's no way to choose a specific icon.

Pocket ID wasn't storing the AAGUID before so it only works with new keys :shrug:. Of course an additional picker feature could be added.

Oh and this only supports the community AAGUID repo not MDS because they don't constrain their icons as much so guaranteeing a nice looking icon is harder.

The CI changes weren't tested in vivo.

## Screenshots

passkey-icons

## AI Usage

Used Claude but made improvements to narrow its types, aggressively removed comments, tested and questioned every thing it did with disdain because I do truly hate that it exists.

## Contributing Guidelines


backend/internal/dto/webauthn_dto.go
```diff
@@ -15,9 +16,13 @@ type WebauthnCredentialDto struct {
...
+	AAGUID string                  `json:"aaguid"`
+	Icon   model.WebauthnIconState `json:"icon" enums:"none,shown,hidden" swaggertype:"string"`
+
 	CreatedAt datatype.DateTime `json:"createdAt"`
 }
 
 type WebauthnCredentialUpdateDto struct {
-	Name string `json:"name" binding:"required,min=1,max=50"`
+	Name       *string `json:"name" binding:"omitempty,min=1,max=50"`
+	IconHidden *bool   `json:"iconHidden"`
```
stonith404: Tbh I don't really see a reason why the user wants to hide an icon explicitly?

backend/internal/model/webauthn.go
```diff
@@ -20,9 +20,31 @@ type WebauthnCredential struct {
 	BackupEligible bool `json:"backupEligible"`
 	BackupState    bool `json:"backupState"`
 
+	AAGUID     string `gorm:"column:aaguid;default:00000000-0000-0000-0000-000000000000"`
```
stonith404: Instead of having a zero AAGUID as default value I would make the column nullable with a pointer. This makes it more explicit that no AAGUID is set.

backend/internal/webauthn/handler.go
```diff
@@ -206,3 +211,24 @@ func (h *handler) reauthenticate(c *gin.Context) error {
...
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	utils.SetCacheControlHeader(c, 24*time.Hour, 7*24*time.Hour)
+
+	c.Header("X-Content-Type-Options", "nosniff")
+	c.Header("Content-Security-Policy", "default-src 'none'; style-src 'unsafe-inline'; sandbox")
+
```
stonith404:
```suggestion
```
`X-Content-Type-Options` is not necessary and the global CSP should be enough for how we load the image.

backend/internal/dto/webauthn_dto.go
```diff
@@ -15,9 +16,13 @@ type WebauthnCredentialDto struct {
...
 
+	AAGUID string                  `json:"aaguid"`
+	Icon   model.WebauthnIconState `json:"icon" enums:"none,shown,hidden" swaggertype:"string"`
+
 	CreatedAt datatype.DateTime `json:"createdAt"`
 }
 
 type WebauthnCredentialUpdateDto struct {
-	Name string `json:"name" binding:"required,min=1,max=50"`
+	Name       *string `json:"name" binding:"omitempty,min=1,max=50"`
```
stonith404:
```suggestion
	Name       string `json:"name" binding:"required,min=1,max=50"`
```
Since we currently only support `PUT` and not `PATCH` we shouldn't  make `Name` optional.

backend/internal/dto/webauthn_dto.go
michaelbeaumont: Yeah, I can remove it. Some of the icons are kinda ugly tbh and look kinda cheap, just IMO, so I thought maybe some people'd rather not have them but :shrug:

backend/internal/model/webauthn.go
michaelbeaumont: AFAIK there isn't really the option to "not set" the AAGUID. There's always 16 bytes. Authenticators zero AAGUID on purpose so as not to identify the authenticator. So we'd have null and 00.. but there's no semantic difference between null and 00...

backend/internal/dto/webauthn_dto.go
stonith404: Yeah I definitely I agree some icons look very ugly but I still feel like this feature won't use anyone. If there is really demand for it we can wait for a feature request.

backend/internal/model/webauthn.go
stonith404: Oh okay you are right, I was just assuming it because of this check: https://github.com/pocket-id/pocket-id/pull/1756/changes#diff-cf1f8fcb084fee477a3b9dc1163aed3b12a56f515d0dfb2a600ddbd69d0be9a5R189

In this case we can just remove check for empty IDs there, since it will never be empty anyways.

backend/internal/dto/webauthn_dto.go
michaelbeaumont: Will adjust the tests

backend/internal/dto/webauthn_dto.go
michaelbeaumont: done

backend/internal/model/webauthn.go
michaelbeaumont: done

backend/internal/utils/aaguid_util.go
```diff
@@ -4,19 +4,38 @@ import (
...
 	"sync"
 
 	"github.com/pocket-id/pocket-id/backend/resources"
 )
 
+const aaguidIconsDir = "aaguid-icons"
+
+// ZeroAAGUID is authenticators that don't want to be identified and is the
+// default value.
+const ZeroAAGUID = "00000000-0000-0000-0000-000000000000"
```
ItalyPaleAle: use uuid.Nil().String() instead of re-defining

.github/workflows/update-aaguids.yml
```diff
@@ -25,6 +25,54 @@ jobs:
...
+      - name: Extract authenticator icons
+        env:
+          ICON_DIR: backend/resources/aaguid-icons
+          MAX_ICON_BYTES: "65536"
+        run: |
+          mkdir -p "$ICON_DIR"
+          rm -f "$ICON_DIR"/*.svg
+
+          written=0
+          while IFS=$'\t' read -r aaguid variant payload; do
```
ItalyPaleAle: This is starting to look like a lot of code... How about we write this in Go, for example in a tool in the scripts folder? Or at least move the script to the scripts folder

backend/internal/webauthn/handler.go
```diff
@@ -206,3 +211,21 @@ func (h *handler) reauthenticate(c *gin.Context) error {
 	c.Status(http.StatusNoContent)
 	return nil
 }
+
+func (h *handler) getAuthenticatorIcon(c *gin.Context) error {
```
ItalyPaleAle: I know you said you removed all comments but we actually do want comments, for example here :) It's instructed in AGENTS.md too

backend/resources/aaguid-icons/README.md
```diff
@@ -0,0 +1,3 @@
+Authenticator icons generated by the `Update AAGUIDs` workflow from the community AAGUID list.
```
ItalyPaleAle: Just replace this file with a `.gitinclude` as commonly used

backend/resources/migrations/postgres/20260916120000_webauthn_credential_icon.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE webauthn_credentials ADD COLUMN aaguid TEXT NOT NULL DEFAULT '00000000-0000-0000-0000-000000000000';
```
ItalyPaleAle, line 1: Do we need to set a nil UUID as default? Couldn't we just set an empty value and read it as nil when retrieving the row?

ItalyPaleAle (changes requested):

backend/resources/aaguid-icons/README.md
michaelbeaumont: Unfortunately goembed doesn't appreciate:

```
internal/model/types/string_list.go:7:2: could not import github.com/pocket-id/pocket-id/backend/internal/utils (internal/utils/aaguid_util.go:15:2: could not import github.com/pocket-id/pocket-id/backend/resources (resources/files.go:7:87: pattern aaguid-icons: cannot embed directory aaguid-icons: contains no embeddable files)) 
```

backend/internal/webauthn/handler.go
michaelbeaumont: None of the other functions in this webauthn handler have comments so it seemed out of place. What should the comment cover? I renamed the function in any case

backend/resources/migrations/postgres/20260916120000_webauthn_credential_icon.up.sql
michaelbeaumont: I figured we don't need two values that mean unset. Is there an advantage to having NULL as well?

backend/resources/aaguid-icons/README.md
michaelbeaumont: alternatively i could run the script and commit the icons in this PR instead of waiting for the nightly update and we just don't worry about the directory being empty, but it seemed very noisy.

.github/workflows/update-aaguids.yml
michaelbeaumont: done. it now sorts the JSON by UUID keys because golang but should be ok?

backend/internal/webauthn/handler.go
ItalyPaleAle: Yeah and that was a problem when that code was written. The guidelines for agents are in AGENTS.md - I would just tell Claude to add comments to modified code following AGENTS.md

backend/resources/aaguid-icons/README.md
ItalyPaleAle: Sure it will happen anyways!

backend/resources/migrations/postgres/20260916120000_webauthn_credential_icon.up.sql
ItalyPaleAle: No, I don't mean "NULL" as in the SQL NULL. I just mean to not store the zero UUID,but rather an empty string. When reading the row, empty string would be treated same as zero UUID.

ItalyPaleAle: No, I don't mean "NULL" as in the SQL NULL. I just mean to not store the zero UUID,but rather an empty string. When reading the row, empty string would be treated same as zero UUID.

michaelbeaumont: yeah seems like we have two values that mean the same thing then still.
i like the invariant of "this string is always a valid UUID" better but if you prefer

ItalyPaleAle: I prefer it because it's 36 less bytes per each row, especially if it's common to have no UUID here. But I don't feel strongly about this.

backend/internal/webauthn/handler.go
michaelbeaumont: sure thing

backend/resources/migrations/postgres/20260916120000_webauthn_credential_icon.up.sql
michaelbeaumont: a client never sends fewer bytes, they always send an AAGUID, they just send 16 0x00 bytes. AFAIU at least. so do you want to also normalize the zero UUID back to empty string? or maybe i misunderstood

backend/resources/aaguid-icons/README.md
michaelbeaumont: done

backend/resources/migrations/postgres/20260916120000_webauthn_credential_icon.up.sql
michaelbeaumont: as far as what's more common. found in the webauthn level 3 spec:

> [aaguid](https://www.w3.org/TR/webauthn-3/#authdata-attestedcredentialdata-aaguid) in [attested credential data](https://www.w3.org/TR/webauthn-3/#attested-credential-data) is no longer zeroed when [attestation](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-attestation) preference is [none](https://www.w3.org/TR/webauthn-3/#dom-attestationconveyancepreference-none): [§ 5.1.3 Create a New Credential - PublicKeyCredential’s [[Create]](origin, options, sameOriginWithAncestors) Internal Method](https://www.w3.org/TR/webauthn-3/#sctn-createCredential)

it would seem it's very likely going to be more common than it was to send an AAGUID? Apparently most were sending it no matter what anyway, it seems like only WebKit that would zero it out with `"none"`.

Abhirup0 (commented):
Reviewed the full diff across the backend, migrations, and frontend. The design is very clean—embedding the SVGs at build time and indexing them lazily in memory avoids runtime filesystem access and completely closes any path traversal vectors. Backwards compatibility for existing credentials (defaulting to the zero UUID and falling back to the generic key icon) is solid.

A couple of minor suggestions:

1. **UUID Case Sensitivity**: In `HasAuthenticatorIcon` and `OpenAuthenticatorIcon` (`backend/internal/utils/aaguid_util.go`), the map lookup uses `aaguid` directly. Because filenames on disk are lowercase, passing an uppercase UUID string (e.g. via direct API calls or tests) returns a 404. Normalizing with `aaguid = strings.ToLower(aaguid)` at the start of both functions would ensure full RFC 9562 case-insensitivity.
2. **AAGUID Validation in Update Script**: In `scripts/update-aaguid-icons.go`, adding `if _, err := uuid.Parse(aaguid); err != nil { continue }` before `writeIcon` would be good defensive practice to guarantee no malformed keys from upstream can write outside the destination directory.
3. **Route Integration Test**: Adding a test asserting that `GET /api/webauthn/authenticator-icons/:aaguid` returns 200 with `image/svg+xml` without requiring auth headers would ensure the route registration remains unauthenticated in future refactors.

Overall looks great and ready to go once those small points are addressed.

stonith404 (approved):
Thanks for your contribution. Especially the discussions and iterations where good, nowadays this is pretty rare because of AI drive-by PRs.

ItalyPaleAle (approved):

PR #1762, opened 2026-09-19
feat(oidc): add TRUSTED_LOGO_HOSTS to allow private IP logo downloads

### LINKED ISSUE #1694: Allowlist for logo/darkLogo URL ssrf check (private IPs)
repro: setting an oidc client's logo url to an internal-only host (private ip) fails outright: `{"error":"Logo URL is not allowed","code":"validation_failed","details":{"fields":[{"field":"logoUrl","code":"invalid_value","message":"must be a public HTTP or HTTPS URL"}]}}`

cause: `IsURLPrivate` (`backend/internal/utils/ip_util.go`) rejects any resolved ip matching `IsPrivateIP` (loopback, RFC1918, link-local, unspecified, tailscale CGNAT) unconditionally, no override anywhere. checked `LOCAL_IPV6_RANGES` too, that only extends the blocklist, doesn't allowlist anything.

ask: an allowlist for this specific check so self-hosted admins can point logo urls at their own internal asset servers (icon mirrors, internal CDNs) without disabling SSRF protection wholesale.

shape: you already have this exact pattern for `TRUST_PROXY` (`env_config.go`, `TrustProxyConfig.UnmarshalText`) — single ip, CIDR, or comma-separated list of either, via `net.ParseIP`/`net.ParseCIDR`. reusing that type/parser for something like `TRUSTED_LOGO_HOSTS` would be consistent with what's already there, not a new mechanism.

workaround for now: pointing logoUrl at a public CDN instead of self-hosting the mirror. works, defeats the point for anyone running this fully internal/air-gapped.

Abhirup0:
## What does this PR do?

Closes #1694

When fetching OIDC client logos (`logoUrl` / `darkLogoUrl`), Pocket ID checks whether target addresses resolve to private IPs to protect against SSRF. In self-hosted setups where assets are served on internal IPs or local subnets, this prevents downloading client logos.

This PR adds an optional `TRUSTED_LOGO_HOSTS` environment variable:
- Accepts comma-separated CIDR ranges, IP addresses, hostnames, or boolean flags (`true` permits all private IPs via `0.0.0.0/0, ::/0`, `false` keeps default strict behavior).
- Extends `utils.IsURLPrivate()` to check whether resolved private IPs match the allowlist before flagging as private. If any resolved IP is not allowlisted, the request remains blocked to guard against DNS rebinding.
- Adds regression unit tests in `internal/common`, `internal/utils`, and `internal/service`.

## Screenshots
N/A (Backend configuration and network utility change)

## AI Usage
Used an LLM to help draft the configuration parser and regression test cases. I reviewed, tested, and verified all changes locally against the test suite.

## Contributing Guidelines


ItalyPaleAle (changes requested):
The linked issue has no upvotes and has a very viable workaround.

IMHO this PR adds another env var which is unnecessary complexity, for a very small issue.

Before opening a PR where the solution isn't immediate, it is always best practice to comment on the issue and work with the maintainers to come up with a design. This avoids wasting both your tokens, and most importantly the maintainers' time.

PR #1768, opened 2026-09-20
feat: add FRANCIS_HOST to connect to a standalone Francis runtime

ItalyPaleAle:
## What and why

FRANCIS_HOST decides where the Francis actor runtime lives. When set to "embedded" (the default), Pocket ID starts the runtime inside its own process.

Any other value is the address, or a comma-separated list of addresses, of a standalone Francis runtime. Pocket ID then connects to it as a remote actor host and starts no embedded runtime.

Because when using a remote runtime, it's likewise not possible to enforce a single instance of Pocket ID is running at once, the env vars currently have the `EXPERIMENTAL_` prefix, are **undocumented**, and show a warning if used.

Notes:

- Connecting to a standalone runtime also needs FRANCIS_HOST_PSK or FRANCIS_HOST_JWT_FILE, and optionally (but recommended) FRANCIS_CA.
- When connecting to a remote runtime, exporting Pocket ID data does not include the actor state, which will need to be backed up and restored separately

## Related issue

#1549 

## Contributing Guidelines


ItalyPaleAle:
image

Note these tests have been renamed because now they include the runtime type (embedded or remote)

.github/workflows/e2e-tests.yml
```diff
@@ -57,14 +65,14 @@ jobs:
         run: depot configure-docker
 
       - name: Cache Playwright Browsers
-        uses: actions/cache@v5
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```
stonith404, line 68: Is there any reason why we use the digest instead of the tag here?

ItalyPaleAle: Yes it's actually what GitHub recommends: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

backend/internal/cmds/export.go
```diff
@@ -49,18 +50,27 @@ func runExport(ctx context.Context, flags exportFlags) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to initialize the actor host's data provider: %w", err)
+		}
+		defer func() {
+			_ = provider.Close()
+		}()
+
+		actorsProvider = provider
+	} else {
+		printRemoteActorDataNotice("The actor data is NOT included in this export", "back it up separately with: francis runtime backup -f actors.bin")
```
stonith404, line 72: Would it be a huge effort to add support for remote hosts?

stonith404 (approved):
Besides my comment above everything LGTM, thanks!

.github/workflows/e2e-tests.yml
stonith404: I didn't know that, we probably should adapt the other actions as well.

backend/internal/cmds/export.go
ItalyPaleAle: I would need to think about this more, it's not really simple. The backup requires accessing the state store directly, among other things.

I am leaning towards making support for remote Francis runtimes as an advanced thing that few users will want to do, and for those users they need to accept the trade-offs such as having to manage backups themselves. Most users running Pocket ID in HA will use 2, maybe 3 replicas at most, which are totally fine with the embedded runtime.

.github/workflows/e2e-tests.yml
kmendell: I think adapting the other actions to pinned shas isnt a bad idea theres a option in the org settings to only allow pinned actions too which we could enable.

kmendell (approved):

PR #1772, opened 2026-09-21
chore: pin all GH Actions to a specific SHA

ItalyPaleAle:
As discussed on the other PR, this is recommended by GitHub


.github/workflows/release.yml
```diff
@@ -68,18 +68,9 @@ jobs:
...
-        with:
-          run_install: false
-
-      - name: Setup Node.js
-        if: steps.version.outputs.tag != ''
-        uses: actions/setup-node@v6.5.0
-        with:
-          node-version: 24
-          cache: pnpm
+        uses: pnpm/setup@fbda4c85fc2e1e08721cd8763afea8f48d60f024 # v3.0.0
```
ItalyPaleAle, line 73: Note also replaced the pnpm action with the new one, which both installs pnpm and Node.js

.github/workflows/e2e-tests.yml
```diff
@@ -47,22 +47,17 @@ jobs:
...
-
-      - name: Setup Node.js
-        uses: actions/setup-node@v6.5.0
-        with:
-          node-version: 24
-          cache: "pnpm"
+      - name: Setup Node.js and pnpm
+        uses: pnpm/setup@fbda4c85fc2e1e08721cd8763afea8f48d60f024 # v3.0.0
 
       - name: Set up Depot CLI
```
kmendell, line 55: Github wont let me highlight the damn lines to suggest changes, i can do this tomorrow or when i have some time, but i made a PR to depot action awhile back to add the configure-docker input so we can consolidate all of these into. If you want to change that feel free otherwise i should have some time tomorrow 

```
    - name: Set up Depot CLI
      uses: depot/setup-action@91bc8495a33ebfc504ffc89e5674379ccf23c29c # v1.7.2
      with:
        configure-docker: true
```

kmendell (approved):
looks good to me!

.github/workflows/e2e-tests.yml
kmendell: I got it done.

kmendell:
@stonith404 Did we miss adding this test to the ruleset to be required? 
image

Wanted to check before i just modified it to include it

